Issue #269, opened 2014-07-29
Help with coding to support same-origin iframes


swang:
Hi

Probably related to #183, I am trying to get medium-editor to support text selection around localized iframes. I want to do this because I'm writing a HTML template editor that stores the output inside of an iframe to isolate its stylesheets. It would be nicer to not have to inject the medium-editor library into the iframe.

As long as the iframe is in the same origin as the document it's inside of we can access it as any regular DOM element. However, the iframe actually has it's own window/document objects so using window.getSelection() / document.getSelection() etc will not actually work. I've attempted to work around this by creating a function that returns their respective window/document objects from `document.activeElement`.

The problem I am running into is when I click on any of the toolbar buttons, a lot of different events fire before the button's "click" event. Mainly the blur/mouseup event on [line 505-506 here](https://github.com/daviferreira/medium-editor/blob/08aa1946150f27798665a576bba597f1ba30b95b/dist/js/medium-editor.js#L505-L506) fires before the click event for the buttons on the toolbar.

What happens is the entire event will fail because it goes through the `checkSelectionWrapper()`/`checkSelection()` function and fails because `newSelection` seems to return an empty string. I've tried changing `newSelection` to be `(document.activeElement.contentWindow || window).getSelection()` but unfortunately it seems that the iframe loses focuses when the toolbar becomes activated.

Is there a reason why those events needs to fired there? Because it fires first and fails, causing the toolbar to be hidden, I can't just save the last selected text and then pass that into the click event for the button.

Any recommendations/suggestions?

Thanks.

Issue #183, opened 2014-03-24
Toolbar not showing in iframe


phpMagpie:
I need to call the editor on elements within an iframe.  

I'm using the following code within the parent window (element is part of the iframe, or a snippet to be inserted into the iframe):
var editors = new MediumEditor( $(element).find('h1, p') );

This is adding contenteditable params to the elements and I can edit content, but the medium-editor toolbar is not showing.  No errors are being output, the toolbar just fails to show.

I have tested that medium-editor works when called on an element in the parent window, so just an issue with working on elements in the iframe.

PR #324, opened 2014-11-13
added custom doc and win functionality

botverse:
This helps to load the editor in a parent frame but edit inside an iframe.

This has been discussed before here #183 and here #269. The scenarios where this can come handy is where you are developing a drag and drop editor where the text is edited inline but everything happens inside an iframe, completely isolated from the upper frame.

You still need to override the `setToolbarPosition` in your side of things, a future improvement could be pass functions to calculate this:

``` javascript
this.toolbar.style.top = buttonHeight + boundary.bottom - this.options.diffTop + offset - this.toolbar.offsetHeight + 'px';
```

But that's another story.

### DIFF
--- a/spec/init.spec.js
+++ b/spec/init.spec.js
@@ -81,6 +81,8 @@ describe('Initialization TestCase', function () {
                 disableEditing: false,
                 disableToolbar: false,
                 elementsContainer: document.body,
+                contentWindow: window,
+                ownerDocument: document,
                 firstHeader: 'h3',
                 forcePlainText: true,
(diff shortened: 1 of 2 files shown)


swang:
Nice to see this implemented, although the project I needed it for is no longer being developed. Will remember this for future reference though.

pingram3541:
Yeah, I wound up pushing the script into the iframe and running it within the child frame since I already had event listeners from my external codeMirror object, this worked out just fine =)

Good to now, this is a great add!

MichalSchwarz:
:+1:

PR #345, opened 2014-12-16
Add optional 'onhidetoolbar' method

nchase:
Use case: as a developer, I want to add some custom behavior that will fire on my MediumEditor instance when the toolbar is hidden.

This changeset also adds a little bit of code that will clear any existing `setTimeout` function that exists when `blurFunction` is re-called.


daviferreira:
Nice job, thanks :)

daviferreira:
@nchase Btw, could you be so kind to update the README with your new options from your pull requests? Thanks!

PR #347, opened 2014-12-17
support Japanese Kanji conversion

auxout:
when press enter key to confirm Kanji conversion. it was created the unwanted paragraph.
so check the keyup and keydown keycode and if both are equarl then trigger the event.

this is because when press enter key for Kanji conversion. it's returns different keycodes both up(229) and down(13)

http://en.wikipedia.org/wiki/Japanese_input_methods#Kana_to_kanji_conversion


daviferreira:
Somehow it broke the anchor form tests. Could you take a look? :)

PR #351, opened 2014-12-31
Support disabling anchorForm

nmielnik:
When 'disableAnchorForm' is passed as an option:
- Don't create the anchor form element
- Don't attach event handlers to the document, window, and/or anchor form which are used soley for handling user interaction with the anchorForm


daviferreira:
Awesome :)

PR #352, opened 2014-12-31
Support disabling placeholders

nmielnik:
When `disablePlaceholders` is passed as an option to medium-editor:
- Do not add/remove the 'medium-editor-placeholder' class on the contenteditable element
- Don't attach event handlers which had the sole purpose of handling when a placeholder should be shown/hidden

### DIFF
--- a/spec/init.spec.js
+++ b/spec/init.spec.js
@@ -80,6 +80,7 @@ describe('Initialization TestCase', function () {
                 disableDoubleReturn: false,
                 disableEditing: false,
                 disableToolbar: false,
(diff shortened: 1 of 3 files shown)


daviferreira:
Hey @nmielnik, thanks for all your work. Could you be so kind to update the README with your new options? Cheers!

nmielnik:
Thanks for pulling this in!
I've updated the documentation (#357) to cover everything that @nchase and I had added, as well as the 'staticToolbar' and 'stickyToolbar' options, when I noticed weren't mentioned in the documentation.

PR #360, opened 2015-01-08
Handle ESC key in link editor

jbellsey:
When the link editor is open, let the user hit ESC to close it and return to the main toolbar (just like the X button)

### DIFF
--- a/src/js/medium-editor.js
+++ b/src/js/medium-editor.js
@@ -1220,6 +1220,11 @@ else if (typeof define === 'function' && define.amd) {
 
                     self.createLink(this, target, button);
                 }
+                else if (e.keyCode === 27) {
+                    e.preventDefault();
+                    self.showToolbarActions();
+                    restoreSelection.call(self, self.savedSelection);
+                }
             });
 
             this.on(linkSave, 'click', function(e) {


src/js/medium-editor.js
```diff
@@ -1220,6 +1220,11 @@ else if (typeof define === 'function' && define.amd) {
 
                     self.createLink(this, target, button);
                 }
+                else if (e.keyCode === 27) {
+                    e.preventDefault();
+                    self.showToolbarActions();
+                    restoreSelection.call(self, self.savedSelection);
```
nmielnik: It seems like we would want ESC to behave similar to ENTER/RETURN except that it would actually create the link.  If that's the case, would it be a good idea to factor out the part of the `createLink` method that hides the anchorForm, shows the toolbar, and manipulates the selection into a separate function and just call that from here? This would make ESC go down the same code path as ENTER, just without actually embedding/updating the link.

tmcconechy: I think ESC should cancel and not create the link thats the usualy standard for dialog behavior for that key... 

From:
http://access.aol.com/dhtml-style-guide-working-group/#dialogmodal
"ESC There should be a method to close the dialog without taking any action."

nmielnik: Agreed, there was actually a typo in my first sentence.  I'm not saying ESC should behave differently than you described, I was suggesting that ESC share a code path with ENTER/RETURN.

jbellsey: If you're suggesting to DRY it out by creating a single function that can be called by all UI actions intended to do the same thing, then I'm all for it. 

Something like this? In this case, it's a two-line function, which may or may not be worthy of factoring out. 

    function restoreEditorState() {
        this.showToolbarActions();
        this.restoreSelection(this.savedSelection);
    }

nmielnik: Yes, I'm suggesting something like that.  However, this is what is at the end of 'createLink', which is called when the user hits ENTER

```
this.checkSelection();
this.showToolbarActions();
input.value = '';
```

Ideally, `restoreEditorState` would be something that both `createLink` and your ESC handling would both call.

jbellsey:
That sounds good. Please feel free! I'm not familiar enough with the code to do the refactoring.

src/js/medium-editor.js
jbellsey: That sounds good. Please feel free! I'm not familiar enough with the code to do the refactoring.

PR #361, opened 2015-01-08
Standardize usage of setTimeout for UX delays vs debouncing vs deferring

nmielnik:
There is a delay when hiding the toolbar, so that if you have multiple instances of medium-editor on the page, you can cause multiple toolbars to appear at the same time.  This delay is NOT controlled by the `delay` option, it is hard coded and doesn't appear to be necessary, at least as far as I can tell.

The goal of these changes is to make sure we're using the `delay` option properly, and try to remove places where a hard-coded delay is actually creating a UI lag. 

+ The 'delay' option indicates that it should be used for the amount of time to delay the toolbar or anchorPreview from being displayed.  So, ensure this value is only being used in places where we would show the toolbar or anchor preview
+ For places in the code where we just want to debounce (prevent a method from being called too many times too rapidly) use a standard time interval and a `debounce` helper method
+ For places in the code where we just want to delay a method from being called immediately, again use the debounce time interval via the `debounce` helper
+ Create standard helper methods for hiding/showing toolbar as well as checking if toolbar is visible.  Use these everywhere to help control flows which manipulate the toolbar.

### DIFF
--- a/src/js/medium-editor.js
+++ b/src/js/medium-editor.js
@@ -1,6 +1,4 @@
 /*global module, console, define*/
-var DEBOUNCE_INTERVAL = 50;
-
 function MediumEditor(elements, options) {
     'use strict';
     return this.init(elements, options);
@@ -40,7 +38,8 @@ else if (typeof define === 'function' && define.amd) {
 
     // https://github.com/jashkenas/underscore
     function debounce(func, wait) {
-        var timeout, 
+        var DEBOUNCE_INTERVAL = 50,
+            timeout, 
             args, 
             context, 
             timestamp, 
--- a/src/js/medium-editor.js
+++ b/src/js/medium-editor.js
@@ -320,6 +320,10 @@ else if (typeof define === 'function' && define.amd) {
             }
         },
 
+        // handleBlur is debounced because:
+        // - This method could be called many times due to the type of event handlers that are calling it
+        // - We want a slight delay so that other events in the stack can run, some of which may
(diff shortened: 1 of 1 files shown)


nchase:
this looks cool. if we're going to have some places in the code that use `setTimeout` and others that use `debounce`, could we have a comment in the code explaining when each one is appropriate?

src/js/medium-editor.js
```diff
@@ -1,4 +1,5 @@
 /*global module, console, define*/
+var DEBOUNCE_INTERVAL = 50;
```
hacfi: Is this var global on purpose? Why not move it down to line 22?

nmielnik: I had placed it there originally as I agree that's the right place for it. However, jslint was not a fan of a var being declared there.  Since this constant is really only used inside of `debounce()`, do you think we should just move its declaration in there?

hacfi: I don’t see any disadvantages from moving it to `debounce()` - it’s not used anywhere else and can be changed with the `wait` argument.

nmielnik: Ok, I've moved the declaration inside of `debounce()`, thanks for the feedback

nmielnik:
@nchase added some comments to explain when using setTimeout vs debounce

madwire:
Just want to check that this works with multi editor using a single instance as currently version 2.0 does not work and I was about to start looking at this.

This is v2.0.0
editor

nmielnik:
Issues related to this pull request should be resolved by #365

PR #367, opened 2015-01-16
Extended extensions

CharlGottschalk:
This change passes the editor instance before creating the toolbar so
that we can access properties when creating buttons in extensions. It
also adds functionality to allow extensions to add forms to the toolbar.

It adds the following functions:

* `initExtensions()` which gets called after the instance is passed.
This allows extensions to create buttons etc.
* `addExtensionForms()` which gets called after the toolbar is created.
This appends any forms from extensions.
* `showForm()` which displays an extension's form.
* `hideForm()` which hides an extension's form

Some other minor changes:

* changed `.medium-editor-toolbar-form-anchor` to more generic
`.medium-editor-toolbar-form` in *medium-editor.scss*.
* changed `.medium-editor-toolbar-anchor-input` to more generic
`.medium-editor-toolbar-input` in *medium-editor.scss*.
* Updated all theme *.scss* files and *medium-editor.js*
 to reflect CSS changes.
* In `setToolbarPosition()` the toolbar is centered horizontally
relative to the editor container when `staticToolbar` is set to `true`.
This looks better when the toolbar has a lot of buttons. Maybe this
should be a property like `centerToolbar`?
medium-editor-center


Dayjo:
I like the idea of `centerToolbar` or maybe even `toolbarAlign: center | left | right `

CharlGottschalk:
I agree, so I added `toolbarAlign` property.

Dayjo:
Lovely stuff! Not sure I'll have any time to test or properly review this today, but perhaps at the weekend! 

Thanks for the contributions!

CharlGottschalk:
Only a pleasure...

src/js/medium-editor.js
```diff
@@ -978,8 +1031,18 @@ else if (typeof define === 'function' && define.amd) {
                     this.toolbar.style.top = containerTop - this.toolbar.offsetHeight + "px";
                 }
 
-                this.toolbar.style.left = containerRect.left + "px";
-
+                if (this.options.toolbarAlign) {
```
nmielnik: Perhaps you could factor out this `toolbarAlign` stuff into a separate pull request?  I would imagine it would be a lot easier to merge a pull request like that, then when it's attached to the form stuff you've added (which is very cool, but also significantly more complicated and impactful).  Just a thought though.

src/js/medium-editor.js
```diff
@@ -299,7 +299,9 @@ else if (typeof define === 'function' && define.amd) {
             }
             // Init toolbar
             if (addToolbar) {
-                this.initToolbar()
+                this.passInstance()
```
nmielnik: :+1: I had tried to find other ways to get the instance of medium-editor passed to the extensions before the calls to `getButton()` were happening, but this is much cleaner (both calling `passInstance()` sooner, and adding an `init()` method into the extensions).  I look forward to using this stuff.

dist/js/medium-editor.js
```diff
@@ -306,6 +308,28 @@ else if (typeof define === 'function' && define.amd) {
             return this;
         },
 
+        /**
+         * Calls an init() function on all registered extensions
+         * Checks whether init() exists before calling
+         *
+         */
+        initExtensions: function () {
```
nmielnik: You can use the `this.callExtensions('init')` helper method here

daviferreira:
Should we merge it like this (looks good to me) or are you going to take @nmielnik suggestions?

dist/js/medium-editor.js
CharlGottschalk: I removed the `initExtensions()` method and replaced it with the `callExtensions('init')` helper method like you suggested.

I had to add `return this` to the helper method.

CharlGottschalk:
What would best @daviferreira?

Dayjo:
:+1:

nmielnik:
@CharlGottschalk if I create a page using the sample code you provided for the Table form, when I click the `table` button, I'm unable to edit the inputs for the table.

If you're able to get this in a working state, could you add an additional page into the `demos` directory with it working?  I am doing some work around extensions and buttons and want to make sure I leave your code in a working state, but I'm not currently able to see it working (as I would expect it to work at least).

CharlGottschalk:
@nmielnik it seems that when you focus on the form or input, the toolbar blur causes it to close.
I'm busy sorting that out now. Will open a new branch with the changes and revert back soon.

CharlGottschalk:
Created a new pull request with fixes and a demo page added.
https://github.com/daviferreira/medium-editor/pull/409

PR #374, opened 2015-01-21
Bugfix: ignore delayed/throttled functions when editor disabled

brianreavis:
Currently, calling `deactivate()` when there's something waiting to be called because of `setTimeout()` or `throttle()` can lead to exceptions, because most methods assume the editor is active at the time of calling. This can lead to errors like:
```
Uncaught TypeError: Cannot read property 'classList' of undefined
    MediumEditor.showAnchorPreview (anonymous function)
```
This PR makes all async delay/throttle logic check `isActive` before calling the function.

### DIFF
--- a/spec/activate.spec.js
+++ b/spec/activate.spec.js
@@ -62,6 +62,31 @@ describe('Activate/Deactivate TestCase', function () {
             expect(editor.events.length).toBe(0);
         });
 
+        it('should abort any pending throttled event handlers', function () {
+            var editor, triggerEvents;
+
+            editor = new MediumEditor('.editor', {delay: 5});
+            triggerEvents = function() {
+                fireEvent(window, 'resize', null, false);
+                fireEvent(document.body, 'click', null, false, document.body);
+                fireEvent(document.body, 'blur', null, false);
+            };
+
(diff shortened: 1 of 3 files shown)


nmielnik:
This is a great improvement, especially the addition of a `delay()` method :+1:

spec/activate.spec.js
```diff
@@ -62,6 +62,31 @@ describe('Activate/Deactivate TestCase', function () {
...
+            triggerEvents();
+            jasmine.clock().tick(1);
+
+            // fire event again (handler delayed because of throttle)
+            triggerEvents();
+
+            spyOn(editor, 'positionToolbarIfShown').and.callThrough(); // via: handleResize
+            spyOn(editor, 'hideToolbarActions').and.callThrough(); // via: handleBlur
+            editor.deactivate();
+            jasmine.clock().tick(600);
```
nmielnik: Why 600 ms here?  Might be worth a comment to explain whether this is just arbitrary or a purposefully chosen time.

brianreavis: Yeah, it's arbitrary... something that should always be longer than the throttle threshold. I'll add a note about that

PR #375, opened 2015-01-21
Bugfix: off() doesn't remove any event listeners

brianreavis:
Currently `off()` doesn't remove any event listeners or remove the entry from `this.events` (memory leak). This is because of the `indexOf` check (which checks strict equality, not deep equality).
```js
[[1,2,3]].indexOf([1,2,3]) === -1
```


nmielnik:
Great catch

Dayjo:
:+1:  Good stuff!

PR #382, opened 2015-01-29
Protect header (h1, h2, etc) tags from common actions

yacomink:
This addresses several cases where header tags are destroyed or extended unexpectedly.

FIX: Hitting the delete key on empty paragraph before the header, deletes header screen recording 2015-01-23 at 05 11 pm

FIX: Hitting return at beginning of header makes it a paragraph screen recording 2015-01-23 at 05 12 pm

FIX: Hitting backspace at beginning of header makes it a paragraph 
screen recording 2015-01-29 at 01 29 pm


src/js/medium-editor.js
```diff
@@ -653,10 +689,74 @@ else if (typeof define === 'function' && define.amd) {
...
+                    p.innerHTML = '<br>';
+                    node.previousElementSibling.parentNode.insertBefore( p, node );
+                    e.preventDefault();
+                }
+
+            }
+            else if ( e.which === 46 // delete
+                && !isHeader.test(tagName) // not in a header
+                && isEmpty.test(node.innerHTML) // in an empty tag
+                && isHeader.test(node.nextElementSibling.tagName) ) // when the next tag *is* a header
```
nmielnik: can't `nextElementSibling` be null when the element doesn't have a sibling?

yacomink: Thanks! Pushed a check

PR #384, opened 2015-01-30
Added anchorInputCheckboxLabel as option to specify the Open in new window label

enricoberti:
Hi! 
Since we are using it in our Hue project (github.com/cloudera/hue) and we need to i18n all the labels/strings I saw this was missing!

Thanks for the library btw, it's awesome!


daviferreira:
:+1:

Issue #386, opened 2015-01-30
Get Saucelabs Working for medium-editor


nchase:
@daviferreira @nmielnik and anyone else, here's what I'm thinking:

First mentioned in #370. We should be able to get feedback on how medium-editor works in different browsers.

The [medium-editor/try-saucelabs-configuration](https://github.com/daviferreira/medium-editor/tree/try-saucelabs-configuration) branch is my first cut at this issue.

So far the branch is connecting and reporting to Sauce Labs whenever a Travis CI build is kicked off. Sauce is running tests for Mac and Windows in Safari, Chrome, Firefox, and IE11.

Here's what we still need to do:
- [x] Investigate/agree on some account to use for the Sauce Labs reports. Currently it's linked to my personal Sauce Labs account which is almost definitely not what we will want to do.
- [x] Make the tests work fairly well in all browsers (a few failures is OK, so long as they make sense and either report known issues and/or we can fix them up fairly reliably)
- [x] Report test result details to sauce (e.g. via the "Jasmine JS Reporter" mentioned [here](https://github.com/axemclion/grunt-saucelabs#test-result-details-with-jasmine))

daviferreira:
@nchase about the account, we can create a medium e-ditor email somewhere and share the access info to all collaborators and use that e-mail to create the Sauce Labs account.

nchase:
@daviferreira sounds good! feel free to make one or I'll get to it later.

nchase:
Tests are looking good on the new account on Sauce Labs. We need to set up reporting between the test runner and Sauce. Before we close that bullet point, we also need to nail down what platform/browser combinations we want to test against.

PR #387, opened 2015-01-30
Fix some tests that were broken in browsers

nmielnik:
- Fix two tests where the behavior between running in node vs the browser is inconsistent


nchase:
:thumbsup: – this is gonna help quite a bit when we're trying to get a read on what works/doesn't work in different browsers via #386

PR #388, opened 2015-01-30
nodeValue can be null, so make sure it exists before checking its length

nmielnik:


nchase:
weird whitespace changes :smirk: but otherwise looks good to me!

nmielnik:
I know, I'm the one that added the trailing spaces accidentally so hopefully they'll all be gone now.

PR #393, opened 2015-02-03
Fix a few issues with the standardizeSelectionStart option

patrick-webs:
* IE 11 issues an "Argument not optional error" on the call to createNodeIterator without 4 parameters being passed.
* IE 11 (and possibly other browsers) could get caught in an infinite loop if the DOM contained text nodes that had nothing but whitespace characters in them.


nmielnik:
awesome :+1:

PR #396, opened 2015-02-04
Fix Browser Tests in Non-Webkit-based Browsers

nmielnik:
+ Fix all the tests that were failing when running in firefox
+ Thanks to @nchase , also fixed all the tests that were still failing in IE
+ In cases where firefox or IE are doing something very differently from other browsers, branch the test cases to keep the intent of the case without having to remove them completely

### DIFF
--- a/spec/placeholder.spec.js
+++ b/spec/placeholder.spec.js
@@ -66,14 +66,24 @@ describe('Placeholder TestCase', function () {
     it('should add the default placeholder text when data-placeholder is not present', function () {
         var editor = new MediumEditor('.editor'),
             placeholder = window.getComputedStyle(editor.elements[0], ':after').getPropertyValue('content');
-        expect(placeholder).toEqual("'" + editor.options.placeholder + "'");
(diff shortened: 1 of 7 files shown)


spec/placeholder.spec.js
```diff
@@ -66,14 +66,24 @@ describe('Placeholder TestCase', function () {
     it('should add the default placeholder text when data-placeholder is not present', function () {
         var editor = new MediumEditor('.editor'),
             placeholder = window.getComputedStyle(editor.elements[0], ':after').getPropertyValue('content');
-        expect(placeholder).toEqual("'" + editor.options.placeholder + "'");
+        if (placeholder.indexOf('attr(') === 0) {
+            // In firefox, getComputedStyle().getPropertyValue('content') can return attr() instead of what attr() evaluates to
+            expect(placeholder).toEqual('attr(data-placeholder)');
+        } else {
```
nchase: to reduce duplication, you _could_ extract this `if/else` into a function here in placeholder.spec.js that accepts `placeholder` as an argument.

PR #397, opened 2015-02-05
Image drag&drop support

javiermarinros:
Image drag and drop support using HTML5 FileReader API, with custom styles when the editor is hovered and CSS animation when the image is loading. Tested on lastest Chrome and Firefox

### DIFF
--- a/src/js/medium-editor.js
+++ b/src/js/medium-editor.js
@@ -1,4 +1,4 @@
-/*global module, console, define, NodeFilter */
+/*global module, console, define, NodeFilter, FileReader */
 function MediumEditor(elements, options) {
     'use strict';
     return this.init(elements, options);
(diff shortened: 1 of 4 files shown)


src/js/medium-editor.js
```diff
@@ -1029,6 +1031,64 @@ if (typeof module === 'object') {
...
+                var files;
+                e.preventDefault();
+                e.stopPropagation();
+                files = Array.prototype.slice.call(e.dataTransfer.files, 0);
+                files.some(function (file) {
+                    if (file.type.match("image")) {
+                        var fileReader = new FileReader();
+                        fileReader.readAsDataURL(file);
+
+                        var id = 'medium-img-' + Math.random();
```
nmielnik: Perhaps `Date.getTime()` could be a better uniqueId generator?

nmielnik:
Looks like the build is failing, looking at the code I'm guessing it's some jslint issues.  Could you look into fixing those?

src/js/medium-editor.js
javiermarinros: If you drag and drop two files at a time, Date.getTime() would have the same value

javiermarinros:
Build fixed :)

daviferreira:
:+1:

PR #401, opened 2015-02-05
Fix showing and writing encoded links 

HHogg:
We're currently using ME on a platform that uses links similar to the following format... 
```html
http://{custom_domain}.com/
```
However the current anchor preview and anchor form fetch the link address by using the `href` property directly on the anchor element, the browser returns the encoded link, something like ....
```html
http://%7Bcustom_domain%7D.com/
``` 

By fetching from the attributes property instead it will return the unencoded link.


daviferreira:
Great stuff :+1:

Issue #398, opened 2015-02-05
Use constants for key codes to improve readability


robinwassen:
I would suggest to use named constants instead of using the raw number when checking key codes, for example:

```javascript          
if (e.which === 32) {
    node = getSelectionStart.call(self);
    .....
```

Would be easier to read if it was:

```javascript
if (e.which === KeyCode.ENTER) {
    node = getSelectionStart.call(self);
    .....
```

I can contribute with a pull request if you think this is a good idea.

nmielnik:
This has been on my to do list for a while, I think it's a great idea.

PR #402, opened 2015-02-06
Changed from number constants to named constants when checking key codes

robinwassen:
Changed the numeric constants to named "constants" to improve readability of the code, issue: #398 

Only included key codes that are in use and started to remove comments that got redundant by this change.


nmielnik:
:+1:

PR #404, opened 2015-02-06
Introduce DefaultButton object for built-in buttons

nmielnik:
+ Factor out built-in button behavior into a generic `DefaultButton` class, and separate JSON for each button
+ Introduce `commands` array, which has both buttons and extensions.
+ Prevent both a button and an extension from both executing (extensions replace buttons, but don't run both)
+ Remove "custom form" support until we have some test cases so it can be verified as working in this new paradigm
+ Add tests for all built-in toolbar buttons that pass in all browsers


nmielnik:
@daviferreira @nchase @Dayjo If you have time to look through this or give any feedback it'd be appreciated.

This could introduce 2 potential breaks to existing functionality:
1) The custom form work done in #367  There aren't any tests, and the code sample given in the README doesn't quite work, so I decided to remove that code and if we want to add it back later, we can figure out how to implement it with the new paradigm.
2) If users were passing an extension for a built-in command (ie a `bold` or `pre` extension) before all of the button's handling code would run, and then the extension code could run.  This seems like it would introduce problems and produce some weird cases, so I felt like it was better to leave it out and just have extensions completely replace the built-in commands if they are passed.

If we do move forward with this, it seems like a more major release since it could break some existing functionality, depending on how medium-editor was being used.  Any thoughts?

PR #410, opened 2015-02-09
update aria label so that setting secondHeader activates the toolbar but...

dvanderbeek:
...ton


nmielnik:
This looks good, could you fix the jslint failures?

dvanderbeek:
Sure, they should be fixed by the last commit

PR #414, opened 2015-02-09
Don't use styles for detecting underline + strikethrough

nmielnik:
`text-decoration` is not inherited by default, so we can't use it reliably to determine if text is underlined or strikethrough in the toolbar :disappointed:


nchase:
> `text-decoration` is not inherited by default

kind of surprising, but makes sense makes sense too – these styles will affect their child elements implicitly, but the style has no need to descend further into the tree.

PR #416, opened 2015-02-10
Improved command detection

nmielnik:
+ For most commands (where we're not using `formatBlock`) we can use the browser's built-in `document.queryCommandState()` method to detect whether the current selection is enabled or not.  This means we can do less manual checks to determine whether selected text is already **bold**, _italic_, etc.
+ By using built-in commands, we can do an even better job of detecting the style of selected text, even if it's being done indirectly by outside classes :smile: 
+ Re-add some tests for detecting bold, underline, and strikethrough that weren't possible without `queryCommandState()`
+ Update dist files


nchase:
:clap:

nchase:
I'm really into this.

Just want to mention that scribe (another `contenteditable`-related library) has a list of bugs/cross-browser inconsistencies that they have found in the `contenteditable`-related APIs. [they mention a few inconsistencies with `document.queryCommandState`](https://github.com/guardian/scribe/blob/master/BROWSERINCONSISTENCIES.md#documentquerycommandstate) but the list is small.

So long as we can ensure that this works everywhere that we list as supported in the medium-editor documentation, I think this is really awesome, but we should wait to merge until we can ensure that.

PR #423, opened 2015-02-12
Concat

daviferreira:
@nmielnik got some failing tests with buttons, can you help me?


daviferreira:
@nmielnik mmm, alright.. I just think this is basic behavior broken, when I click outside a selection I want the toolbar to disappear

nmielnik:
Yeah, I definitely don't disagree about fixing it, I'm just asking if there might be another way to fix it? This branch isn't in master just yet, so maybe we can find another way to accomplish it before we merge into master.

So the repro steps are:
1) Select some text (toolbar appears)
2) Click outside of the contenteditable element (toolbar is visible, but it should have disappeared)

Is that right?

daviferreira:
Yeha, sure, no problem breaking stuff here, didn't meant that :)

To reproduce: https://github.com/daviferreira/medium-editor/issues/417

nmielnik:
@daviferreira I believe I've worked around the issue and fixed the tests.  It's the same fix you had put in before, but with only a single `setTimeout` wrapped event handler, and without killing the event propagation.  Let me know if you think this is ok...if so, I think we're good to merge

PR #425, opened 2015-02-14
RegExp with 'g' param maintain state - don't declare globally

nmielnik:
When creating RegExp, if you use the `g` param the RegExp will maintain state as long as the variable exists in scope.  This means that the RegExp will behave differently each time its used after the first, depend on the original string that was used.

To avoid any issues with this, we should create the paste RegExp objects each time the paste runs (instead of declaring them one time and reusing them).


nchase:
:shipit:

PR #427, opened 2015-02-15
Add a wrapper function to avoid global scope pollution

daviferreira:
The idea here is to be free to use names like Util, Selection, Button etc. without polluting the global scope and just exporting the MediumEditor constructor function.


daviferreira:
cc @nmielnik @nchase

nmielnik:
:+1:

PR #428, opened 2015-02-15
Break anchorForm out into extension-anchor.js

nmielnik:
+ Move all the anchorForm handling logic into a new `extension-anchor.js` file
+ Fix issue where we were including the script files into the generated `_SpecRunner.html` in the wrong order
+ Replace references of anchorForm and other inputs with `this.anchorExtension`, which eventually should be merged into the `this.commands` array.
+ Fix up a few busted / brittle test cases


daviferreira:
Beautiful :+1:

PR #429, opened 2015-02-16
Full content actions

nmielnik:
Introduce 'full-content' actions, which apply an action (ie bold, underline, justifyLeft) to all of the contents within the editor.

These actions can be triggered by:
+ Calling `.execAction('full-<action>')` on MediumEditor
+ Creating a button with `action` set to 'full-<action>'


daviferreira:
Would be nice to have something about it on the README

PR #436, opened 2015-02-18
Simplify execAction and how buttons in the toolbar behave

nmielnik:
**NOTE:** Any release we cut after this commit will require a bump in the major release version, as any external dependencies on `execAction()` may no longer work as expected.

* Always check the state of the toolbar buttons after each button click (via `checkSelection()`)
* Simplify `execAction()` to simply exec an action, and then update the toolbar to reflect the result (via `checkSelection()`)


src/js/core.js
```diff
@@ -1015,35 +1013,38 @@ function MediumEditor(elements, options) {
...
+            /*jslint regexp: false*/
+
             // Actions starting with 'append-' should attempt to format a block of text ('formatBlock') using a specific
             // type of block element (ie append-blockquote, append-h1, append-pre, etc.)
             match = appendAction.exec(action);
             if (match) {
                 this.execFormatBlock(match[1]);
-                this.setToolbarPosition();
-                this.setToolbarButtonStates();
                 return;
```
nchase: this conditional block runs `execFormatBlock` which ultimately runs `execCommand`. `execCommand` returns a boolean that says whether or not the command was successful.  it would be great to pass that return value along instead of returning `undefined` here.

In other words, do you think it would be OK to `return this.execFormatBlock(match[1])` instead of having one line that runs `this.execFormatBlock(match[1])` and a second line for `return`? We aren't using that return value now, but  it might be useful at some point.

nchase: in the scenario above, it would also make sense to `return` from the `action === 'image'` and `else` case below too.

nmielnik: Yeah, good point, I will make an update.

nmielnik:
@daviferreira @nchase Just another reminder.
We need to increment the major build number before the next release, unless we decide these changes to execAction don't warrant that.

We also may want to hold off on doing an other release so that we can get some more functionality added in to help justify a major release.

PR #441, opened 2015-02-20
Clarify button terminology

nmielnik:
Activate/Deactivate are used for building and tearing down medium-editor.  So, we shouldn't use those exact  same terms when specifying whether a button should be displayed as active or inactive in the toolbar.

This is especially important as we will now be calling `deactivate` on all commands when medium-editor is deactivated, and calling `activate` on all commands when medium-editor is being activated.

I am suggesting `isActive()` as the getter/setter, and `isAlreadyApplied()` as the method for checking whether or not a specific node already has the command applied to it. 

Open to other ideas for terminology, or we can just take this as is?


nchase:
looks good to me. to @nmielnik's description above, it's worth noting that we discussed and settled on separating get/set behavior  from `isActive`. We now have `setActive`  and `setInactive`.

PR #442, opened 2015-02-20
Simplify blur check selection

nmielnik:
If we detect a blur in the contenteditable element, we want to prevent the call to `checkSelection()` if the `relatedTarget` was the input for the anchor form.

We can make this check more generic (by checking if the related target is a descendant of the toolbar) and only do the check on blur (we were attaching to multiple events, even though we only cared about this special case for blur).

This makes the logic more straight forward, and also helps extract more anchor form specific logic.  This check can now work for any extension which might add something into the toolbar which can grab focus.


daviferreira:
:+1:

PR #443, opened 2015-02-20
Configure coveralls

daviferreira:
Does this ```spec/jasmine-jsreporter.js``` needs to be at the sources for jasmine? Can't it be a helper or vendor? Otherwise it will get hit by the coverage report.

/cc @nchase @nmielnik


daviferreira:
https://coveralls.io/r/daviferreira/medium-editor

nmielnik:
It can definitely be moved out, I just put it there because there didn't seem to be any other pre-existing directory that made sense.  In fact, I would feel much better if you moved it out :hamster:

nchase:
We actually have higher than 70% code coverage.

The `classList` polyfill that I added is being covered because it's rolled into `dist/js/medium-editor.js`.

Is there some comment-based directive we can provide to tell coveralls _not_ to cover a particular section of code?

daviferreira:
That was going to be my next question :dancer:, can't we use the ```srcFiles``` instead of ```dist/js/medium-editor.js```? We can add the polyfill and everything else needed by saucelabs to the helpers dir on the jasmine config.

nchase:
makes good sense to me. @nmielnik what do you think?

nmielnik:
We need to be able to control the order of srcFiles.  When we specified them all to jasmine for creating `_SpecRunner.html`, it was including them in alphabetical order, which won't work.

nchase:
ick. forgot about that. does the grunt jasmine task _not_ inject the scripts in the same order we provide it?

daviferreira:
hmm, apparently moving the jsreporter broke saucelabs. I'm off now, if you guys want to take a look go ahead. Otherwise I will try again tomorrow :beers:

nchase:
@daviferreira if I remember correctly, the JSReporter setup stuff needs to happen before the tests get run. one possibility here is that we write our own custom grunt-jasmine template. there are other possibilities, but if we went that route, I think the ideal outcome would be be a template that could include two sets of ordered scripts.

first set would be anything that needs to happen for setup, e.g. injection of jasmine library itself, injection of anything else we need to set up, e.g. jsreporter for sauce labs.

second set would be the list of tests to run.

plenty of other possibilities. code coverage reporting is a really great initiative and it looks like you're _super_ close to making it happen!

daviferreira:
Think I got it now. Can you guys double check and merge?

P.S. util.js needs some love :love_hotel:

nchase:
This looks pretty good to me :sun_with_face: 

1. It's nice to see that we can use the `polyfills` and `vendor`  options on `grunt-contrib-jasmine` here. I didn't know about those options. Can 19f39a7b and c9b18139 be collapsed into a single commit to keep the commit history clean and atomic? I think they do the same thing, you can confirm/deny.

2. Surprised to see that utils.js is so low coverage. It looks like the majority of `insertHTMLCommand` is a fallback to execute when the `insertHTML` command isn't supported. Is that true? That is surprising (not necessarily bad, just surprised).

I'd like to get this in tomorrow or the next day. If I have free time I'll try to increase code coverage here, especially in `util.js` if I can figure out how to exercise some of the un-covered pieces.

nchase:
looks like `core.js` and `util.js` both have identical code for the `insertHTML` command. figuring out which one to use and deleting the other will be a quick win for coverage.

nchase:
next thought: anyone have a strong opinion on whether or not we should cover `util.js`?

personally I don't think it's very important to cover utility methods. my assumption is that they'll be exercised when necessary and will not necessarily be used.

daviferreira:
I think we should test the editor methods that use the utility ones.

PR #444, opened 2015-02-23
Finish extension forms

nmielnik:
This completes the refactor of turning the anchor form into a regular extension, and having the core code not do any anchor form specific logic anymore.  Now, the anchor is just another extension in the `commands` array which handles all of its logic internally in a way that any other extension could also add a form into the toolbar (ie table extension)

+ Turn `AnchorExtension` into a true extension that derives from `DefaultButton`
+ Add anchor extension into the commands array, making all handling generic so it can be reused by other extensions with have 'forms'
+ Allow for calls to `execAction('createLink', options)` to create links without using the anchor form itself.
+ Modify `execAction` to take in an argument for `createLink` and potentially other commands in the future.
+ Now, if the `anchor` button is not included in list of buttons during init, an anchor form is not created (thus the `disableAnchorForm` is no longer needed).
+ Remove the `disableAnchorForm` option
+ Added some helper methods, including `derives` to handle object inheritance.


daviferreira:
I think its time to release a new version :smile:

nmielnik:
Yeah, we will definitely need a new version, probably a major release.  I had wanted to update the documentation about the Table extension, and update the README about extensions and buttons in general...I guess that can probably wait until after the release though, since it's not a code change.

Also, do you want to get the code coverage stuff in place before we do the release?

nchase:
medium-editor has generated code coverage for a [long] while now – the big difference in the `coveralls` branch  (correct me if I'm wrong) is that davi has publicized it in the repo's README.

daviferreira:
Sure, sounds good

PR #445, opened 2015-02-23
Make onHideToolbar and onShowToolbar be passed in as options

nmielnik:
`onHideToolbar` and `onShowToolbar` were used as members of the medium-editor object, but in the documentation they were listed as options.  Updating the code to match the documentation, which is the right way to pass these callbacks into medium-editor.


nchase:
:shipit:

PR #447, opened 2015-02-23
Fix break in checkSelection for static toolbars

nmielnik:
+ Fix error introduced by #444 where clicking into contenteditable when static toolbar is enabled throws an error (`this.selectionRange` is assumed to always be written, but it may not be)
+ Add test case which verifies this fix works, and also validates some other static toolbar options


nchase:
:shipit: – I'll merge after everything passes.

PR #450, opened 2015-02-24
99% coverage in extension-anchor.js

nmielnik:
+ Added test for 'open in new window' checbkox
+ Added test for closing anchor form when user presses escape


nchase:
:fire:

PR #452, opened 2015-02-25
Attempt to fix ignored unit-tests

Wildhoney:
I've attempted to fix the unit-tests that are ignored. All tests seem to pass &ndash; although `expect(editor.elements.length).toEqual(3);` was initially set to `2` but in having a look at the unit-test below, that also has `toEqual(3)` despite disabling the toolbar. I assume the codebase changed, and as this test was ignored, wasn't kept up-to-date.

Please correct me if I'm making a gargantuous incorrect assumption :beers: &ndash; which is quite possible.

There is [another ignored test](https://github.com/Wildhoney/medium-editor/blob/master/spec/activate.spec.js#L91) but I am unsure as to why this is ignored. I didn't want to bring it back without knowing the reason why. Although this also passes.

### DIFF
--- a/spec/resize.spec.js
+++ b/spec/resize.spec.js
@@ -31,9 +31,7 @@ describe('Resize TestCase', function () {
         editor.deactivate();
     });
 
-    // I believe some other test is breaking this one, it passes when runs alone
-    // it is calling setToolbar even with no text selected
-    xit('should not call setToolbarPosition when toolbar is not visible', function () {
+    it('should not call setToolbarPosition when toolbar is not visible', function () {
         var editor = new MediumEditor('.editor');
         spyOn(editor, 'setToolbarPosition');
         fireEvent(window, 'resize');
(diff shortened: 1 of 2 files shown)


nchase:
Our CI service (travis) is experiencing some difficulties, so branches and pull requests are timing out without running all of the tests.

Whenever that's fixed (status [here](http://www.traviscistatus.com/)) we will be able to evaluate and address issues. This looks great though, looking forward to being able to evaluate this and get it merged in!

Wildhoney:
Thanks! Also, do you know why [this test](https://github.com/Wildhoney/medium-editor/blob/master/spec/activate.spec.js#L91) is ignored?

nmielnik:
No, I'm not sure why that test would be skipped, it seems like something useful to test for...was the test failing when it was run?

Wildhoney:
Yep, it works perfectly fine. [Let's re-enable it?](https://github.com/daviferreira/medium-editor/pull/461) :beers:

PR #457, opened 2015-02-25
Fix AMD "Uncaught TypeError: undefined is not a function" issue

czerasz:
Proposal fix for [this issue](https://github.com/daviferreira/medium-editor/issues/456)


nmielnik:
@czerasz do you mind merging master back into your branch and updating your pull request?  I would definitely like to get this in as soon as we can, thanks!

PR #460, opened 2015-02-26
Uses MIT license

### LINKED ISSUE #237: MIT License?
Would you consider a different license? The license as is does seem to pose some challenges with Legal situations as it is a little confusing where as MIT license is pretty widely acceptable on Legally. See: https://romanrm.net/beerware

I.E. If i fork and modify mine who gets the beer as the notice goes to the wrong author :) 

So, I think MIT license would get people using this more as the license is a little less questionable. And I'd gladly buy you a beer for this work :)

leobalter:
Closes #237

I know you're gonna miss the beer-ware license v. 42, but this is for the good of the project. 

:+1:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,3 +1,36 @@
-"THE BEER-WARE LICENSE" (Revision 42):
+Copyright Davi Ferreira, http://www.daviferreira.com/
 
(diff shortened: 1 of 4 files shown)


daviferreira:
:beers:

nchase:
:thumbsup:

PR #462, opened 2015-02-27
Use code from selection.js which is duplicated in core.js

nmielnik:
`rangeSelectsSingleNode` and `getSelectedParentElement` were in both `core.js` and `selection.js`.  This change removes the code from `core.js`
  
`getSelectedParentElement` is a useful function that I believe makes sense to leave as a helper on the medium-editor object, so I'm leaving that function there but having it just call the same method in `selection.js`.  `rangeSelectsSingleNode` is just a helper used internally by `getSelectedParentElement`, so it should be fine to  remove from the external object.


nchase:
yes!

PR #467, opened 2015-03-01
Add table-extension back into code using new extensions

nmielnik:
* Make changes to `AnchorExtension` to make it more re-usable for other form extension
* Port @CharlGottschalk 's table-extension demo and javascript back into the code base (I had removed it as part of the button/extension refactor)
* Create a working demo of another extension which accepts user input via a form in the toolbar (`table-extension.html`)


nmielnik:
Once this is in mater, I'll update the example in the wiki

PR #471, opened 2015-03-03
Update core.js

michael-gruber:
Add optional argument "range" to getSelectedParentElement().
Nate, thanks for the optional argument idea.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1071,7 +1071,7 @@ function MediumEditor(elements, options) {
         },
 
         getSelectedParentElement: function (range) {
-            if (typeof range !== 'undefined') {
+            if (typeof range === 'undefined') {
                 range = window.getSelection().getRangeAt(0);
             }
             return Selection.getSelectedParentElement(range);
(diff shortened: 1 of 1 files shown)


src/js/core.js
```diff
@@ -1070,8 +1070,11 @@ function MediumEditor(elements, options) {
             return this.options.ownerDocument.execCommand(action, false, null);
         },
 
-        getSelectedParentElement: function () {
-            return Selection.getSelectedParentElement();
+        getSelectedParentElement: function (range) {
+            if (typeof range !== 'undefined') {
```
nmielnik: should this be `typeof range === 'undefined'`?

src/js/core.js
```diff
@@ -1070,8 +1070,11 @@ function MediumEditor(elements, options) {
...
+                range = window.getSelection().getRangeAt(0);
```
nmielnik: The content window can be passed into medium-editor as an option, so you can use `this.options.contentWindow` instead of `window` here.

nmielnik:
I had some feedback, then we can merge this in.

Thanks a ton for finding this!

michael-gruber:
Yes, of course - Thank you!

michael-gruber:
sorry, Travis CI didn't like my code.
fixed it already.

PR #478, opened 2015-03-09
Factor anchor preview code out into anchor-preview.js

nmielnik:
+ Create `AnchorPreview` extension object, which is instantiated at the same time as other extensions
+ Move logic for showing, hiding, and attaching event handlers into `AnchorPreview` extension
+ Update code references to use object instead of element directly
+ Add `disableAnchorPreview` option
+ Fix bug where hovering over multiple anchors quickly caused the same anchor preview element with the same url to show up in both places (added a test case to cover the bug)
+ Cleanup event handling/detaching and break into multiple methods
+ Cleanup how anchor preview appearance was being tested


daviferreira:
:+1:

PR #481, opened 2015-03-12
Reduce number of methods exposed externally + Introduce custom events and central event management

nmielnik:
The goal of this PR is:
1) Minimize the methods exposed on the MediumEditor object to just be API methods that make sense.  This allows us to maintain a realistic set of contract methods and to have a clear definition of what constitutes a 'major' release for semantic versioning
2) Implement an standardized internal event management system, via the `Events` object, which manages all event listeners, and exposes custom events that can be used my medium-editor, helper objects, and extensions.

* Move placeholder handling code into `placeholders.js`
* Move event managing and attachment into `events.js`
  * Create custom events as a way to expose some event handling externally (`.subscribe()` method)
  * Create custom events for cases when event handler was attached to all elements inside MediumEditor
* Convert `pasteHandler` into an object that MediumEditor creates internally, which attaches its own event handlers
* Have created objects handle attaching their own event handlers (ie `Toolbar`, `AnchorPreview`)
* Hide helper methods that should not be exposed externally


daviferreira:
:heart:

PR #488, opened 2015-03-17
Expose destroy/setup as aliases for deactivate/activate methods

nmielnik:


daviferreira:
I think we should create a deprecated method wrapper, that would console.warn about deactivate/activate being deprectated and that they will be removed soon.

daviferreira:
P.S. I can do that later on :)

daviferreira:
Feel free to change the message. Also, maybe the callback call can be better, I'm kinda rusty :older_man:

nmielnik:
So we would want to support method arguments, and we can avoid actually passing the callback and just use the method name passed in as the replacement.  Something like:

```
deprecateMethod: function (oldName, newName, args) {
    // console.warn
    this[newName].apply(this, args);
}
```

This would mean we'd have to call the deprecateMethod with a specific context (ie `Util.deprecateMethod.call(this, 'old', 'new', arguments);`)

Do you think this is an improvement?

daviferreira:
Awesome thinking, can you change it? Got a meeting now :-1:

daviferreira:
Also, is breaking on IE9 I think. We should check for console maybe?

nmielnik:
Done and Done (hopefully)

nmielnik:
:+1: let's do it

PR #489, opened 2015-03-17
Clean DOM after inserting lists

### LINKED ISSUE #479: Bug when removing first item in a list
Hej, I just stumbled upon this weird little bug: When I remove the first item in a list, an unwanted `div` element is created before the list.

Steps to reproduce:
In an empty editor, write a little bit of text. Select the text and turn it into an unordered list (so that nothing is on top of that list). Now assume that you want to add something before that list. Since you cannot navigate there, one possibility is to hit *enter* when the cursor is in the first list item before the text – this will create an empty list item on the top. Then hit *backspace* on that empty list item. The cursor indentation is returned to normal and we can now write on top of the list. **However**, when checking the DOM, we see that we're now writing inside a `div` element that got silently created and shouldn't really be there.

P.S.: Thanks for the great work on the editor!

mphasize:
There is a known bug in Chrome (Webkit and Blink), when `document.execCommand` is used with `insertorderedlist` or `insertunorderedlist` and the cursor is inside a paragraph, the list get's created in the paragraph, which results in invalid HTML.

This change watches for the `insertorderedlist` and `insertunorderedlist` actions and unwraps the created list from it's paragraph element.

Todo:
- [x] unwrap newly created list elements
- [x] fix #479 
- [x] write a test
- [ ] when an existing list is reversed into normal text, it should be wrapped in a paragraph again


mphasize:
The test run failed on sauce labs – no idea what I could do about that.

nmielnik:
Yeah, don't worry about the sauce labs issues. sauce labs and travis have a known issue where only pull requests from a branch within the main repo are allowed to run in saucelabs.

We haven't decided how we want to work around this, but as long as the tests pass against your branch it's ok.

mphasize:
As far as I can tell, the changes fix the 2 bugs I found in Chrome and it works great for me.
@nmielnik Any chance, you could help with writing some tests for this to prevent future regressions?

nmielnik:
I will try to take a look as soon as I get some time

nmielnik:
@mphasize would you be able to give me permissions on your fork so I propose some changes?

mphasize:
@nmielnik Sorry, a bit late, but you've got access now.

nmielnik:
@mphasize I made some small changes and added a few tests.  Do you want to look over the changes and make sure everything is accurate with what you intended?

Also, I think you should open your last checklist item as a separate issue.  It seems like a more complicated problem to solve, and I wouldn't want to hold up the working changes we have right now if we don't need to.

mphasize:
@nmielnik Sorry I was not able to respond earlier, but the changes look great and having test coverage for this is awesome! Thanks for taking up the work and merging it!

PR #491, opened 2015-03-19
add figcaption exception in paragraphCreation

jchamois:
Hello,

I Added an exception in the condition of  paragraphcreation in order to use the  figcaption tag properly. 
This fix is coming form the common use of medium insert plugin and medium. 
When you set your caption with an image, the tag figcaption is used, end each enter keypres create a p tag into it :-/

Thanks

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -408,7 +408,7 @@ function MediumEditor(elements, options) {
                     editorElement = meSelection.getSelectionElement(self.options.contentWindow);
 
                     if (!(self.options.disableReturn || editorElement.getAttribute('data-disable-return')) &&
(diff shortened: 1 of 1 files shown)


daviferreira:
Hey @jchamois, you had a really old version of MediumEditor. Could you update your PR with the latest master?

Another good thing would be to create a ```disableEnterOnTags``` option or variable instead of comparing directly, but I can do that later.

Cheers!

nmielnik:
@jchamois it would be pretty cool if you added a test as well to make sure no one breaks your changes in the future :smile: 

your choice though, we can take your changes either way (once you update your version of MediumEditor)

daviferreira:
@jchamois ping

PR #494, opened 2015-03-20
Implements font sizing control

### LINKED ISSUE #376: Implement font sizing control

secrettriangle:
Closes #376


README.md
```diff
@@ -85,7 +85,7 @@ var elements = document.querySelectorAll('.editable'),
 
 ### Toolbar options
 * __activeButtonClass__: CSS class added to active buttons in the toolbar. Default: 'medium-editor-button-active'
-* __buttons__: the set of buttons to display on the toolbar. Default: ['bold', 'italic', 'underline', 'anchor', 'header1', 'header2', 'quote']
+* __buttons__: the set of buttons to display on the toolbar. Default: `['bold', 'italic', 'underline', 'fontsize', 'anchor', 'header1', 'header2', 'quote']`
```
nmielnik: In the spirit of having this look like medium by default, we probably should have this as a default button.

There is a section for listing additional buttons further down in the documentation, so you should definitely add this into that list.

nmielnik:
Thanks a ton for this, pretty exciting to finally have a built-in font size control.

Can you add some tests?  We have 99% coverage for the anchor-extension, so it shouldn't be too hard to get a high amount of coverage for this file with a few tests.

dist/js/medium-editor.js
```diff
@@ -2152,6 +2152,229 @@ var AnchorPreview;
...
+                close = doc.createElement('a'),
+                save = doc.createElement('a'),
+                target,
+                target_label,
+                button,
+                button_label;
+
+            // Anchor Form (div)
+            form.className = 'medium-editor-toolbar-form';
+            form.id = 'medium-editor-toolbar-form-anchor-' + this.base.id;
```
nmielnik: This id will conflict with the id for the ancho form.  Can you make it something like `'medium-editor-toolbar-form-font-size-' + this.base.id`?

dist/js/medium-editor.js
```diff
@@ -2152,6 +2152,229 @@ var AnchorPreview;
...
+            close.innerHTML = this.base.options.buttonLabels === 'fontawesome' ?
+                              '<i class="fa fa-times"></i>' :
+                              '&times;';
+            form.appendChild(close);
+
+            // Handle close button clicks
+            this.base.on(close, 'click', this.handleCloseClick.bind(this));
+
+            // (Optional) Add 'open in new window' checkbox
+            if (this.base.options.anchorTarget) {
```
nmielnik: Both this and this next if conditional were specific to the anchor form.  You can just remove these since the `options.anchorTarget` and `options.anchorButton` should be used by this extension.

daviferreira:
Awesome stuff :+1: , let us know if you need help writing tests

oxhey:
I like the idea , but i dont think that a slider is the best way to do it . I think a drop down with a list of fonts would be better and easier to use for people who are familiar with other text editors

secrettriangle:
Looks like PhantomJS is doing some weird stuff. @daviferreira mentioned @nmielnik might be the guy to talk to about this -- @nmielnik, any thoughts?

phiggins42:
looking at the code quickly it appears to be a lot of overlap functionality with the Anchor extension?  Some of the functions are exact copies of the other. others are only modified slightly. Would be ideal to make this a `derives` of built in Anchor button ...

nmielnik:
@secrettriangle would you be able to give me access to your fork so I can try to help out with the tests?

secrettriangle:
Done. Please avoid amending if possible so I can see what you changed!

phiggins42:
https://github.com/phiggins42/medium-editor/commit/470a89f8f3bccb524f7ff0c9428dba3272824952 is my $0.02

phiggins42:
@nmielnik @secrettriangle accidentially realtime coding along with you in https://github.com/phiggins42/medium-editor/tree/issue-376-b

nmielnik:
@phiggins42 I agree if all of this works, there definitely needs to be some use of `derive` here to avoid all the code duplication (including some which is actually wrong and causing some issues).

However, after messing with the tests, there are definitely some issues I ran into that should be addressed:

1) On my mac, when the code calls `execCommand('fontsize',...)` it creates `<font class="Apple-style-span">` which does not get removed when calling `clearFontSize()`.  It seems like, if anything, we should probably remove the `<font>` tag completely instead of just removing the `size` attribute.

2) Since the slider is an input that can have focus, that means focus is stolen away from the contenteditable element.  This will likely cause some cross-browser issues, especially since we're triggering the `execCommand()` call via the `change` event.  The anchor-extension gets around this saving the selection before the form is displayed, allowing the user to edit the inputs, and then once 'save' is clicked, restoring the selection and calling `execCommand()`.  Maybe we should wait until ok is clicked before we actually increase/decrease the font size?

Until we work around this and update the test to respond accordingly, the current tests will fail.

phiggins42:
1) the Apple-style span unwrapping exists in the paste handler perhaps? I've seen that referenced internally before. Really, "clearFontSize" on cancel seems odd? If it didn't respond realtime no font size would be set until explicitly saved, and the new removeFormat option _ideally_ would remove fontsize too? though size/style overlapping might be confusing ... 

2) I haven't actually tried this UX yet at all, just happened to notice the bulk of the extension was identical to Anchor. I see how focus/selection-restoration could get wonky.

nmielnik:
yeah, we might want to try out the way of not updating fontsize in real time, and just waiting until the form is hidden to update.  If that UX seems alright, then we'll be even closer to behaving the same as the anchor form, in which case we'll have even more reason to `derive` from it instead of `DefaultButton`.

secrettriangle:
I've incorporated @nmielnik's fixes and fixed the final remaining unit test. Please take a look. Hopefully we'll get this merged soon!

PS: @phiggins42 I took a look at your `issue-376-b` branch -- I think it's best we wait for a solution to #519 instead extending this from anchor. What do you think?

phiggins42:
I'm inclined to say any pure-c/p functions should be at _least_ a prototype reference and ideally use util.derives until #519 solidifies.  Can just as easily be done in follow up though. There isn't an excuse for carrying around exact dupe functions internally.

nmielnik:
Looks like there are some test-failures in other browsers:

** IE: **
screen shot 2015-03-30 at 9 31 26 am

** Firefox: **
screen shot 2015-03-30 at 9 30 40 am

nmielnik:
@secrettriangle I was able to fix the case that was failing in firefox (which was also the cause of one of the ie failures).

Looks like there is still that 1 last failure in ie, the `should revert font size when slider value is set to 4` test case doesn't seem to work in IE, as the change from '1' to '4' doesn't work, and the expectation that the 'size' attribute will be removed doesn't seem to be true.

nmielnik:
The IE failure is still there:

screen shot 2015-04-09 at 5 27 16 pm

stefanosala:
I would love this feature! Anything that I can do to help?

PR #495, opened 2015-03-21
Turn of saucelabs running on pull requests (for now)

nmielnik:


j0k3r:
Is there a reason why we didn't enable it again? Why did you disabled it at first?

In fact, it's really useful for people who create PR (since when we, _owners_, create a new branch, saucelabs is enabled).

It could be very useful (at least) for #820.

nmielnik:
So SauceLabs had implemented their infrastructure in a way that only pull-requests directly against master to be able to run.  Thus, any PR that includes an external fork is prevented from running against SauceLabs.  The result was that PRs from non-owners always failed and had the red X next to them.

The idea that people have had is having an 'integration' that we maintain, that we can merge PRs from external forks into.  We can then trigger SauceLabs to run against that branch, and if all the tests pass, we can then merge that branch into master.

We could also try enabling this again, but I don't think it will work.  There may be some new workarounds or other ways to accomplish this, it's been a while since I looked into it.

PR #500, opened 2015-03-24
Fix bug that happened when trying to click out of the contenteditable in Chrome

patrick-webs:
In Chrome only, with a certain set of options, there was a bug that wouldn't allow the user to click out of the medium-editor when the caret was positioned right at the end of a text node. This is because the selection would be "standardized" whenever the blur event was called, which resulted in the contenteditable getting re-focused.

    { staticToolbar: true,
     standardizeSelectionStart: true,
     updateOnEmptySelection: true }

This PR fixes the issue.


patrick-webs:
@daviferreira I updated to use `this.options.contentWindow`. Thanks!

PR #506, opened 2015-03-26
Anchor form

phiggins42:
a handful of non-api-changing tweaks for extensibility purposes. commit messages describe the bulk of the changes. basically just brings anchor extension "lifecycle" more in-line with anchor-preview lifecycle (getTemplate, getForm, etc) and breaks things into smaller chuncks to allow overriding parts externally.


src/js/extension-anchor.js
```diff
@@ -51,6 +55,42 @@ var AnchorExtension;
             return this.form;
         },
 
+        getTemplate: function () {
+
+            var template = [
+                '<input type="text" class="medium-editor-toolbar-input" placeHolder="', this.base.options.anchorInputPlaceholder, '">'
```
nmielnik: I don't believe `placeholder` should be camelCased, was that intentional?

src/js/extension-anchor.js
```diff
@@ -24,6 +24,10 @@ var AnchorExtension;
 
         // Button and Extension handling
 
+        // labels for the anchor-edit form buttons
+        formSaveLabel: '&#10003;',
+        formCloseLabel: '&times;',
```
nmielnik: I wonder if we should make the `fa` labels be exposed here as well for consistency. Thoughts?

src/js/util.js
```diff
@@ -96,8 +96,8 @@ var Util;
             return !!(obj && obj.nodeType === 1);
         },
 
-        now: function now() {
-            return Date.now() || new Date().getTime();
+        now: Date.now || function now() {
+            return +new Date();
```
nmielnik: Thanks for finding and fixing this.  I think you were right when you said we can just use `Date.now()` safely in all our browsers, this was copy and pasted from underscore js so we were probably being safer than we needed to be.

src/js/extension-anchor.js
phiggins42: they actually still do use the buttonLabels===fontawesome path, I did this this way to prevent potentially regressing anyone. these two values were simply fallback/defaults in the template rendering, so I exposed them so I could override personally in a non-fontawesome path. if you are setting buttonLabels: "fontawesome" you likely have never seen the `&times;` anyway ... so to answer your question, the fa-labels are preserved by virtue of this just being extracting of a fixed value into a property.  I had thought about moving the buttonlabels===fa check into init and building up formSaveLabel/formCloseLabel one time, but the complicated things by needing to check for a value prior, and yadda yadda yadda ...

src/js/util.js
phiggins42: easy enough to remove them all ... either way this snippet is "fine" to land as it fixes a bug in the fallback implementation, and can later be removed outright.

src/js/extension-anchor.js
phiggins42: hmm, i've always just kind of written it as camelCase ... never had an issue where a browser didn't figure it out, but can lowercase for consistency if desired.

phiggins42:
is there a way to indicate "private but exposed API". the coverage only decresed becaused I broke a function into 3 smaller identical functions, but the 2 I had intended to be "internal implementation" (eg: `_privateFunction:` ...

nmielnik:
We don't have a way to mark something as private but exposed, at least not in a meaningful way that would affect something like code coverage.  

I bet the code coverage reduction is likely because of the changes you made to the 'now()' function, since the 2nd part of the OR will never be run, and it was increased in size slightly.  I bet removing the OR would get coverage back up.

src/js/extension-anchor.js
phiggins42: it speaks to the oddity that is `options.buttonLabels` having that special "fontawesome" path everywhere. You either want to override every individual label, or use the entirety of the pre-defined values for 'fontawesome' icon set.  not sure the best balance there?

phiggins42:
haha you really don't like that unused `+new Date()` huh? +1 me doing a global s/r for Util.now()/Date.now()?

src/js/extension-anchor.js
nmielnik: Yeah, you're right, figuring out how to do the fontawesome labels is a bigger issue than just this, just wanted to get your thoughts.

nmielnik:
Well ever since you pointed it out via email, it's been bugging me, especially since it's my fault that it's there :smile: 

But from my work trying to get our code coverage up, a -0.15% decrease is most likely just one line of code, and I can't see anywhere else in your changes that would have added a new line of code that wouldn't be covered.  

It's my best guess as to why this was lowered. I would expect the rest of your changes to be fully covered by our tests, so your changes in extension-anchor should have actually made the coverage go up (since 100% coverage of newly added code should raise our %).

phiggins42:
how do I trigger the coverage/coveralls locally? hate to just keep throwing commits up to wait for a coverage count to come back.

daviferreira:
@phiggins42 ```coverage/index.html```

nmielnik:
If you run `grunt` on its own, that will run all the tests and run the coverage.

You can check the coverage number by bringing this page up in your browser:

`<root>/reports/jasmine/lcov/lcov-report/index.html`

daviferreira:
We should remove one of those btw :smile:

nmielnik:
Who doesn't love multiple options to look at the same thing? :smiling_imp:

phiggins42:
```
Execution Time (2015-03-26 13:45:11 UTC)
loading tasks        964ms  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 12%
jslint:client        376ms  ▇▇▇▇▇▇▇▇▇▇ 5%
jasmine:suite         4.8s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 62%
uglify:build          1.2s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 15%
autoprefixer:themes  105ms  ▇▇▇ 1%
csslint:strict       154ms  ▇▇▇▇ 2%
Total 7.7s
```

`grunt` alone did not run coveralls, tried `grunt coveralls`, but timed out:

```Running "coveralls:dist" (coveralls) task
>> Failed to submit 'reports/jasmine/lcov/lcov.info' to coveralls: Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
>> Failed to submit coverage results to coveralls
Warning: Task "coveralls:dist" failed. Use --force to continue.```

maybe I need to setup an account or something?

phiggins42:
`92.94` -- oh github beat me to it.

nmielnik:
We can cut a new release today

Issue #505, opened 2015-03-25
Feature request: 'remove formatting' button


courtenay:
This would help in cases where you don't know what the format is to toggle it, or where there's existing block elements (like a PRE) where selecting 'pre' again just nests instead of removing format.

(If there's an easy way to do this, let me know! otherwise I'll probably just implement it myself in a week or three)

nmielnik:
There just so happens to be native support for [this](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) via `document.execCommand('removeFormat', false, true)`.

Because we can leverage `document.execCommand`, this might be as simple as just adding an additional entry into the `ButtonsData` object in `button.js`.

You can see all the existing buttons [here](https://github.com/daviferreira/medium-editor/blob/master/src/js/button.js#L9)

If you add another object into there that looks something like this:
```js
'removeFormat': {
            name: 'removeFormat',
            action: 'removeFormat',
            aria: 'remove formatting',
            contentDefault: '<b>?something?</b>',
            contentFA: '<i class="fa fa-(some-icon)"></i>'
        },
```

I'm not sure if it's possible for the button to be "active" in the toolbar, or if we'd even want it to be.  If we did, you might try setting `useQueryState` to `true` and see if the browser will tell you when a selection has already had its formatting removed..but the more I think about it, the less I think it would be useful.

If you add this in, you should be able to just pass another button name in when you create the MediumEditor instance, and the button should show up:

```js
var editor = new MediumEditor('.editor', {
    buttons: ['bold', 'italic', 'underline', 'removeFormat']
});
```

If you find time to add it, make sure you add some tests! :smile:

phiggins42:
verified that:

```
function RemoveFormat(){
        this.parent = true;
        this.options = {
            name: 'removeFormat',
            action: 'removeFormat',
            aria: 'remove formatting',
            contentDefault: '<b>X</b>',
            contentFA: '<i class="fa fa-eraser"></i>'
        };
        this.name = 'removeFormat';
    }

    return Util.derives(Button, RemoveFormat);
```

creates an extension for this that mostlyworks. does not unlink links. works across multiple paragraphs. removes sup/sub/b/i, does not un-head headings.

nmielnik:
Thanks @phiggins42 for checking that out.

Do you think, based on the behavior you saw, this is worth adding into the codebase?  From the way you describe it, it seems ok that it doesn't unlink links, the header stuff may be a bit annoying though...

[...]

PR #511, opened 2015-03-27
Extension removeformat

phiggins42:
refs #505


daviferreira:
Nice stuff :+1:

daviferreira:
P.S. They are

phiggins42:
hmm, looks like this breaks the ie's :/

nmielnik:
Yeah, one of the joys of contenteditable is when you start adding attributes (`id`, `style`, etc.) into HTML elements inside a contenteditable, IE will generally never actually remove that element.

Generally to workaround things like this, we try to use a regex in the test case, assuming the result from various browsers is still valid.

I'm looking into fixing this so we can cut another release.

steoo:
I made this gist, check it out
https://gist.github.com/steoo/6aa45e6c6d99974fbd2d

PR #530, opened 2015-04-03
Fix pasteHTML call from cleanPaste: add base options to arguments

Sergiy-Babich:
Related to https://github.com/daviferreira/medium-editor/issues/526

### DIFF
--- a/src/js/paste.js
+++ b/src/js/paste.js
@@ -113,7 +113,7 @@ var PasteHandler;
                 // double br's aren't converted to p tags, but we want paragraphs.
                 elList = text.split('<br><br>');
 
-                this.pasteHTML('<p>' + elList.join('</p><p>') + '</p>');
+                this.pasteHTML('<p>' + elList.join('</p><p>') + '</p>', this.base.options);
 
                 try {
                     this.options.ownerDocument.execCommand('insertText', false, "\n");
@@ -145,7 +145,7 @@ var PasteHandler;
                     }
                 }
             } else {
-                this.pasteHTML(text);
+                this.pasteHTML(text, this.base.options);
             }
         },
 


src/js/paste.js
```diff
@@ -145,7 +145,7 @@ var PasteHandler;
                     }
                 }
             } else {
-                this.pasteHTML(text);
+                this.pasteHTML(text, this.base.options);
```
nmielnik: What was the intended result of passing this into `pasteHTML()`?  As far as I can tell, nothing in `pasteHTML` actually uses any of the values expose from `this.base.options`

phiggins42: It would allow you to define the clean attar options via editor props

nmielnik: Ah ok.  In that case, can we make a sub-object being something like `options.pasteOptions`?  It would be good if the options passed into here were scoped down to only contain the stuff we care about (instead of the entire options object).

Also, you can just use `this.options` instead of `this.base.options` since the options are stored during initialization.

phiggins42: agree. I saw this ticket come in and wanted to suggest the exposure of those options be in a "paste" member (like my doc alludes to) ... eg:

```
new Editor(...., {
     extensions: {
          "paste":{
                 filterRegexp: [ /* additional regexps desired */ ],
                 cleanTags:[....], whateverTheOptionNameIs:[]
          }
     }
});
```

nmielnik:
The changes you proposed should now be implemented as part of #535.  If there should be additional features, please open another PR or another issue.  Thanks again for this and I apologize for implementing an alternate way of achieving the same thing, we had an interesting discussion and wanted to do a bit more with the changes you were suggesting.

PR #531, opened 2015-04-03
Add pasteReplacements options for custom RegExps

Sergiy-Babich:
https://github.com/daviferreira/medium-editor/issues/529

### DIFF
--- a/src/js/paste.js
+++ b/src/js/paste.js
@@ -103,8 +103,7 @@ var PasteHandler;
             var i, elList, workEl,
                 el = Selection.getSelectionElement(this.options.contentWindow),
                 multiline = /<p|<br|<div/.test(text),
-                replacements = createReplacements();
-
+                replacements = createReplacements().concat(this.options.pasteReplacements || []);
             for (i = 0; i < replacements.length; i += 1) {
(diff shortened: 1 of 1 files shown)


nmielnik:
This is good stuff, do you mind updating README.md about this new option?

phiggins42:
ideally, before introducing this as a public config we might take steps to breaking out config values into the groups (eg: 'paste': { replacements:[ ... ] }) or something?

nmielnik:
Yeah, @Sergiy-Babich seems like this as well as #530 could be combined into a single option group like

```js
var editor = new MediumEditor('.editor', {
    paste: {
        cleanReplacements: [ ... ],
        cleanAttrs: [ ... ],
        cleanTags: [ ... ]
    }
}
```

That would be a good start and a clean way to, and should be pretty close to where we land with the overall options changes @phiggins42 has proposed.

phiggins42:
yah if we introduce new config props as newly-grouped-type-things we have to deprecate less as we move others to the groupings.

nmielnik:
This should now be supported as part of the changes for #535.  If you need additional support please propose another PR or open an issue.

Thanks for bringing this up!

Issue #486, opened 2015-03-16
toolbar flashing with multiple instances of medium-editor on page


nmielnik:
When multiple instances of medium-editor exist on the page, various toolbars appear and disappear quickly when clicking in areas of the page that are outside of the contenteditable sections.  This can be seen here by clicking on the far left edge of the page multiple times.

http://daviferreira.github.io/medium-editor/static-toolbar.html

daviferreira:
Sick FX :zap:

PR #532, opened 2015-04-03
Fix toolbars behaviour after right-click 

Sergiy-Babich:
Fixed toolbar show on all editors on page after right click on any place of page.

To reproduce you must have at least two editors on page. Both should have 
```javascript
stickyToolbar: true
staticToolbar: true
```
Then click on first editor to show toolbar. After click on second editor (first toolbar hides).
Then right click on any place of document and both toolbars should appear.

image

### DIFF
--- a/src/js/toolbar.js
+++ b/src/js/toolbar.js
@@ -154,7 +154,7 @@ var Toolbar;
             // Do not trigger checkState when mouseup fires over the toolbar
             if (event &&
                     event.target &&
-                    Util.isDescendant(this.getToolbarElement(), event.target)) {
+                    !Util.isDescendant(this.getToolbarElement(), event.target)) {
                 return false;
             }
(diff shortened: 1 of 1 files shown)


nmielnik:
I'm guessing this is probably related to #486 

Looks like there are a lot of test failures, this suggested fix seems to break a bunch of the other core flows.

I've been working on a fix for this as well as some other toolbar issues in a separate branch.  Based on my work so far, I'm guessing the fix for this issue and for #486 require a bit more than one line of code, but if you find another way around it that doesn't break our tests or the other demos, please update this pull request :smile:

nmielnik:
I believe #550 should fix the issue you've called out here, I'm going to close this PR in the meantime.  Please let me know if the issue is not resolved and we can try tackling a fix.

PR #533, opened 2015-04-04
Add textarea support, fixes #103

### LINKED ISSUE #103: Allow using a <textarea> instead of a <div>
Most WYSIWYG editors use `<textarea>`, but since this one uses `<div>`, it's very difficult to integrate with existing CMSes (such as Drupal). Any chance `<textarea>` support could be added?

daviferreira:
Native basic textarea support.


nmielnik:
:+1:

demo/textarea.html
```diff
@@ -0,0 +1,24 @@
...
+    <link rel="stylesheet" href="css/demo.css">
+    <link rel="stylesheet" href="http://netdna.bootstrapcdn.com/font-awesome/4.0.3/css/font-awesome.css">
+    <link rel="stylesheet" href="../dist/css/medium-editor.css">
+    <link rel="stylesheet" href="../dist/css/themes/default.css" id="medium-editor-theme">
+</head>
+<body>
+    <a href="https://github.com/daviferreira/medium-editor" class="github-link"><img style="z-index: 100;position: absolute; top: 0; right: 0; border: 0;" src="https://s3.amazonaws.com/github/ribbons/forkme_right_gray_6d6d6d.png" alt="Fork me on GitHub"></a>
+    <div id="container">
+        <h1>Medium Editor</h1>
+        <textarea class="editable"><p>Textarea is now supported</p></textarea>
```
phiggins42: it should be &gt; etc no?

src/js/core.js
```diff
@@ -268,11 +268,33 @@ function MediumEditor(elements, options) {
...
+        var textarea;
+        div.innerHTML = this.elements[index].value;
+        div.setAttribute('data-textarea', id);
+        this.elements[index].classList.add('medium-editor-hidden');
+        this.elements[index].setAttribute('data-textarea', id);
+        this.elements[index].parentNode.insertBefore(
+            div,
+            this.elements[index]
+        );
+        textarea = this.options.ownerDocument.querySelectorAll('textarea[data-textarea="' + id + '"]')[0];
```
phiggins42: probably don't have to go re-lookup this? just get a ref before you replace the `elements[index]` no?

src/js/core.js
```diff
@@ -268,11 +268,33 @@ function MediumEditor(elements, options) {
...
+        this.on(div, 'input', function (e) {
+            textarea.value = e.target.innerHTML;
```
phiggins42: may want to pass through serialize, in case anyone using any extensions which modify data in/out

daviferreira: Not sure about this one since all serialize does is get the innerHTML and save it to a object.

phiggins42: right but it's a public hook that people rely on. in the original ticket there was at least one mention of using `medium-instert`, which clogs up the contentEditable innerHTML with divs that it removes in an overriden serialize function

daviferreira: OK, convinced me :smile:

bborn:
Because `initCommands` is called before `initElements` some event listeners aren't properly attached to the contenteditable div (instead they're attached to the textarea, which is then hidden). Is there any reason why the `initElements` stuff shouldn't happen first?

daviferreira:
Hmm, not sure about this. Nice catch. Ping @nmielnik

nmielnik:
First of all, this is a great catch @bborn so thanks!

The reason why we're doing things in this order is that we create the Toolbar inside `initElements()`.  The toolbar will look through all the extensions that implement a `getButton()` and call it so that they can be inserted into the toolbar.

Extensions assume that `.init()` is called before `.getButton()`, so this is why `initCommands()` (which calls `.init()` on each extension) must be called before `initElements()`.

I believe the solution here is to extract the toolbar initiating code out of `initElements()` into its own `initToolbar()` method.  Once that's done, we can change the `setup()` function to this:

```js
setup: function () {
  ...
  // Call initialization helpers
  initElements.call(this);
  initCommands.call(this);
  initToolbar.call(this); // New method we've created
  ...
}
```

If anyone has a chance to open a pull request for this in the meantime, that'd be great!

nmielnik:
Also, @bborn will you open a separate issues for this?

Issue #529, opened 2015-04-01
Add option for custom clean RegExp


Sergiy-Babich:

phiggins42:
would #528 work for both our cases?

Sergiy-Babich:
Some kind of. If this should allow add custom RegExp for additional cleaning that should be OK.

phiggins42:
I can see how well-hidden that list of regexp's is :) would something like:

```
new Editor(..., { 
      pasteReplacements:[
           [new RegExp(), ""]
      ]
});
```

and then at https://github.com/daviferreira/medium-editor/blob/master/src/js/paste.js#L106

make it `replacements = createReplacements().concat(this.options.pasteReplacements || [])`

work? #528 wants to expose each of the replacements as "filters" (or some other word), as well as allow custom `function clean(text){ return cleaned(text) }` pass-through for content.

Sergiy-Babich:
Yes, that is exactly what I want! Great.

nmielnik:
This is now supported as of #535

PR #535, opened 2015-04-06
Paste options

nmielnik:
* Implement fixes for #529, #530 & #531 in a way that aligns with future options model
* Introduce nested, extension-specific options (paste options are now a set of properties a layer deeper inside the options object)
* Add more paste tests + tests for overriding `pasteHTML()` and `cleanPaste()` methods
* `defaults()` and `extend()` work
  * Create `extend()` method
  * Expand support for to handle dynamic list of arguments
  * Add tests targeting these methods specifically


nmielnik:
@phiggins42 @daviferreira @Sergiy-Babich thoughts?

nmielnik:
So I'm definitely strongly against any huge refactors which mean other things can't go on in the meantime.  There's definitely a way to approach changing the way we pass options as well as converting things to use extensions in small iterations and not in one shot.

Why does it need to be such a sweeping change?  If these limited changes can go in via one shot, why can't you changes be scoped down to this as well?

phiggins42:
> So I'm definitely strongly against any huge refactors which mean other things can't go on in the meantime. There's definitely a way to approach changing the way we pass options as well as converting things to use extensions in small iterations and not in one shot.

no i agree. was just being vocal about a fun WIP/thing I had inflight. I have in fact done a huge sweeping refactor that provides the new syntax/functionality while providing a deprecation path forward through to the next major bump. my vocal-ness was in regards to introducing the pattern of the "new style" as a single one-off when I am "so close" to pulling off working POC before anyone has seen how the new options style would "really" look (as per my various brain dumps and discussions ... again, this newoption style is still just a brainstorm/proposal and has never seen the light of day)

> Why does it need to be such a sweeping change? If these limited changes can go in via one shot, why can't you changes be scoped down to this as well?

it is the pattern overlap. the stuff i've done assumes none of the extensions/settings are intendended to be passed as deep objects, so this introduces an inconsistency in the current version and does so in spite of the yet-to-be-seen experimental "new way" as proposed. I am not against new features landing, forward-porting the properties and providing a deprecation pass is already working for those style options ... but introducing this new style way of passing options conflicts directly with what I have in progress (which I'm 24 tests away from maintaining CI back-compat ... and likely nitpicking on some of those tests) which accomodates this feature in both option forms.

tldr- i'm against introducing any "new" way of defining config values and significant pattern changes while i am actively attempting to get a proposal/working-prototype of a new pattern for us to play with.

daviferreira:
LGTM :+1:

PR #538, opened 2015-04-07
use forEach's 3rd arg. convert some eaches to somes

phiggins42:


nmielnik:
Thanks, there are definitely other places in the code we should be using `some()`, it's nice to know about these things.

PR #539, opened 2015-04-07
add an Extension thinger. nothing uses it yet.

phiggins42:


nmielnik:
Can you modify anchor-preview to use `Extension.extend`?

dist/js/medium-editor.js
```diff
@@ -803,6 +803,129 @@ var Util;
...
+
+    Extension.prototype = {
+        init: function(/* instance */){
+            // called when properly decorated and used.
+            // has a .base value pointing to the editor
+            // owning us. has been given a .name if no
+            // name present
+        }
+        // core API definition for an "extension" follows:
+    };
```
nmielnik: If you have anchor-preview use this, can you add `.deactivate()` in for now as well?

phiggins42: since nothing is using it yet the actual "core extension" api is intentionally empty (easier to add things than deprecate). The plan of attack here is to put the tooling necessary to convert the things over one by one (though i've already done them all en masse, and this is just an exercise in c/p) (this is why i've exposed that .extensions as empty, as default built-in extensions are made they will be added there and deprecated from .statics).

src/js/extension.js
```diff
@@ -0,0 +1,73 @@
+var Extension;
+(function(){
+
+    /* global Util */
+
+    Extension = function (options) {
+        Util.extend(this, options);
```
nmielnik: For this first pass with converting things over to be extensions and use this base class, can we just make `.options` be a member variable containing the options? It'll make converting over to using this model simpler for the first pass and once we have all of the extensions made we might have different needs/requirements for how we handle options passed into extensions.

phiggins42: fair enough, I guess. though -- the success of this shift (and all the regression work i've done) kind of relies on getting rid of the explicit "options" object for the individual things. the thinking there was: people are either passing new options or just wanting the defaults. by defining each extensions's "defaults" on the prototype you eliminate 1 level of nested objects, provide a mindlessly simple way to either pass options or extend the reference directly.  incidentially the root cause of me proposing any of this way the inconsistency in where options were read from and how they were passed, and this is kind of a pivotal moment/decision ... if the fundamental change of going from flat-options to nested-options also requires an explicit .options object to pass around nothing has been bettered and I should just stop this line of thinking all together (though once you see the end result it becomes way clearer)

phiggins42: (take Anchor for example. some of the form label options were defined on the .prototype of the anchor extension, and the rest of the options are read from the core options hash + defaults. for consistency one would either move those prototype props into a default "options" object, and migrate the core options to that, or "just migrate core options to the extension prototype". the latter way means all the defaults and definitions are defined right there at the extension in a single place, and "justworks" for preserving defaults through prototypical inheritance rather than carrying around an object literal you have to work to mix and overwrite (.options) ... maybe it's a personal preference, but I see no real distinction between protoprops and an explicit opions object, and would go so far as to say the options object path makes things more complicated.

PR #548, opened 2015-04-10
Justify buttons should use style, not queryCommandState

patrick-webs:
In Firefox and IE11, queryCommandState('justifycenter') ignores a
text-align: center CSS style on a parent element.

See https://bugzilla.mozilla.org/show_bug.cgi?id=748315

Removing the "useQueryState: true" uses medium's built in comparison of the style prop using window.getComputedStyle instead. This takes into account "text-align: center" CSS on parent elements.


patrick-webs:
new test failed in IE11, oops :-( closed until I fix it.

PR #549, opened 2015-04-10
When pasting text without linefeeds in FF, it was still wrapping text

greg-heidel:
in a `<p></p>`. In FF, when pasting this text in a parapgraph, it would
result in 3 paragraphs.

For example:

This is some sample text.	// copy and paste sample right after sample

Expected:

This is some sample sample text.

Actual:

This is some sample

sample

text.

### DIFF
--- a/spec/exploits.spec.js
+++ b/spec/exploits.spec.js
@@ -29,7 +29,7 @@ describe('Exploits', function () {
             test = {
                 source: 'img onerror handler',
                 paste: '><img src="x" onerror="alert(\'xss\')">',
-                output: '<p>&gt;&lt;img src="x" onerror="alert(\'xss\')"&gt;</p>'
+                output: '<span id="editor-inner">&gt;&lt;img src="x" onerror="alert(\'xss\')"&gt;</span>'
             };
 
         // mock event with clipboardData API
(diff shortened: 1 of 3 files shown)


nmielnik:
Could you add/modify a test to handle pasting of more than 1 newline (the other part of the code branch you added)?

daviferreira:
Let us know if you need any help with that :)

greg-heidel:
Will do…thx

> On Apr 10, 2015, at 3:55 PM, Nate Mielnik <notifications@github.com> wrote:
> 
> Could you add/modify a test to handle pasting of more than 1 newline (the other part of the code branch you added)?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/daviferreira/medium-editor/pull/549#issuecomment-91686104>.
> 


-- 
Please NOTE: This electronic message, including any attachments, may 
include privileged, confidential and/or inside information owned by Demand 
Media, Inc. Any distribution or use of this communication by anyone other 
than the intended recipient(s) is strictly prohibited and may be unlawful. 
If you are not the intended recipient, please notify the sender by replying 
to this message and then delete it from your system. Thank you.

PR #553, opened 2015-04-11
Add custom editableInput event

nmielnik:
* create custom event which fires whenever a `contenteditable` element monitored by MediumEditor has a value change (including toolbar triggered changes)
* attach via 
```javascript
editor.subscribe('editableInput', function (event, contentEditableElement) { ... } );
```

More documentation on attaching custom events will be coming :wink:


nmielnik:
Oh if only IE supported the `input` event for contenteditable...closing this until I find a workaround for IE

PR #554, opened 2015-04-12
fix BC break warn for cleanPastedHTML option

adrianorsouza:
I noticed after migrating from v3.0.9 to v4.x the option name `cleanPastedHTML` has been deprecated and should be used `paste.cleanPastedHtml` instead. it turns out that this deprecated option is not displaying warning in the console and also when passing the newly option `paste.cleanPastedHtml` the content does not get cleaned as expected.

This PR addresses to:
- fix missing deprecated warning for passing the option `cleanPastedHTML` used in v3.0.9 and earlier.

- fix `cleanPastedHtml` option. As of v4.0.0 this option is defined in camelCase format when this options is passed along with `paste` option it does not clean up the pasted html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ var editor = new MediumEditor('.editable', {
     paste: {
         // This example includes the default options for paste, if nothing is passed this is what it used
         forcePlainText: true,
-        cleanPastedHtml: false,
+        cleanPastedHTML: false,
         cleanReplacements: [],
         cleanAttrs: ['class', 'style', 'dir']
         cleanTags: ['meta']
@@ -128,8 +128,8 @@ var editor = new MediumEditor('.editable', {
 ```
 
 * __forcePlainText__: Forces pasting as plain text. Default: true
-* __cleanPastedHtml__: cleans pasted content from different sources, like google docs etc. Default: false
-* __cleanReplacements__: custom pairs (2 element arrays) of RegExp and replacement text to use during paste when __forcePlainText__ or __cleanPastedHtml__ are `true` OR when calling `cleanPaste(text)` helper method. Default: []
+* __cleanPastedHTML__: cleans pasted content from different sources, like google docs etc. Default: false
+* __cleanReplacements__: custom pairs (2 element arrays) of RegExp and replacement text to use during paste when __forcePlainText__ or __cleanPastedHTML__ are `true` OR when calling `cleanPaste(text)` helper method. Default: []
 * __cleanAttrs__: list of attributes to remove when ... default: ['class', 'style', 'dir']
 * __cleanTags__: list of element tag names to remove... default: ['meta']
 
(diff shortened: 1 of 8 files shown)


daviferreira:
Hey man, first of all thanks for the PR :)

I think would prefer using `cleanPastedHTML` everywhere, even on the option name. I know it would break old stuff, but we could add some sort of fallback. I just think we should aim for consistency. We should probably fix `getSelectionHtml` and `getSelectionHtml` too, but I can do that later on.

nmielnik:
Agreed, can we stick with `cleanPastedHTML`?  This looks like a mistake by me.  The general javascript standard is to capitalize acronyms (ie `innerHTML`, `encodeURIComponent`, etc) so I agree we should stick with that.

adrianorsouza:
+1 to stick with option name `cleanPastedHTML`, at first I thought its name has changed as of 4.x, I agree and prefer to keep it the same as it was in previous versions. I'll revert this PR changes and update with corrections to the option name `cleanPastedHtml` to `cleanPastedHTML` everywhere in the code also README. this makes less confusing for those who are migrating from 3.x to 4.x. I also think it wouldn't break older versions because this new option name `paste.cleanPastedHtml` is not working at all when you want to clean up paste rich text content, but let me know if I'm wrong.

phiggins42:
Really it's two issues? The name changed and the old option didn't translate to the new option. Kind of disappointed this made it through to release with another release (minor) while this known issue existed. Also releasing invalidates the work I had done to prevent this exact scenario.

nmielnik:
Yeah, this is a pretty crappy mistake by myself :disappointed: and the releases make it much worse.

If I understand the issue correctly though, it would appear the `cleanPastedHtml` option -- as a nested property inside the `paste` option -- never worked (is that correct @adrianorsouza ?)

If that's the case, then we don't need to be backwards compatible to support something that was broken to begin with, so this would be a patch as of now. Is that correct?

adrianorsouza:
@nmielnik it's correct, as of 4.x the new nested option `paste.cleanPastedHtml` never worked so I'd consider as a patch to fix this feature. I don't think a BC for nested option is needed as it was broken.

Look at config examples before this patch:
```js
{
	// doesn't clean the html content
	// doesn't warn deprecated option for cleanPastedHTML
	cleanPastedHTML: true,
	
	// OK warn deprecated option for forcePlainText
	forcePlainText: false
}

paste : {
	// works fine
	cleanPastedHTML: true,
	forcePlainText: false
}

paste : {
	// doesn't work at all
	cleanPastedHtml: true,
	forcePlainText: false
}
```

after fixes:

```js
{
    // OK clean the html content
    // OK warn deprecated option for cleanPastedHTML
    cleanPastedHTML: true,

    // OK warn deprecated option for forcePlainText
    forcePlainText: false
}

paste : {
    // works fine
    cleanPastedHTML: true,
    forcePlainText: false
}

paste : {
    // invalid option
    cleanPastedHtml: true,
    forcePlainText: false
}
```

nmielnik:
Ok, thanks a ton for clarifying, and thanks a ton for finding and fixing this :+1:

nchase:
> Ok, thanks a ton for clarifying, and thanks a ton for finding and fixing this :+1:

Seconded. Excellent find @adrianorsouza!

adrianorsouza:
OK cheers guys!

PR #556, opened 2015-04-13
Expose method of detaching custom event listeners + detach on destroy

nmielnik:
* Expose `unsubscribe()` as way to detach custom event listeners
* Detach all custom listeners when `destroy()` is called


nchase:
:shipit:

PR #557, opened 2015-04-14
Blur fixes

nmielnik:
* Handle case where clicking on an element that does not clear selection (ie, an element with `user-select: none`) does not cause the toolbar to blur
* Add tests for this fix, as well as a test to validate previous fix for dragging selection outside of editor (the fix prevented the toolbar from hiding, which it had been doing before).


nchase:
looks good to me. :shipit:

Issue #563, opened 2015-04-16
Pasted link disappears


nmielnik:
Exposing an issue found by @websirnik in #36 

> I'm testing 4.5.2. With the following options:
> ```js
paste: {
   cleanPastedHTML: true,
   forcePlainText: false
   }
```
> When I paste `<a href="https://google.com">Google</a>`, I expect it would render as a hyperlink [Google](https://google.com), but it renders as a plain-text `Google` (not link). 

> Can you check this behaviour. I'm testing on the` /demo/clean-paste.html` that's included in the repo.

nmielnik:
Also from @websirnik :

> I'm testing on Safari/ChromeCanary. I must say after further testing I got some content pasted correctly. But if you take for instance this [link](http://techcrunch.com/2015/04/15/sources-yahoo-in-talks-to-buy-foursquare/), and copy-paste the 1st paragraph from the article, links does not preserve.

nmielnik:
@websirnik can you pull down the branch I made `paste-fixes` and see if this fixes your issue?  I made some fixes that might help with this.

websirnik:
Testing `paste-fixes` branch. Safari - no results. Chrome - only first links renders, the rest not.
![screen shot 2015-04-16 at 14 44 44](https://cloud.githubusercontent.com/assets/282871/7182142/6310ba7e-e447-11e4-9526-c8c29c5a7e23.png)

nmielnik:
Interesting, I'll try to investigate further.

Once thing to look at in the meantime is to attach to the `paste` event, and inspect what is inside of `event.clipboardData.getData('text/html')` and `event.clipboardData.getData('text/plain')`.

If you can look at that, it might help see the inconsistencies here and lead to a fix.

websirnik:
Hope that helps. Tested on the same article.
Chrome:
![screen shot 2015-04-16 at 15 14 17](https://cloud.githubusercontent.com/assets/282871/7182752/4e6ec4ea-e44b-11e4-966d-df8977722853.png)
Safari:
![screen shot 2015-04-16 at 15 15 15](https://cloud.githubusercontent.com/assets/282871/7182769/759d9b90-e44b-11e4-9c04-38966c969852.png)

nmielnik:
@websirnik so I was able to address some of the issues that you called out, but it looks like things are going to depend on just how you copy and paste.

When I copy and paste the text from the page you sent, I was able to get all the text with the links preserved in chrome (using my branch).

[...]

PR #567, opened 2015-04-20
Paste fixes

nmielnik:
* Improve the regex used for extracting and preserving links from pastes
* In cases where there isn't any html in the clipboard during paste, fall back to plain text
* When removing `<span>` elements during paste, unwrap the element and leave the innerHTML as is, instead of converting the `<span>` contents to plain text


nmielnik:
This should address some of the issues called out in #563

nchase:
:shipit:

PR #568, opened 2015-04-20
fix anchor preview

jchamois:
Hello,

Here is a fix for the preview anchor. If a tag is present in the a tag, the preview of the anchor fail. You can reproduce by create an anchor in the editor and then put a bold on it via the toolbar. This way the 'b' tag is in the 'a' tag and the preview fails. 
I put a test on the target to check if we hover a a tag or a children.

Thanks


dist/js/medium-editor.js
```diff
@@ -2468,13 +2468,24 @@ var AnchorPreview;
...
-        handleEditableMouseover: function (event) {
-            if (event.target && event.target.tagName.toLowerCase() === 'a') {
+       handleEditableMouseover: function (event) {
+            var target;
+
+             if(event.target){
+
+                if(event.target.tagName.toLowerCase() === 'a'){
+                    var target = event.target;
+                }else if(event.target.parentNode.tagName.toLowerCase() === 'a'){
```
j0k3r: What if the text is bold **and** italic ?
Maybe you can take some inspiration from the [`setTargetBlank`](https://github.com/daviferreira/medium-editor/blob/master/src/js/util.js#L322-L333).

j0k3r:
You should do your modification on `extension-anchor-preview.js` instead of on the dist file which is generated by Grunt

nmielnik:
@jchamois , as @j0k3r pointed out, you should make your proposed changes in `extension-anchor-preview.js` and then run `grunt` to generate the dist files.  Without this your change will be wiped out the next time the dist files are generated.

jchamois:
thanks for your comments, here is another commit. I check the closest a tag in order to allow to put italic, bold ad everything else.

nmielnik:
@jchamois Hey, instead of making a change to the demo page, could you add a test to verify this change?  We probably don't want to change the demo as a proof that this works.

If you need any help, just let us know and we can try to help make a test.

jchamois:
@nmielnik Hello ! I'va added test for my fix, I hope I've done it right ^^ Please let me know and thanks

j0k3r:
Could you rebase your branch against the master? You'll have some conflicts to resolve.
Then, Travis will be able to launch the test suite.

src/js/extension-anchor-preview.js
```diff
@@ -133,13 +134,36 @@ var AnchorPreview;
             this.instance_handleAnchorMouseout = null;
         },
 
+        getClosestTag : function(el, tag) {
```
j0k3r: I guess that function will need to be in the Util class instead.
Also, don't forget to add a comment above the function to tell what it should do.

jchamois:
@j0k3r it's done ^^ Hope it's ok

nmielnik:
@jchamois last things, could you revert the changes in `/demo/index.html`?  Since we have tests, it's best not to put this in the demo page.

nmielnik:
Awesome, thanks @jchamois!

PR #569, opened 2015-04-20
Use real link in anchor preview

j0k3r:
This convert the `<i>` tag used to preview the link to an actual `<a>` so we can right click on the link and open in a new window (for example) to actually visit the link.
The (left) click still displays the form to edit the link.

Here is gif to illustrate (badly resized by Github – _click on it to view it in full size_) :

medium2

I've also updated the test to use `previewValueSelector` instead of hardcoded `i`.


spec/anchor-preview.spec.js
```diff
@@ -40,7 +40,7 @@ describe('Anchor Preview TestCase', function () {
             expect(anchorPreview.showPreview).toHaveBeenCalled();
 
             // link is set in preview
-            expect(anchorPreview.getPreviewElement().querySelector('i').innerHTML).toBe(document.getElementById('test-link').attributes.href.value);
+            expect(anchorPreview.getPreviewElement().querySelector(anchorPreview.previewValueSelector).innerHTML).toBe(document.getElementById('test-link').attributes.href.value);
```
nmielnik: I'm wondering if we should still be using `a` instead of referencing what's inside the anchorPreview itself.  I'm assuming the tests broke when you changed away from a `<i>` and perhaps that's what we would want should someone attempt to change the DOM again?

What do you think?

nmielnik:
For some reason in Chrome only, when you right click the anchor preview link, the preview disappears.  This doesn't seem to happen in any of our other supported browsers.  Do you see this behavior?

j0k3r:
Actually I've made the gif in Chrome for Windows. On which OS did you get this behavior?

nmielnik:
Mac OSX 10.10.2...a bit odd that it only reproduces on Chrome on Mac OSX.  It seems like the right click still works so perhaps we can look into this issue later.

Also, thanks a bunch for the PR, this improvement makes a lot of sense to me and I wonder why it wasn't there to begin with :smile:

spec/anchor-preview.spec.js
j0k3r: Well, I'm ok to use `a` but it feel more relevant to re-use the defined option (ie: `previewValueSelector`).

What if someone extends the AnchorPreview extension to change this selector without updating the spec.
Should the test pass or fail ?

j0k3r:
I was wondering why the `i` element was chosen in the first place :)

That's lead me to the css, I should remove the `font-style: normal;` since it's now a normal style, no longer an italic.

spec/anchor-preview.spec.js
nmielnik: I think I would expect the test to fail actually, because they may have broken your behavior which allowed for the link to be right-clicked to be navigated to!

If they're overriding the extension in their own usage of medium-editor, I don't think they'll be running our set of tests against it, but even if they were, they'll need to change the test accordingly to make sure it passes.  If the `<i>` had given us some sort of behavior that was now different, by not referencing the `previewValueSelector` directly in the case, it at least caused you to look at the tests to see what was failing, and I think we'd want someone to do that again if they changed away from an `<a>` tag.  If we use `previewValueSelector` the tests will just pass even though they may have removed your right click to navigate behavior.

My $0.02 anyway

j0k3r: I agree :)

PR #571, opened 2015-04-20
Speed up Travis build

j0k3r:
Regarding my previous PR, I've seen you didn't use the container based infra of Travis.

This infra speed up test by starting them faster than the non-container one.
If you check http://www.traviscistatus.com/ you'll see that the non-container can form a bottleneck (usually when both US and EU are awake) while the container infra is never full.

:arrow_right: http://docs.travis-ci.com/user/workers/container-based-infrastructure/

Also, when switching to container based infra, we can cache deps and they can be shared between builds.

:arrow_right: http://docs.travis-ci.com/user/caching/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,13 @@
+# faster builds on new travis setup not using sudo
+sudo: false
+
+# cache vendor dirs
+cache:
+    directories:
+        - node_modules
+
 language: node_js
+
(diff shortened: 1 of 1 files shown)


nmielnik:
:clap:

daviferreira:
Nice job :+1:

Issue #544, opened 2015-04-09
expose `paste` as first Extension


phiggins42:
all of the tooling should be in place already. expose PasteHanlder as the first 'extension', allowing override/replacement/magic-api. 

(paste best candidate because hasn't been released as even a minor bump, and already setup to handle newstyle options, and has never previously been exposed as replaceable, so low back compat risk)

deprecated options:
```
        cleanPastedHTML: [false, "paste.cleanPastedHTML"],
        forcePlainText: [true, "paste.forcePlainText"],
```

Unreleased functionality includes: custom paste regexps, relaying of base options to paste options (which will be simplified by this change?)

disable builtin paste handling:
```
new MediumEditor(".editor", { 
    paste: false
});
```

(paste handling on by default)

forcePlainText: false

```
new MediumEditor(".editor", {
    paste: { forcePlainText: false }
});
```

providing a replace pastehandler:

```
new MediumEditor(".editor", {
     paste: MediumEditor.extensions.paste.extend({
            forcePlainText: false,
            init: function(instance){ /* i can connect and do special things */ }
     })
});
```

note, as a back-compat thing to allow already-instantiated plugins in options passed, the following would "work":

```
new MediumEditor(".editor", {
      extensions:{
            "paste": new MySpecialPasteHandler()
      }
});
```

but the preferred syntax would be:

```
new MediumEditor(".editor", {
    paste: MySpecialPasteHandler
});
```

passing non-default(?) deprecated options would set the default `paste` options for you, and warn about deprecation. this would provide all functionality slated for 4.3 with a warning the options would be removed in 5.0

etc

phiggins42:
cc @nmielnik @daviferreira

nmielnik:
So I think the preferred syntax should remain:

```
new MediumEditor(".editor", { 
    extensions: {
        "paste": new MySpecialPasteHandler()
    }
});
```

For my org, where we have several complex and dependent extensions, being able to instantiate the extensions before passing them in allows us to create our own set of base classes that take in their own set of options to do set themselves up.  We can also keep a reference to these instantiated objects, in places where we might not necessarily want to keep a reference to the medium editor object itself.

Since we have the `init()` method as the method MediumEditor calls to initialize it, if we modify that method to actually take in the `options` for that specific built-in extension, then we can cover all the bases with our existing model, and not necessarily need to have the `{ paste: MySpecialPasteHander }` syntax.

[...]

PR #572, opened 2015-04-21
Convert PasteHandler into an extension

nmielnik:
Should be implementing majority of #544 

* Convert PasteHandler into a "new-style" extension and treat it as such internally
  * This means options are now members of the PasteHandler prototype instead of a separate `.options` property
* Define default properties and functions for extensions
  * Including adding lots of comments to describe the properties/functions
* Fix bug where extensions that aren't buttons were not getting put into `.commands` array
* Support overriding PasteHandler via extensions option:
```js
var pasteHandler = new MyCustomPasteHandler();
var editor = new MediumEditor('.editor', {
  extensions: {
    paste: pasteHandler
  }
});
```
* Support overriding PasteHandler via paste options:
```js
var editor = new MediumEditor('.editor', {
  paste: {
    forcePlainText: false,
    cleanPastedHTML: true,
    init: function () { // custom code }
  }
});
```


nmielnik:
@daviferreira @nchase @phiggins42 Let me know what you think...

src/js/core.js
```diff
@@ -356,21 +356,20 @@ function MediumEditor(elements, options) {
...
-                targetBlank: this.options.targetBlank,
-                contentWindow: this.options.contentWindow,
-                ownerDocument: this.options.ownerDocument
-        });
-
-        this.pasteHandler = new PasteHandler(this, pasteOptions);
+    function initPasteHandler(options) {
+        // Backwards compatability
+        var defaultsBC = {
+            forcePlainText: this.options.forcePlainText, // deprecated
```
daviferreira: Maybe we should add the deprecated warning somehow?

daviferreira:
:+1:

src/js/extension.js
```diff
@@ -66,8 +66,124 @@ var Extension;
...
-        // core API definition for an "extension" follows:
+        },
+
+        /* parent: [boolean]
+         *
+         * setting this to true will set the .base property
+         * of the extension to be a reference to the
+         * medium-editor instance that is using the extension
+         */
+        parent: false,
```
phiggins42: i would almost default this to true for all the things. especially if most of the internal things already make the assumption the .base is there

src/js/extension.js
```diff
@@ -66,8 +66,124 @@ var Extension;
...
+        checkState: null,
+
+        /* getButton: [function ()]
+         *
+         * If implemented, this function will be called when
+         * the toolbar is being created.  The DOM Element returned
+         * by this function will be appended to the toolbar along
+         * with any other buttons.
+         */
+        getButton: null,
```
phiggins42: would this not be better served in the Button (defaultButton) extension?

src/js/paste.js
```diff
@@ -47,37 +47,53 @@ var PasteHandler;
...
+        cleanAttrs: ['class', 'style', 'dir'],
+
+        /* cleanTags: [Array]
+         * list of element tag names to remove during paste when __cleanPastedHTML__ is `true` or when
+         * calling `cleanPaste(text)` or `pasteHTML(html, options)` helper methods.
+         */
+        cleanTags: ['meta'],
+
+        /* ----- internal options needed from base ----- */
+        contentWindow: window,
```
phiggins42: these aren't right. you are overriding the value of whatever they passed to base options. if someone passed an explicit not-`window` contentWindow value to base, the paste will be using this reference no?

src/js/paste.js
```diff
@@ -129,15 +145,15 @@ var PasteHandler;
...
                 }
-                Util.insertHTMLCommand(this.options.ownerDocument, html);
+                Util.insertHTMLCommand(this.ownerDocument, html);
             }
         },
 
         cleanPaste: function (text) {
             var i, elList, workEl,
-                el = Selection.getSelectionElement(this.options.contentWindow),
+                el = Selection.getSelectionElement(this.contentWindow),
```
phiggins42: unless the defaultsMixer thing is passing contentWindow and friends as ctor props, this is the wrong reference I think? not that I am _opposed_ to Extension's getting a default .document and .window property relayed from base options ... so instead of `Util.insertHTML(this.base.option.ownerDocuemnt, html)` you'd `Util.inserHTML(this.document, html)`

dist/js/medium-editor.js
```diff
@@ -4267,11 +4408,11 @@ function MediumEditor(elements, options) {
         },
 
         cleanPaste: function (text) {
-            this.pasteHandler.cleanPaste(text);
+            this.getExtensionByName('paste').cleanPaste(text);
         },
 
         pasteHTML: function (html, options) {
-            this.pasteHandler.pasteHTML(html, options);
+            this.getExtensionByName('paste').pasteHTML(html, options);
```
phiggins42: i like this. in my initial version I had all the extensions register themselves on the editor instance (eg: passing a `paste` extension would override the `this.paste` value ... then code could defensively be written as

```
if(this.paste){ this.paste.pasteHTML(html, options) }
```

your above pasteHTML call will throw if someone turned off paste with `paste: false` option, no?

src/js/extension.js
```diff
@@ -66,8 +66,124 @@ var Extension;
...
+         * If implemented, this function will be called once on each extension
+         * when the state of the editor/toolbar is being updated.
+         *
+         * If this function returns a non-null value, the exntesion will
+         * be ignored as the code climbs the dom tree.
+         *
+         * If this function returns true, and the setActive() function is defined
+         * setActive() will be called
+         */
+        queryCommandState: null,
```
phiggins42: I mean, perhaps it does make sense to expose "any extension can be a button or not", but logically my mind goes to "if i'm making my own "button" directly I'd `MediumEditor.extensions.Button.extend({...})` instead of Extension.extend({}) and provide my own versions of working Button.prototype things from a DefaultButton?

src/js/extension.js
```diff
@@ -66,8 +66,124 @@ var Extension;
...
+        setInactive: null,
+
+
+        /* onHide: [function ()]
+         *
+         * If implemented, this function is called each time the
+         * toolbar is hidden
+         */
+        onHide: null
     };
```
phiggins42: and since we're talking here about the "core extension" API i'd like to propose a few things (that fall in line with paste being the first official Extension) ... put an `on` `off`, `subscribe` and `unsubscribe` method on the Extension base that relays to the `this.base.on/etc` calls it assumes are there (because parent: true) ... and in `destroy` iterate any internally subscribe'd/on'd handlers and unregister them. then extensions could "just connect" as needed and know they will be cleaned up as part of the lifecycle of the editor.

src/js/paste.js
```diff
@@ -47,37 +47,53 @@ var PasteHandler;
...
+
+        /* cleanTags: [Array]
+         * list of element tag names to remove during paste when __cleanPastedHTML__ is `true` or when
+         * calling `cleanPaste(text)` or `pasteHTML(html, options)` helper methods.
+         */
+        cleanTags: ['meta'],
+
+        /* ----- internal options needed from base ----- */
+        contentWindow: window,
+        ownerDocument: document,
```
phiggins42: would suggest leaving these as this.base.options.ownerDocument internally ... but like the .document and .window idea. if they are in fact "core" options, that's where they should be read from

src/js/paste.js
nmielnik: The `initPasteHandler()` function in `core.js` is passing these values into the constructor, so I believe this should be working fine.

However, I do like the chance to rename these options (`this.document` instead of `this.ownerDocument`).  I will do that

src/js/paste.js
nmielnik: These options should be getting overriden when the PasteHandler is instantiated in `initPastehandler()` in `core.js`

src/js/paste.js
nmielnik: I've gone back and forth on this, and I seem to be leaning towards having the options passed from the instance into this, at least for our well defined extensions (anchor-preview, anchor, etc.)  

I'm not set on it, but I feel like it's better to have all the options in one spot as opposed to some option passed into the editor, and some the pastehandler gets from the instance itself...

src/js/extension.js
nmielnik: Yeah, I agree.  I haven't seen any extensions which haven't needed this property.  Maybe we should just do away with it and always set `.base` on all extensions?

src/js/core.js
nmielnik: Agreed, let me see if there's an easy path to do that, or let me know if you have ideas.

dist/js/medium-editor.js
nmielnik: Yeah, because I wanted to think through those parts more, I purposefully made it so you can't turn off the pastehandler.  I want to consider what you're suggesting vs keeping everything in the extensions array in another iteration after these changes get in.

For now, this is a safe workaround without making a decision one way or another.

src/js/core.js
phiggins42: I had a fancy mixDefaults thing I was working on, that simply detected the use of a non-default deprecated arg and passed along the right value to the right place ... it's why I added getObject and setObject to Util ... i've been trying to get this same patch going for some time now, but ended up breaking a buch of tests because of half-implementation ... (my mixDefaults was doing the initCommands work too, but that change requires Button to be an extension, too)

src/js/extension.js
nmielnik: If we go down the path of always have `.base` set to an instance of the editor, I definitely agree having helpers like the event methods on this object are good.

If `.parent` is still respected, and thus you can have `.base` not set to an instance of MediumEditor, then I'm not as big of a fan of this.

src/js/paste.js
phiggins42: okay i missed it in the review then is all. carry on :)

src/js/core.js
nmielnik: Yeah, I'm definitely ok with something ugly, these are deprecated properties/method after all so I wouldn't want to spend much time making it convenient.  That's my opinion though.

src/js/extension.js
nmielnik: I've been trying to think about this a bit, specifically because of extensions that support 'forms' as well. A Button extension would extend Extension.  But should a FormExtension extend Button or extend Extension?  I'm thinking it might be better for FormExtension to extend from Extension since it's probably unnecessary for it to share all the same code as the Button.  In that case, it would be good to keep `getButton` defined here.

I think the value in Button is that it's a really simple way to create a button that just maps to a `document.execCommand` command.  That's what I originally called it `DefaultButton` instead of button because it doesn't really help you out a lot if you're trying to create a button that does something other than just call `execCommand`.

I also look at an example like [this](https://github.com/daviferreira/medium-editor/wiki/Custom-Buttons-and-Extensions#highligher-extension-using-checkstate) and I think it might be ok to leave `getButton` in the base extension, because of the simplicity for some cases...

Because of this, I'm wondering if we should even have a Button extension right now. Just some thoughts, I might just be rambling and not giving a good argument. 

Either way, if we create a Button extension, that'll happen in another pull request, and if we want to remove `getButton` from Extension then, we still can.

src/js/core.js
nmielnik: And I forgot, I actually already implemented this and we are already giving warnings when people use these props :stuck_out_tongue_winking_eye: 

the `mergeOptions()` method in `core.js` takes care of this.

src/js/extension.js
nmielnik: I'm gonna track this in #576

src/js/extension.js
nmielnik: We can talk more about this in an issue or a separate PR.

PR #577, opened 2015-04-22
jscs

daviferreira:


nchase:
:smirk: I like where this is heading

nchase:
I know the project is fairly deeply-invested in Grunt as a tool for handling the processes that the library needs to undergo, but have you considered just using the JSCS binary? Its [options](http://jscs.info/overview.html) will let you configure all of the things you're interested in.

Travis will know how to deal with the output – both use standard unix exit code to signal errors.

nmielnik:
I think I need spellcheck for commit messages :disappointed:

daviferreira:
@nchase sure, but why not use grunt for consistency? Unless we move everything away from it.. I'm game :dancers:

PR #588, opened 2015-04-30
Fix unwrap util

yacomink:
The unwrap util was iterating over a nodeList, but the loop body was altering the length of the nodeList.

This caused cleanPaste'd spans with multiple child nodes to be partially truncated.


daviferreira:
:+1: :ok_hand:

nmielnik:
@yacomink nice catch, thanks!

PR #593, opened 2015-05-06
Create isMetaCtrlKey

### LINKED ISSUE #591: Ctrl-K shortcut for link breaks default Mac OS X text editor behavior
Mac OS browsers, like other Mac applications, [support emacs-inspired keybindings for editing text](http://www.quora.com/Why-are-emacs-style-key-bindings-built-into-Cocoa-text-fields-on-Mac-OS-X). This includes Ctrl-A to go to the beginning of the line and Ctrl-K to go to the end of the line.

As a result of PR #589 , Ctrl-K no longer works in Medium on Macs, instead it creates a link if the text is highlighted or does nothing if it isn't. Ctrl-A still works.

(The original medium.com editor supports ctrl-A, but ctrl-k deletes the entire line when executed on a mac)

A good fix would be to make the link shortcut CMD-K on Mac and Ctrl-K on all other platforms (since windows and linux use Ctrl as their shortcut key).

j0k3r:
To handle metaKey & ctrlKey on Mac & non-Mac.
Also fix #591


j0k3r:
_Changes Unknown_, thanks @coveralls for the feedback :frog:

nmielnik:
awesome :+1: 
coveralls :-1:

PR #594, opened 2015-05-06
Fix "this.getAttribute is not a function" error

jmlacroix:
This update fixes the `Uncaught TypeError: this.getAttribute is not a function` error message.


daviferreira:
Thanks btw, nice catch :+1:

jmlacroix:
My pleasure, thanks for such a great editor!

PR #597, opened 2015-05-08
Create custom 'oninput' event which works in all supported browsers!

nmielnik:
This work exposes a new `editableInput` event which will fire whenever the user makes changes to MediumEditor maintained element.

Currently, the actual `input` event is supported for contenteditable elements in all browsers except for IE.  None of IE9, IE10, or IE11 fire the `input` event on contenteditable elements (though they do fire for things like textboxes and textareas).

This change will feature detect whether the current browsers properly supports the `input` event.  If it exists, we will attach to `input` and the `editableInput` custom event will trigger whenever that event occurs and the HTML has changed.

When `input` is not properly supported, we will:
1. Attach to the `keypress` event for each contenteditable.  If the innerHTML has changed, we'll fire `editableInput`
2. Attach to `selectionchange` on the document.  If this fires, and one of our contenteditable elements is the `activeElement`, we'll check to see if the innerHTML has changed, and if so, fire `editableInput`
3. Listen to all calls to `document.execCommand`.  After each call, if the innerHTML has changed for one of our contenteditable elements, first `editableInput`.

This `contentEditable` custom event will only fire if something in the html has changed.  This means we'll be tracking the contents of each contenteditable element monitored by MediumEditor and doing a diff against it as the events fire.


dist/js/medium-editor.js
```diff
@@ -1832,6 +2013,53 @@ var Events;
...
+
+    // Create a temporary contenteditable element with an 'oninput' event listener
+    tempElement = document.createElement('div');
+    tempElement.setAttribute('contenteditable', true);
+    tempElement.innerHTML = 't';
+    tempElement.addEventListener('input', tempFunction);
+    tempElement.style.position = 'absolute';
+    tempElement.style.left = '-100px';
+    tempElement.style.top = '-100px';
+    document.body.appendChild(tempElement);
```
j0k3r: Since this modification, I have to wrap my `new MediumEditor('.editable')` inside a `onload` otherwise the `document.body` is null :confused: 

As far as I understand, this `oninput` was made only for IE right? Shouldn't we test if we are on IE and not attach this element to the body?

In a quick and drity mode I commented this line and 2053 and everything works again.

PR #598, opened 2015-05-10
Cleanup Test Cases

nmielnik:
In addition to doing some big test cleanup, this PR also ensures that when an element is generated for a `<textarea>` that element is deleted when `destroy()` is called.

Test Cleanup Tasks:
* Fix tests for IE9 which were hanging indefinitely in SauceLabs and causing failures
* Create some test initialization/tearDown helpers for managing created editors and elements
* Ensure all editor instances and created elements are cleaned up at the end of each tests
* Fix tests that were dependent on remnants from other tests :disappointed:
* Fix some issues with toolbar + textarea tests


j0k3r:
Nice work :+1:

Issue #229, opened 2014-05-12
Adding support for autolink


ikhattab:
Hi,

I'm working on adding support for autolink when user types/paste some url that looks like a valid url and press space after/before it, it should be converted to link, like this
![autolink](http://f.cl.ly/items/301K1n0q1q1K2y0Z2L2n/autolink.gif)

You can see what I did so far in https://github.com/ikhattab/medium-editor/tree/FEAT/autolink

I will submit pull request when I finished probably this weekend.

TODO:
- [ ] Adding support for converting link after/before pressing ENTER.
- [ ] Adding testes.

nmielnik:
looks like we have a proposed implementation for autolink in #599

nmielnik:
Now that #599 exists, auto-link is supported.  This should be released in v4.9.0 when that comes

PR #599, opened 2015-05-11
Auto link support

### LINKED ISSUE #304: Auto URL detection when creating links
It would be great when one selects some text and wants to make it a link that editor would automatically detect whether selection is a URL. It it is, it should be automatically converted to a link
* either without opening popup at all
* opening link creation popup but prepopulating it with the same URL

> This feature should likely get on the list for Medium Editor 2.0

patrick-webs:
This code may fix #304 or perhaps @NoelBaron's solution mentioned in that will turn out to be better overall.

The goal here is to provide auto-link support across all browsers - MSIE had a built-in implementation but none of the other browsers do, and MSIE's has issues when dealing with links that have formatting split across part of the URL text.

Link detection is performed using a regular expression; the code then iterates over all links in the text and if they are not already inside an anchor tag, creates an anchor tag around them. Any inline elements such as B, Strong, Em, I, and Span tags are duplicated as needed to preserve the style both inside and outside the element (e.g., a link that is half-italic will remain half-italic after being linked).


daviferreira:
Nice stuff, this is a nice addition and hard to achieve :)

nmielnik:
This also addresses #229

src/js/extensions/auto-linker.js
```diff
@@ -0,0 +1,418 @@
...
+        div.appendChild(document.createTextNode('ab'));
+        div.appendChild(document.createTextNode('c'));
+        it = document.createNodeIterator(div, NodeFilter.SHOW_TEXT, null, false);
+        node = it.nextNode();
+        node.splitText(1);
+        node = it.nextNode();
+        if (node.nodeValue === 'c') {
+            nodeIteratorDetectsNewNodes = false;
+        } else if (node.nodeValue !== 'b') {
+            throw new Error('Unexpected behavior of NodeIterator. Code will not work properly');
```
nmielnik: Perhaps instead of throwing an error here we could just write to the console and turn off auto-linking?

dist/js/medium-editor.js
```diff
@@ -2997,6 +2997,422 @@ var AnchorPreview;
...
+        }
+    }());
+    // end feature test
+
+    function assignHttpToProtocolLessUrl(url) {
+        if (url.indexOf('://') === -1) {
+            return 'http://' + url;
+        } else {
+            return url;
+        }
```
j0k3r: You can avoid this `else` and just write `return url;` since the `return` inside the `if` will exit the function

dist/js/medium-editor.js
```diff
@@ -2997,6 +2997,422 @@ var AnchorPreview;
...
+
+        nodes.sort(function (a, b) {
+            var aIndex = lookup.indexOf(a),
+                bIndex = lookup.indexOf(b);
+
+            return aIndex - bIndex;
+        });
+    }
+
+    function complexify(root, descendants) {
```
j0k3r: Could you add a description on what it is supposed to do?

src/js/extensions/auto-linker.js
```diff
@@ -0,0 +1,418 @@
+/*global Extension, Util */
+
+var AutoLinker,
+    AutoLinkerStatics = {},
+    KEY_CODES = {
```
daviferreira: Use `Util.keyCode` instead? https://github.com/daviferreira/medium-editor/blob/master/src/js/util.js#L62

src/js/extensions/auto-linker.js
```diff
@@ -0,0 +1,418 @@
...
+            }
+        }
+
+        while (vv !== ww) {
+            vv = vv.parentNode;
+            ww = ww.parentNode;
+        }
+
+        return vv;
+    }
```
daviferreira: I would go for more declarative variable names instead of vv, www etc.

src/js/extensions/auto-linker.js
```diff
@@ -0,0 +1,418 @@
...
+                matches.push({
+                    href: match[0],
+                    start: match.index,
+                    end: match.index + match[0].length
+                });
+            }
+            return matches;
+        },
+
+        createLink: function (textNodes, href) {
```
daviferreira: Really big method, maybe break it? :)

src/js/extensions/auto-linker.js
```diff
@@ -0,0 +1,418 @@
...
+            // (Medium deletes the spaces/returns between P tags so the textContent ends up without paragraph spacing)
+            var paragraphs = contenteditable.querySelectorAll('p'),
+                linkCreated = false;
+            for (var i = 0; i < paragraphs.length; i++) {
+                linkCreated = this.performLinkingWithinElement(paragraphs[i]) || linkCreated;
+            }
+            return linkCreated;
+        },
+
+        performLinkingWithinElement: function (element) {
```
daviferreira: break? :hammer:

README.md
```diff
@@ -74,6 +74,7 @@ MediumEditor also supports textarea. If you provide a textarea element, the scri
 * __allowMultiParagraphSelection__: enables the toolbar when selecting multiple paragraphs/block elements. Default: true
 * __delay__: time in milliseconds to show the toolbar or anchor tag preview. Default: 0
 * __disableAnchorPreview__: enables/disables the anchor preview element, which appears when hovering links and allows the user to edit the link when clicking. If toolbar is diabled (via __disableToolbar__ or `data-disable-toolbar attribute`) the anchor preview is always disabled so this option will be ignored. Default: false
+* __enableAutoLinker__: enables/disables the auto-linker feature, which automatically turns URLs entered into the text field into HTML anchor tags (similar to the functionality of Markdown). Default: false
```
nmielnik: I would call this `autoLink` so that in the future, if we add options for the auto link extension, we can just nest them inside the options:

ie:
```js
{
  autoLink: {
    option1: true,
    option2: 'something'
  }
}
```

or you can enable it like this:
```js
{
  autoLink: true
}
```

and the default options would still be valid:
```js
{
  autoLink: false
}
```

src/js/extensions/auto-linker.js
patrick-webs: done

src/js/extensions/auto-linker.js
patrick-webs: yes, that makes sense. updated.

src/js/extensions/auto-linker.js
patrick-webs: I renamed the variables.

src/js/extensions/auto-linker.js
patrick-webs: perfect, thanks

dist/js/medium-editor.js
patrick-webs: I added comments.

dist/js/medium-editor.js
patrick-webs: done

src/js/extensions/auto-linker.js
patrick-webs: I set up a catch statement in onKeypress now that will disable the event handling if any error/exception is thrown from internal code.

README.md
patrick-webs: OK, done.

nmielnik:
Thanks @patrick-webs this is awesome

Issue #601, opened 2015-05-13
dist file can no longer be included in <head> section of page


yacomink:
The `document.body.appendChild` introduced in `events.js` in the commit https://github.com/daviferreira/medium-editor/commit/5c24725e4bd7ca977c133884e6045982f410d66e means that this js can no longer be included in the `<head>` of a document. `document.body` will be null when this code is executed in that case.

j0k3r:
Related to what I wrote few hours in the PR: https://github.com/daviferreira/medium-editor/pull/597/files#r30237312

nmielnik:
This should be fixed in the 4.8.1 release

PR #602, opened 2015-05-13
Fix pasting multi-paragraph plain text chunks when cleanPaste is enabled

yacomink:
Also includes a stop-gap for #601


src/js/paste.js
```diff
@@ -124,9 +125,10 @@ var PasteHandler;
 
                 if (!pastedHTML) {
                     pastedHTML = pastedPlain;
+                    onlyText = true;
```
j0k3r: What if you didn't introduce this new variable `onlyText` and just removed this `if`?
On the next if below, `pastedHTML` will be empty and the condition will be false.

nmielnik:
Will #603 fix your issue, and thus you don't need the stop-gap?

nmielnik:
So I've checked in a fix for the issue you opened, and cut a new release.  Sorry to change things underneath you, but you'll need to update your fork with the new changes

yacomink:
good lookin out Nick, your update fixed my issue. I'm going to close this one and submit a rebased one

nmielnik:
ok, apologies again for the extra overhead, but I really wanted to get the issue you and @j0k3r found fixed and released as quickly as possible

PR #604, opened 2015-05-13
Fix pasting multi-paragraph plain text chunks when cleanPaste is enabled

yacomink:


spec/paste.spec.js
```diff
@@ -139,6 +139,43 @@ describe('Pasting content', function () {
             }
         });
 
+        it('should add paragraphs to pasted plain-text', function () {
+            var editorEl = this.el,
+                editor = this.newMediumEditor('.editor', {
+                    delay: 200,
+                    paste: {
+                        forcePlainText: false,
+                        cleanPastedHTML: true
```
nmielnik: So if you used the default options instead, you'd get the expected result without the code change:

default options:
```js
paste: {
    forcePlainText: true,
    cleanPastedHTML: false
}
```

Could you explain what the use case is that is causing this issue you're trying to fix?

yacomink:
Sure, my use case is I have one medium editor instance on my site that represents the main body of content in, let's say, a blog post. Different users are bringing in content from different sources. They may be pasting content from any of the following depending on their preferred writing environment:

- An article they wrote in Word, with links and bold/italics that need to be preserved
- An article they wrote in Google docs, with links and bold/italics that need to be preserved
- An paragraph from another website they want to quote, with links and bold/italics that need to be preserved
- A few paragraphs of text from a text editor they like to write in, or a plain text email they received

Items 1-3 above need cleanPaste to preserve markup they want. Item 4 requires that double newlines in a plain text paste be converted to paragraphs. So, my config is:

paste: {
   forcePlainText: false,
   cleanPastedHTML: true
}

yacomink:
Maybe an easier way to explain this: I don't think the value of cleanPastedHTML should affect what happens to pasted plain-text. This is what's happening currently that's causing me problems:

medium editor 7c demo

nmielnik:
I see.  The reason this extra line of code was added was because there are some cases where you can copy and paste some rich-text, and the pasted data actually resides in the `text/plain` data as escaped HTML.

I see what you're talking about, so I wonder if a more appropriate fix would be to always convert '\r\n' into `<p>` tags, regardless of whether it's treated as HTML or plain text.

So this line of code would remain:
```javascript
if (!pastedHTML) {
  pastedHTML = pastedPlain;
}
```

but we could instead move the call to `this.cleanPaste(pastedHTML)` to be after the code that replaces newlines with `<p>` tags.  So the handlePaste() function would look something like:

```javascript
handlePaste: function (event, element) {
  ...
  if (!pastedHTML) {
    pastedHTML = pastedPlain;
  }

  // Move this logic up
  if (!(this.disableReturn || element.getAttribute('data-disable-return'))) {
    // logic that splits by `\r\n`, wraps in `<p>` tags
    // this logic should NOT call Util.htmlEntities yet though
    ...
  }

  if (this.cleanPastedHTML && pastedHTML) {
    this.cleanPaste(html);
  } else {
    Util.insertHTMLCommand(this.document, Util.htmlEntities(html));
  }
}
```

what do you think?

yacomink:
Ahh, interesting. Might run into a slight problem because Word will
sometimes slip errant newlines within paragraphs that should _not_ be
converted. What if that conditional was something like `if (!pastedHTML &&
pastedPlainContainsEncodedHTML)`?

nmielnik:
So you don't think we'd be able to detect these errant newlines within paragraphs?

If that's the case, then I think the `pastedPlainContainsEncodedHTML` idea will work too.  Just need to inspect the plain text somehow?  Maybe a helper method like `Util.containsHtmlEntities()`?

nmielnik:
I'm on board with the idea of checking if the plaint-text contains encoded HTML and then not calling `cleanPaste()` if it doesn't.  If you have time to make that change I'll be happy to merge it in and include it in our next release.

Also, saw you quoted on the [slack website](https://slack.com/is/team-communication).  Pretty cool to be discussing PRs with a celebrity :sunglasses:

daviferreira:
@yacomink is a MediumEditor veteran :)

yacomink:
Haha, that quote keeps following me.

I should be able to get to this early this week. Do you recall in the rich
text case here is the html html-encoded or url encoded?

nmielnik:
So I've tried, but I've been unable to reproduce what I thought I remembered happening :grimacing: 

So, after wasting lots of time, I think it's fine to take this PR as-is.  My apologies for the runaround, do you mind updating your fork with the latest master so we can merge this in?

yacomink:
No worries, that makes my fix a bit easier! I've rebased and pushed, should
be good to merge.

nmielnik:
hmm, it seems like this rebase might have gone a little funk, any ideas why your commits are in here twice?

yacomink:
hrm, I'm not sure ... the overall diff looks clean though:
https://github.com/daviferreira/medium-editor/compare/yacomink:fix-multi-paragraph-text-paste-with-clean-paste-v2

I can cut a fresh v3 and a new PR if that would be easier

j0k3r:
Well, I can fix that using a clean rebase. Can I rebase your PR?

j0k3r:
I've rebased and squashed your commits, the cleaned PR is here #613 :smile:

PR #609, opened 2015-05-16
Various Extension Work

nmielnik:
* Convert image-dragging handling into an extension (so it can be overriden)
* Allow auto-link extension to be overriden
* Move paste.js into same folder as other extensions

Warning: the diff github did for these changes is terrible. Don't look at the dist files, just look at the src.


src/js/core.js
```diff
@@ -279,9 +235,21 @@ function MediumEditor(elements, options) {
     }
 
     function shouldAddDefaultAutoLinker() {
```
j0k3r: This is unrelated to the goal of this PR, but this shouldn't be `shouldAddDefaultAutoLink` instead? Since you rename AutoLinker to AutoLink?

nmielnik: Yeah, good point

PR #615, opened 2015-05-19
New-style base extensions for Buttons and Forms

nmielnik:
This is the 1st part of the final push towards our new style of extensions.
These changes create new extension base classes, and officially deprecate `DefaultButton`, `AnchorExtension`, and `FontSizeExtension` so they can be removed in v5.0.0.

* New `Button` base class for extensions
  * Provides base functionality for buttons (ie `getButton()` method)
  * Implements new-style extensions, where options are merged into prototype
  * Exposes `.extend()` for easily creating new buttons that derive from this button
  * Exposed as `MediumEditor.extensions.button`
* New `FormExtension` class, which extends from `Button`, for extensions with toolbar forms
  * Defines the required methods for extensions which have an in-toolbar form
  * Exposes `.extend()` for easily creating new form extensions
  * Exposed as `MediumEditor.extensions.form`
* Convert `AnchorExtension` to new-style `AnchorForm` extension, which derives from `FormExtension`
  * Exposed via `MediumEditor.extensions.anchor`
* Convert `FontSizeExtension` to new-style `FontSizeForm` extension, which derives from `FormExtension`
  * Exposed via `MediumEditor.extensions.fontSize`

Deprecating the following:
* `MediumEditor.statics.AnchorExtension`
* `MediumEditor.statics.FontSizeExtension`
* `MediumEditor.statics.DefaultButton`


phiggins42:
\o/ sweet ... i'll wait on official 4.9.1 for this, but I was able to seamlessly upgrade to 4.9.0 utilizing only minimal hackage of copying .extend onto anchorpreview and anchor.  this marks the last of what is necessary to "do everything evil i was doing" in a sanctioned way.

nkalush:
I'm having an issue with an extension I had that's probably related to this. The old extension followed the method in the wiki, but gets an error that getButton method can't be found, it does exist in the extension object though.

Looking at the base extensions the format is a lot different so I was trying to follow that but I'm getting `Button is not defined` errors even though it is totally in the code. 

Here's a trimmed down (though regrettably long still) example of what I'm doing:

```javascript
// importing dist/js/medium-editor.js here via grunt concat

var MediumLink;
(function (Button) {
    'use strict';

    /*global Util, Selection, Button */

    MediumLink = Button.extend({
        name: 'anchor',
        action: 'createLink',
        aria: 'link',
        tagNames: ['a'],
        contentDefault: '<b>#</b>',
        contentFA: '<i class="fa fa-link"></i>',
        key: 'k',

        init: function () {
            console.log('linkInit');
        },

        handleClick: function (event) {
            event.preventDefault();
            event.stopPropagation();

            console.log('btn click');
        }
    });
}(Button));

/*function MediumLink() {
    'use strict';

    // this.parent = true;
    this.button = document.createElement('button');
    this.button.className = 'medium-editor-action';
    this.button.innerHTML = '<i class="icon-link-outline"></i>';
    this.button.onclick = this.onClick.bind(this);
    this.button.dataset.target = '#modalID';

    //from https://github.com/jillix/medium-editor-custom-html
    this.insertHtmlAtCaret = function (html) {
        //code unrelated to question
    };
}

// MediumLink.prototype.setEvent = ;

MediumLink.prototype.init = function (e) {
    'use strict';
    var mediumlink = this;

    //Init stuff
};

MediumLink.prototype.handleEvent = function (e) {
    'use strict';

    if (e.type === 'click' && e.target === this.modalSubmit) {
        //Get data and open modal
    }
};

MediumLink.prototype.onClick = function () {
    'use strict';

    var mediumlink = this,
        sel = window.getSelection(),
        selectedText = window.getSelection().toString();

    if (sel.rangeCount >= 1) {
        //Open Modal Here
    }
};

MediumLink.prototype.getButton = function () {
    'use strict';

    return this.button;
};

MediumLink.prototype.checkState = function (node) {
    'use strict';

    if (node.tagName === 'A') {
        this.button.classList.add('medium-editor-button-active');
    }
};*/

document.addEventListener("DOMContentLoaded", function () {
    'use strict';
    link_extension = new MediumLink();
    editor = new MediumEditor('.editable_region', {
        buttons: ['customlink'],
        extensions: {
            'customlink': link_extension
        }
    });
    editor.setup();
});
```

nmielnik:
So the problem is that the global `Button` variable you're passing into your function does not exist:

```js
(function (Button) {
    ...
}(Button));

```

As long as you're using version 4.10.0 or newer, you should reference this variable via `MediumEditor.extensions.button` and not `Button`.

So this should work:

```js
(function(Button) {
  ...
}(MediumEditor.extensions.button);
```

Let me know if you're still having problems.

I'm in the process of updating our documentation, I apologize that the documentation is quite there yet :disappointed:

nkalush:
Thanks for helping, your suggestion fixed the not finding the class problem, but the getButton issue persists.

I made your change and removed the init function since I didn't want to overwrite the existing init. I get this error: `Uncaught TypeError: Cannot read property 'getButton' of undefined` on line 57 of `src/js/toolbar.js`. I added the getButton method to my extension's class but that didn't work, so there must be something in my initial setup that I'm doing wrong but I'm not sure what it is since I can't find a suitable example.

nmielnik:
Can you supply your updated code?  Perhaps a jsfiddle?

nkalush:
Sorry, that would have been helpful wouldn't it.

https://jsfiddle.net/rj6ufrtv/

nmielnik:
Ah, so the problem is `name` property of your custom extension.  You are currently setting the name to be `'anchor'`, but you specify the name as `customLink` for both the `buttons` array, and the `extensions`.

nmielnik:
You also don't need to call `editor.setup()` once you create the instance of MediumEditor.  Creating the object will take care of everything you need.

I've created an updated version of your fiddle that appears to be working.  Apologies again that the documentation is not there, I might try to use your jsfiddle as an example when I get the documentation up :smile: 

https://jsfiddle.net/rj6ufrtv/1/

nkalush:
Thanks, that worked.

PR #616, opened 2015-05-19
Fix cursor position for list

### LINKED ISSUE #561: Cursor position on create list
1 Type some text
2 Move to new line after text
3 Press list button

Empty list is created but cursor jumps to the end of previously typed text, not to first list item.

j0k3r:
The behavior was really strange...

I've tested it on Chrome on Windows, can someone try it on Chrome & Safari on Mac for example.

To reproduce:

1. start from an empty content
2. write some text
3. add a new line
4. select the written text
5. create a list
6. the cursor must be positionned right after your written text

Fix #561


j0k3r:
I've no idea why the test with *Platform: WIN7,internet explorer,9* fails :confused: 
I've tested on my working machine WIN7 with IE9, everything seems normal.

How can we see which test is failing on Saucelabs?

daviferreira:
https://saucelabs.com/jobs/764731d9bc5d407abf440387163dbb98

image

j0k3r:
@daviferreira I didn't check screenshots :) Thanks

spec/content.spec.js
```diff
@@ -18,12 +18,24 @@ describe('Content TestCase', function () {
...
+                buttons: ['orderedlist']
+            }),
+            target = editor.elements[0].querySelector('p'),
+            range, sel;
         selectElementContentsAndFire(target);
         fireEvent(editor.toolbar.getToolbarElement().querySelector('[data-action="insertorderedlist"]'), 'click');
         expect(this.el.innerHTML).toMatch(/^<ol><li>lorem ipsum(<br>)?<\/li><\/ol><ul><li>dolor<\/li><\/ul>?/);
+
+        // for Chrome & Safari we manually moved the caret so let's check it
+        if (!isFirefox() && !isIE()) {
```
j0k3r: Since `cleanListDOM` apply changes only for Chrome & Safari (as far as I understand from the commits history), the bug is only related to these browsers. That's why I restrict tests.

dist/js/medium-editor.js
```diff
@@ -4328,6 +4341,7 @@ function MediumEditor(elements, options) {
                 // parent also does not have a sibling
                 !node.parentElement.previousElementSibling &&
                 // is not the only li in a list
+                node.nextElementSibling &&
```
j0k3r: This remove a nasty `call tagName on null variable`, or sth like that

daviferreira:
:+1:

nmielnik:
Not related to your change, but I just realized we have 2 unwrap methods in Util.  I think the newer one `unwrap()` is similar to the jQuery method of unwrapping, so it might be better to delete the `unwrapElement()` helper and change this code to use `unwrap()`.  What do you think?

src/js/util.js
```diff
@@ -402,12 +402,25 @@ var Util;
...
+                this.unwrapElement(list.parentElement);
+
+                // move cursor at the end of the text inside the list
+                // for some unknown reason, the cursor is moved to end of the "visual" line
+                range = ownerDocument.createRange();
+                sel = ownerDocument.getSelection();
+                range.setStart(element.firstChild, element.firstChild.textContent.length);
+                range.collapse(true);
+                sel.removeAllRanges();
+                sel.addRange(range);
```
nmielnik: This chunk of code is almost completely duplicated at the end of the `handleBlockDeleteKeydowns()` in core.js.  Seems like a great candidate for a helper method in `selection.js`.

j0k3r:
@nmielnik Seems perfect to me to remove `unwrapElement` :+1:

src/js/util.js
j0k3r: Good point, are you ok with `Selection.moveCursor` ?

nmielnik: :+1:

nmielnik:
Yeah, you're right, I had forgotten that the Util methods are now exposed.  Good call.

PR #619, opened 2015-05-20
Convert AnchorPreview to new style extension

nmielnik:
This PR covers converting anchor-preview into the new style extension.

* Use `Extension.extend()` to create extension
* Embed options as properties of the prototype instead of a passed in object
* Options are passed as a `anchorPreview` sub-object of the large options object
* Expose reference via `MediumEditor.extensions.anchorPreview`

Other changes:
* Deprecate old anchor-preview extension
  * Renamed internally as `DeprecatedAnchorPreview`
  * Moved into `extensions/deprecated` folder
  * Remains exposed via `MediumEditor.statics.AnchorPreview`
* Update documentation


nmielnik:
So coveralls seems to be acting up and that's causing the build to fail.  The build is actually succeeding, and all tests are passing in SauceLabs, so don't worry about that little ol' red X :sunglasses:

j0k3r:
I've restarted the travis build and it's now fine :tada:

PR #621, opened 2015-05-20
allow nested contenteditables

phiggins42:
all the tests remain passing, can add another for the case. added a random demo/ file to show the usecase.


nmielnik:
Do you mind rebasing?  Our dist file situation leaves something to be desired :disappointed:

phiggins42:
i did a `make js` prior to commit I thought? based on current upstream/master ... what did I miss? (and yes, I personally am opposed to maintaining dist files in repos for this very reason)

nmielnik:
Another PR and a commit got merged into master at pretty much the exact same time you opened your PR, I'm guessing that's where the disconnect was.

Agreed, I would definitely like to explore removing the dist files as soon as we can.

phiggins42:
i think i got it this time. github pattern noob, sorry for the ugh commits. was confused.  +1 to figuring out how to not need to keep the dist/ files in branch (in tags makes sense, doesn't npm et al use whatever is current as the dist file? unsure). thanks for quick merge.

PR #622, opened 2015-05-20
Convert Placeholders to extension

nmielnik:
This PR converts the previously internal `Placeholders` object into an actual extension that is overrideable and is initialized + passed options as any other extension.


README.md
```diff
@@ -210,8 +257,10 @@ var editor = new MediumEditor('.editable', {
     },
     anchorPreview: {
         hideDelay: 300
+    },
+    placeholder: {
+        'Click to edit'
```
j0k3r: It should be `text: 'Click to edit'`

nmielnik: Good catch

PR #624, opened 2015-05-21
Fix line breaks on Safari

ilessiivi:
On recent Safari browsers, it's impossible to insert a line break as Shift + Enter just makes a new paragraph.

Ctrl + Enter works better, creating the `<br>`, but it also creates a new paragraph and results in a paragraph inside another.

In addition to checking !event.shiftKey in handleKeyup(), the code should check for !event.ctrlKey so that execCommand('formatBlock', false, 'p') won't be called.

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -1,7 +1,7 @@
 /*global describe, it, expect, spyOn,
     fireEvent, prepareEvent, firePreparedEvent, afterEach, beforeEach,
     selectElementContents, setupTestHelpers, placeCursorInsideElement,
-    isFirefox, isIE, Util, selectElementContentsAndFire */
+    isFirefox, isIE, isSafari, Util, selectElementContentsAndFire */
 
 describe('Content TestCase', function () {
     'use strict';
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -195,7 +195,7 @@ describe('Content TestCase', function () {
 
             var editor = this.newMediumEditor('.editor'),
                 p = editor.elements[0].querySelector('p'),
-                isSafari = isSafari();
+                s = isSafari();
 
             spyOn(document, 'execCommand').and.callThrough();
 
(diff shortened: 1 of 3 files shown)


daviferreira:
@ilessiivi want to write a test for it? :)

ilessiivi:
I've never done one before and it's failing. Help... :)

daviferreira:
:boom:

daviferreira:
Nothing to with the test actually, just a lint error.

screen shot 2015-05-21 at 13 45 14

You can run grunt on your local environment to check if everything is ok before commiting. Let me know if you need any help.

spec/content.spec.js
```diff
@@ -189,6 +189,26 @@ describe('Content TestCase', function () {
...
+
+            spyOn(document, 'execCommand').and.callThrough();
+
+            placeCursorInsideElement(p, 0);
+
+            fireEvent(p, 'keyup', {
+                keyCode: Util.keyCode.ENTER,
+                ctrlKey: (isSafari() ? true : false),
+                shiftKey: (!isSafari() ? true : false)
+            });
```
daviferreira: Why not:

```
ctrlKey: isSafari(),
shiftKey: !isSafari()
```
Could probabably cache the `isSafari` too :)

:)

daviferreira:
Actually I don't think you need the `isSafari` helper at all. Just set `ctrlKey` as true and `shiftKey` as false.

ilessiivi:
Oh yeah, the `? true : false` is irrelevant. However the check for Safari needs to be there so it won't fail on other browsers, right? It seems to fail at something else than those whitespaces now though...

daviferreira:
The way it is now the `ctrlKey` check works on all browsers since you are not checking for Safari on the `keyUp` handler (which I think it's fine, it should be like that for all browsers in my opinion).

ilessiivi:
Now the test is just to confirm that formatBlock isn't called with ctrlKey. I couldn't get the innerHTML expectation to work (it just never added the `<br>` before lorem ipsum) but is this good enough to merge?

j0k3r:
You choose the right solution. The `<br>` will never be inserted because we only dispatch the event. The action generated for this event is never perform.

I already fought that kind of stuff in #620 :)

Could you [squash your commit](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) into a single one ? And we'll be ready to merge!

PR #625, opened 2015-05-21
Auto link fixes

patrick-webs:
Fix some issues we discovered while testing the new auto-link feature:

* Removed links are again activated when hitting space or enter key
* Autolinking is not working with Tab key.
* Bare hostname with more than three "www" preceding only links final three "www" in the name
* Some links are not detected and some word fragments are detected as if they were links
* Autolink is not working when hitting enter key in IE 11. (link appears and then immediately disappears)


j0k3r:
Could you remove all the css from your PR? I guess it's the default built that generate them but it pollute the PR :)

src/js/events.js
```diff
@@ -220,6 +220,10 @@ var Events;
                 this.setupListener('externalInteraction');
                 this.listeners[name] = true;
                 break;
+            case 'editableUnlink':
+                // Custom event fired from MediumEDitor core code.
```
j0k3r: Typo on MediumE**D**itor

src/js/core.js
```diff
@@ -945,6 +949,37 @@ function MediumEditor(elements, options) {
                 }
             }
 
+            // If the selection is right at the ending edge of a link, put it outside the anchor tag instead of inside.
```
j0k3r: Can all this stuff be moved to the extension instead? Put here just a `if` to test if the extension is enabled and make the call to method in the extension.

src/js/core.js
```diff
@@ -894,7 +898,7 @@ function MediumEditor(elements, options) {
         // http://stackoverflow.com/questions/17678843/cant-restore-selection-after-html-modify-even-if-its-the-same-html
         // Tim Down
         // TODO: move to selection.js and clean up old methods there
-        importSelection: function (inSelectionState) {
+        importSelection: function (inSelectionState, favorLaterSelectionAnchor) {
```
j0k3r: Could you add as a comment an information about what should be `favorLaterSelectionAnchor`, why it is used, in which context and also the value it should receive?

patrick-webs: Added something

src/js/events.js
patrick-webs: thanks, fixed

patrick-webs:
I updated my node_modules and recompiled the css again, looks like that fixes the difference with master on all those CSS files.

src/js/core.js
patrick-webs: I think that core code should not usually depend on extension code directly. I moved most of the code into a helper method now, so it should be easier to read.

nmielnik: Yeah, I agree that we wouldn't want anything core calling out into an extension which may or may not be there.  An alternative would be to have the extension override the 'importSelection' method and catch all calls to it, but I think that seems a bit unnecessary.

I think a separate helper function is a good idea for now, maybe we can improve on this and try to make it more and more generic and reusable over time.

j0k3r: And in any case, as the _todo_ says, we should move that in `Selection`.

nmielnik:
Looks great, thanks @patrick-webs !

PR #627, opened 2015-05-22
Add an option to not remove the placeholder on click

### LINKED ISSUE #534: Hide Placeholder on input/typing
It looks like the original intention was to hide the placeholder only when the user begins typing https://github.com/daviferreira/medium-editor/issues/27#issuecomment-25487543

however this pull request is hiding it on click https://github.com/daviferreira/medium-editor/pull/339 

I agree with the original post that hiding only on input is better UX. Was this not achievable? Could this become an option? I would prefer to show the placeholder on click and only hide it once the user begins to type

j0k3r:
Fix #534 

Also, I've added same custom events on keydown (for tab, enter, delete, etc ..) to keyup

_Note:_ There are a lot of _fake_ diff because I've just re-indented some `switch`.


j0k3r:
Ouch **-0.34%**. Humm

src/js/extensions/placeholder.js
```diff
@@ -58,8 +63,14 @@ var Placeholder;
             // Check placeholder on blur
             this.base.subscribe('editableBlur', this.handleBlur.bind(this));
 
+            // if we don't want the placeholder to be removed on click but when user start typing
+            if (this.hideOnClick) {
+                this.base.subscribe('editableClick', this.handleHidePlaceholderEvent.bind(this));
+            } else {
+                this.base.subscribe('editableKeyupDelete', this.handleBlur.bind(this));
```
nmielnik: Could you try this:
`this.base.subscribe(`editableKeydown`, this.handleHidePlaceholderEvent.bind(this))`

I think we can avoiding creating all of those new custom events and just use that.  I also think we'll be able to get rid of the listener for `editableKeypress` if we went that route.

src/js/events.js
```diff
@@ -196,133 +196,147 @@ var Events;
...
+                    this.listeners[name] = true;
+                    break;
+                case 'editableKeyup':
+                    // Detecting keyup in the contenteditables
+                    this.base.elements.forEach(function (element) {
+                        this.attachDOMEvent(element, 'keyup', this.handleKeyup.bind(this));
+                    }.bind(this));
+                    this.listeners[name] = true;
+                    break;
+                case 'editableKeyupEnter':
```
nmielnik: It would probably be best to not create `editableKeyupEnter` and `editableKeyupTab` until we actually have a use for them.  

I'm also thinking we might be able to avoid creating `editableKeyupDelete` as per my other comment about using `editableKeypress`.

PR #629, opened 2015-05-23
WIP – Cleanup & coverage

j0k3r:


daviferreira:
:ok_hand:

j0k3r:
I didn't find a way to test [these lines](https://github.com/daviferreira/medium-editor/blob/cleanup-coverage/src/js/selection.js#L138-L141):

````javascript
// Selection encompasses a single element
if (this.rangeSelectsSingleNode(range) && range.startContainer.childNodes[range.startOffset].nodeType !== 3) {
    return range.startContainer.childNodes[range.startOffset];
}
```

If anyone got an idea.
I've played with range but didn't find one that match that case.. http://jsfiddle.net/tkatq6Lt/

src/js/util.js
```diff
@@ -275,43 +275,25 @@ var Util;
...
 
         getSelectionRange: function (ownerDocument) {
-            var selection = ownerDocument.getSelection();
-            if (selection.rangeCount === 0) {
-                return null;
-            }
-            return selection.getRangeAt(0);
+            this.deprecated('Util.getSelectionRange', 'Selection.getSelectionRange', 'v5.0.0');
+
+            return Selection.getSelectionRange(ownerDocument);
```
nmielnik: This might be a bit dangerous, I believe Util will be defined before Selection...

src/js/extensions/paste.js
```diff
@@ -176,11 +176,6 @@ var PasteHandler;
                     workEl.innerHTML = workEl.innerHTML.replace(/\n/gi, ' ');
 
                     switch (workEl.tagName.toLowerCase()) {
-                        case 'a':
```
nmielnik: Why did you end up needing to remove this case?  And why did you need to move it to `pasteHTML()`?

nmielnik:
These changes are outstanding, thanks a ton @j0k3r!

Regarding the lines you weren't able to cover, I'm not really sure what they are for either, did you ever figure that out?

daviferreira:
Just remove and wait for someone to open an issue? :bomb:

src/js/extensions/paste.js
j0k3r: Well I've tested on Chrome and Firefox (on my Mac) and `_target` attribute was never created on links, event with the option enabled. When I `console.log` the element (`workEl`) it seems to properly have the attribute, but the dom doesn't have it ..

I then checked the non-multiline support (in the `else`) and see that targetBlank was never called, and so never apply on pasting non-multiline content. That's why I moved the targetBlank `if` in `pasteHTML`. And this function is called on multiline support (line 162).

This way:
- I fixed the element which wasn't properly updated with `_target` attribute on multiline pasting
- I added targetBlank support on non-multiline pasting

src/js/util.js
j0k3r: Well, are you sure? The demo still works great on my Mac and tests seems to be ok with that. Maybe you can give a try on your own to validate there is no such case.

I really prefer to move these functions in Selection.

j0k3r:
@daviferreira I still prefer to let the code like that instead of removing it and wait for an issue :scream:

src/js/util.js
nmielnik: Agreed, they definitely belong in Selection.  I don't believe this code will actually fail, since these functions can't be called until all of the js has loaded.  

What I was pointing out though, if you look at the Gruntfile, we are including the JS files in this order:
```javascript 
srcFiles = [
  'src/js/util.js',
  'src/js/defaults/buttons.js',
  'src/js/defaults/options.js',
  'src/js/extension.js',
  ...
]
```

So we'll be calling into Selection from inside Util, even though Util is defined before Selection.  Just something we'll want to remember.

PR #630, opened 2015-05-23
Deprecate onShowToolbar + onHideToolbar options

nmielnik:


phiggins42:
curious (because i'm using this for having a hook for, when showing the toolbar, change the available buttons based on a traverseup/blacklist/whitelist eg no bold option in headings) what the replacement would be? do the toolbar events ticket supply this functionality? I haven't had a moment to look, this just caught my eye from the title (and lack of context in desc)

dist/js/medium-editor.js
```diff
@@ -4854,7 +4853,10 @@ var Toolbar;
             clearTimeout(this.hideTimeout);
             if (!this.isDisplayed()) {
                 this.getToolbarElement().classList.add('medium-editor-toolbar-active');
+                this.base.trigger('showToolbar', {}, this.base.getFocusedElement());
```
phiggins42: okay ignore me. this should suffice.

PR #631, opened 2015-05-24
Deprecate .parent property of extensions

### LINKED ISSUE #576: Should the .parent option for extensions be deprecated?
Currently, any extension passed into MediumEditor can choose to set a `.parent` boolean on itself.  During setup, if the value of this property is true for any extension, a reference to the MediumEditor instance is assigned to the `.base` property of that extension.  Some more details are in the [documentation](https://github.com/daviferreira/medium-editor/wiki/Custom-Buttons-and-Extensions#parent-boolean--base-mediumeditor)

I'm proposing we get rid of this, and just always set `.base` to an instance of MediumEditor.  I have trouble envisioning scenarios where assigning the instance to `.base` would be problematic (except if the extension already has a `.base` property it's using for something else).

As @phiggins42 pointed out, by assuming that `.base` is always set, we can start adding some helper methods into the extension base class. Some examples:
* `.on()` and `.off()` for attaching DOM Event handlers
* `.subscribe()` and `.unsubscribe()` for custom events
* `.saveSelection()` and `.restoreSelection()`
* or any other method exposed on MediumEditor which we can just add as a helper member to each extension

Thoughts? Suggestions? Angelic Poetry :angel:?

nmielnik:
Fix #576 

This PR deprecates the `.parent` property for extensions.  Now, all extensions will have their `.base` property set to the instance of MediumEditor they are initialized as part of (unless they already have `.base` set to something else or they have the deprecated `.parent` property set to `false`, but this will only be supported until v5.0.0)

How `.parent` and `.base` work:

* BEFORE this PR
  * ONLY IF `.parent` property of an Extension is `true`, MediumEditor sets `.base` to be a reference to itself (MediumEditor)
* AFTER this PR (but before v5.0.0)
  * IF `.parent` property of an Extension is explicitly `false` OR if `.base` property of an Extension is already set to something, MediumEditor will NOT set `.base` property to be a reference to itself (MediumEditor)
  * For all other cases, MediumEditor will set the `.base` property of all Extensions to be a reference to itself (MediumEditor)
  * This will NOT be completely backwards compatible in all cases, since some extensions will now have a `.base` property when they didn't before.  However, since we won't override if it already exists, this should be good enough to not break existing extensions.
* AFTER v5.0.0
  * All Extensions will have their `.base` property set to an instance of MediumEditor, unless they override the `.base` property to something else before initializing MediumEditor.

### DIFF
--- a/demo/pass-instance.html
+++ b/demo/pass-instance.html
@@ -24,8 +24,6 @@
     <script src="../dist/js/medium-editor.js"></script>
     <script>
         function Extension() {
-          this.parent = true;
-
           this.button = document.createElement('button');
           this.button.className = 'medium-editor-action';
           this.button.innerText = 'X';
--- a/spec/extension.spec.js
+++ b/spec/extension.spec.js
@@ -105,7 +105,6 @@ describe('Extensions TestCase', function () {
             var Sub, editor, e1, e2;
 
             Sub = Extension.extend({
-                parent: true,
                 y: 10
             });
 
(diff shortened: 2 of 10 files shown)


j0k3r:
:+1:

daviferreira:
Merged manually, I think :dancer:

PR #633, opened 2015-05-25
Create extension helper methods

nmielnik:
This PR adds some helper methods to all extensions, including accessors into MediumEditor methods + properties.  There are also some minor changes to cleanup tests and remove unneeded defaults from extensions.

* `.on()` + `.off()` -> helpers that call directly into `MediumEditor.on()` + `MediumEditor.off()`
* `.subscribe()` -> calls directly into `MediumEditor.subscribe()`
* `.execAction()` -> calls directly into `MediumEditor.execAction()`
* `.getEditorElements()` + `.getEditorId()` -> accessors for MediumEditor's array of elements and unique id (respectively)
* `.getEditorOption()` -> accessor for specific MediumEditor options

### DIFF
--- a/src/js/extension.js
+++ b/src/js/extension.js
@@ -5,6 +5,10 @@ var Extension;
     /* global Util */
 
     var passThroughHelpers = [
+        // general helpers
+        'execAction',
+
+        // event handling
         'on',
         'off',
         'subscribe'
--- a/src/js/extension.js
+++ b/src/js/extension.js
@@ -4,16 +4,6 @@ var Extension;
 
     /* global Util */
 
-    var passThroughHelpers = [
-        // general helpers
-        'execAction',
-
-        // event handling
-        'on',
-        'off',
-        'subscribe'
-    ];
-
     Extension = function (options) {
         Util.extend(this, options);
     };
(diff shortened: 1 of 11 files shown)


src/js/extension.js
```diff
@@ -193,6 +203,41 @@ var Extension;
...
+         * Helper function which returns the value of an option
+         * used to initialize this instance of MediumEditor
+         */
+        getEditorOption: function (option) {
+            return this.base.options[option];
+        }
     };
+
+    passThroughHelpers.forEach(function (helper) {
+        var methodName = helper;
```
j0k3r: Why don't you call the argument `helper` `methodName` ?

src/js/extension.js
```diff
@@ -4,6 +4,16 @@ var Extension;
 
     /* global Util */
 
+    var passThroughHelpers = [
```
j0k3r: Could you add a short description of the goal of this variable?
Also, it could be helpful to add a description on the `forEach` bellow to explain what is done

src/js/extensions/anchor-preview.js
```diff
@@ -20,11 +20,9 @@ var AnchorPreview;
         previewValueSelector: 'a',
 
         /* ----- internal options needed from base ----- */
-        'window': window,
-        'document': document,
```
j0k3r: `document` & `window` are automatically attached to any extensions now?

nmielnik: Yes indeed.  They're passed from the core editor's `ownerDocument` and `contentWindow` options.

src/js/extension.js
nmielnik: I think I was overthinking scope of closures within loops.  Seems like it works just find using `helper` so I'll go back to that.

Issue #634, opened 2015-05-26
Elements array lost after destroying


nmielnik:
As per @kevincampanella (https://github.com/daviferreira/medium-editor/commit/398743094c370992f87d23842379c1fd3c0eb100#commitcomment-11354089):

> I think this may have broken the setup method, if called after calling destroy. For instance, suppose you call destroy to disable editing, then later call setup to re-enable it. Perhaps only the elements specifically added for textareas should be removed here? Or, is there an alternative way to disable/re-enable editing besides calling destroy/setup?

This is definitely an issue, I have a fix ready

PR #635, opened 2015-05-26
Fix issue where elements array is lost after destroy

nmielnik:
This fixes issue called out in issue #634 

The elements array is emptied after calling `destroy()`, however this means that calls to `setup()` afterwards will not have any elements and thus the editor will be useless.

The fix is to store the selector passed in during initialization, and re-use this same list each time `setup()` is called.


j0k3r:
:+1:

PR #638, opened 2015-05-26
provide toolbarUpdate event hook

phiggins42:
if a user wants to update the toolbar based on selection change the showToolbar and hideToolbar events are insufficient: if a selection is made and toolbar is showing, making another selection in a different region will not show hideToolbar and showToolbar, though the toolbar goes through an update state call in this case. This simply provides an event trigger when that condition happens.

usecase: I am providing dynamic button lists based on the parent wrapping element (eg: no bold in headings). this works fine with showToolbar (determine parent, decide which buttons to show) and hideToolbar (reset to orig), however, I need to reset and redetermine the available buttons when the above selection change-but-not-hiding-toolbar case is hit.


src/js/toolbar.js
```diff
@@ -356,6 +356,7 @@ var Toolbar;
         showAndUpdateToolbar: function () {
             this.modifySelection();
             this.setToolbarButtonStates();
+            this.base.trigger('updateToolbar', {}, this.base.getFocusedElement());
```
nmielnik: Can this go either at the beginning or the end of the logic (before `this.modifySelection()` or after `this.setToolbarPositon()`)?

phiggins42: probably? i feel like there was a compelling case for putting it exactly where it is (this is a direct monkey-patch we've been carrying around locally for the dynamic-button functionality we created), though I couldn't defend the decision as it was made long ago.

phiggins42: probably can't go before modifySelection because that would/could change the focusedElement, possibly. I know it has to go _after_ setToolbarButtonStates because of [something it is doing in there to reset buttons, I don't recall). but in the interest of not having to reposition a 2nd time it must go _before_ setToolbarPosition, because the size of the toolbar is calculated there, so the possibly-shorter-or-longer list of buttons needs to be already created by the time position calc happens. so I'm going to fallback to "yes, maybe, but can think of reasons why it is good to be right here in the middle"

nmielnik: Ok, if that's the case then I would likely rename this event and ideally add some sort of test to make sure these kind of restrictions aren't broken in the future.  Maybe `toolbarButtonStateUpdate` or something like that? This event seems much more targeted than just a toolbarUpdate no?

phiggins42: naming things is hard but that's superoverly verbose, so -1 imo. I can add a test or two, but can't get to it today or tomorrow for sure. I only merged this over because I *believe* it to be the last monkey patch I've been maintaining locally and am attempting to forward port all of our internal app usage to newest editor (post extension and cleanup stuff, which btw thanks for hitting all of our talking points mostly, with .document and .base/.parent et al).  If this won't be merged I'll just extension-in a new version of this 5-line function to get past last of deprecation hurdles and double back on tests and this MR later in the week.

PR #641, opened 2015-05-27
destroy() for extensions

nmielnik:
This PR deprecates the `deactivate()` method on extension and standardizes on the `destroy()` method.  It also implements `destroy()` for auto-link and placeholder extensions.


j0k3r:
:+1:

phiggins42:
just leaving a back compat note: this breaks anyone with an extension that utilized `deactivate`, but also implemented a `destroy` method in the extension.  also people who weren't using `deactivate` in an extension, but had an existing `destroy` method defined. i think. the "had a deactivate but no existing destroy method" people see the apropriate deprecation warning, and nothing explodes.

Issue #639, opened 2015-05-26
Spellcheck stays on after destroy called on editor


nodrog:
The red underlining stays on after destroy is called on the editor. 

Is this something that needs to be fixed in the editor or in the browser.

I am happy to try and fix it if anyone else is experiencing the same issues.

Many thanks

nodrog:
Seems silly to me but after some goggling you need to reset the innerHTML on the elements
Placing:

    element.innerHTML = element.innerHTML 

inside the the elements each of the  destroy method dose the trick.

I am not sure what the overhead of this is?

I have tried other suggestions suck as focusing and then bluring the element, but not work.

nmielnik:
Hey @nodrog thanks for finding this and opening an issue.  If you're able to validate that this fixes the issue you should definitely open a pull request.  That's a pretty interesting fix.

j0k3r:
@nodrog you mean, you tried this solution http://stackoverflow.com/a/9325528/569101 ?

nodrog:
@j0k3r yes tried that and had no success on chrome Version 43.0.2357.81 (64-bit)

nmielnik:
Looks like this was fixed in #643

PR #643, opened 2015-05-27
Fix for spellcheck underlines after destroy

nodrog:
Hi 

This is in relation to issue #639.

I have tried tried my best, have run the tests and they are all good. But i am not sure i have done everything right as i have only changed src/js/core.js and the associated built files. But have managed to cause many other updates.

Would love to know what i have done wrong and fix it.

Also had a look at the test and am not quite sure how to test the new code.

Sorry and many thanks


nmielnik:
So we have a bunch of changes that have been made since the previous release, but we haven't updated the dist files to help avoid merge conflicts until we do another release.

If you don't mind, could you fix this PR to not include the dist files, or just open a new pull request that only includes your changes to core.js?

Since this is a particularly difficult thing to test (I can't think of a good test case to add), it might be ok to not include a new test for this specific change...I'll keep thinking on it, but if you're not able to come up with a good test I don't think it's a reason to block your changes for now.

Thanks again for this.

src/js/core.js
```diff
@@ -636,6 +636,12 @@ function MediumEditor(elements, options) {
             }
 
             this.elements.forEach(function (element) {
+
```
j0k3r: Could you avoid this empty line?

nmielnik:
Thanks @nodrog !

PR #646, opened 2015-05-28
Global cleanup

j0k3r:
One other cleanup before v5 :rocket: 

- avoid `else` condition when it's not necessary
- remove unnecessary empty line
- add some empty line for readability

### DIFF
--- a/src/js/toolbar.js
+++ b/src/js/toolbar.js
@@ -13,7 +13,6 @@ var Toolbar;
     };
 
     Toolbar.prototype = {
-
         // Toolbar creation/deletion
 
         createToolbar: function () {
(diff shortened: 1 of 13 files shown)


j0k3r:
I really hate when a build failed because of @coveralls :hurtrealbad:

src/js/toolbar.js
```diff
@@ -323,25 +319,23 @@ var Toolbar;
...
                 // If we don't have a 'valid' selection -> hide toolbar
                 if (this.options.contentWindow.getSelection().toString().trim() === '' ||
                     (this.options.allowMultiParagraphSelection === false && this.multipleBlockElementsSelected())) {
-                    this.hideToolbar();
-                } else {
-                    this.showAndUpdateToolbar();
+                    return this.hideToolbar();
                 }
+
+                this.showAndUpdateToolbar();
```
nmielnik: Seems like we should probably `return this.showAndUpdateToolbar()` here as well, no?

j0k3r: This is the end of the `checkState` function, so it's not necessary.
And thanks for that, I've just seen an horrible long if inside this function :wink:

PR #647, opened 2015-05-28
Create Util.key to return key code from an event

j0k3r:
Instead of always trying to find which key is associated to an event for all browsers, I created a helper in Util to return this key.

I follow what is already done by jQuery.

It'll ease the transition when the `.key` will be compatible with all browsers (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key)

Are you all ok with the `Util.key(event)` name ? Or should we use `Util.keyCode(event)` ?

### DIFF
--- a/spec/util.spec.js
+++ b/spec/util.spec.js
@@ -342,4 +342,31 @@ describe('Util', function () {
             expect(result).toBe(false);
         });
     });
+
+    describe('isKey', function () {
+        it('should return the key from the event', function () {
+            var event;
+
+            event = {
+                which: 13,
+                keyCode: null,
+                charCode: null
+            };
+            expect(Util.isKey(event, 13)).toBeTruthy();
+
+            event = {
+                which: null,
+                keyCode: 13,
+                charCode: null
+            };
+            expect(Util.isKey(event, [13, 12])).toBeTruthy();
+
+            event = {
+                which: null,
+                keyCode: null,
+                charCode: 13
+            };
+            expect(Util.isKey(event, [65])).toBeFalsy();
+        });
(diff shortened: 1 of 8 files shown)


daviferreira:
I prefer keyCode :)

j0k3r:
Well, it can't be `keyCode`, since `Util.keyCode` is already in use :cry: 

```
keyCode: {
    BACKSPACE: 8,
    TAB: 9,
    ENTER: 13,
    ESCAPE: 27,
    SPACE: 32,
    DELETE: 46
},
```

daviferreira:
Time to create a constants file? :dancer: My beef with `key` is that it may lead to confusion with getting a property name etc. We can go like this now if you want.

j0k3r:
Yeah good point.
We can wait for @nmielnik suggestion :smile: 

What about `getKeyCode`? But won't this be confusing with the `keyCode` constant?

nmielnik:
I don't mean to add yet another opinion, but what about changing the helper to be something like:
`Util.isKey()` and having it take both a single KeyCode or an array of KeyCodes.

```javascript
  // case 1
  if (Util.isKey(event, Util.keyCode.ENTER)) {
    // cunning logic
  }

  // case 2
  if (Util.isKey(event, [Util.keyCode.DELETE, Util.keyCode.BACKSPACE])) {
    // fancy moves
  }
```

j0k3r:
I like that :+1:

nmielnik:
It kills that one case-switch, but I'm ok with that if you guys are

daviferreira:
:+1:

PR #648, opened 2015-05-28
Better way to display medium-editor version on the demo page

j0k3r:
image


index.html
```diff
@@ -39,6 +42,8 @@
         document.getElementById('sel-themes').addEventListener('change', function () {
             cssLink.href = 'bower_components/medium-editor/dist/css/' + this.value + '.css';
         });
+
+        document.getElementById('version').innerHTML = 'This demo uses medium-editor v'+MediumEditor.version.toString();
```
phiggins42: the casting makes the implicit toString happen. not a big deal really. just wanted to see how you'd used it.

j0k3r: Yep, I first would like to display without casting, so I let the `.toString()` after :)

phiggins42: note sure I follow. I'm saying you are explicitly calling something that would be called by the type casting of `"this demo uses v" + MediumEditor.version;`

https://github.com/daviferreira/medium-editor/blob/master/src/js/version.js#L8

The only time you'd need to dive into `version` is if you were interested in discrete `major`, `minor`, or `patch` values (as numbers)

PR #651, opened 2015-05-28
implement updateToolbar

phiggins42:
* and tests to ensure it works
* and test to ensure is fired before positionToolbar


phiggins42:
#638 reworked as this. seems a bit odd the need to register the custom event in this pseudo pubsub model? added tests to ensure.

nmielnik:
Based on your test case, I would call this event something like `positonToolbar`, but definitely not `updateToolbar`. Changing the names of custom events later won't be something easy to do, so it's going to be better to be more specific now.

PR #653, opened 2015-05-29
Don't execute button shortcut when shift key is pressed

daviferreira:
I want `ctrl+shift+I` to open devtools :P


nchase:
nice. I have been wanting this too!

PR #654, opened 2015-05-29
Make custom events not require 'registering' before triggering

nmielnik:
Yeah, I don't know why I did this originally, but it's not great to require custom events to be defined before they can trigger.  This fixes that.


j0k3r:
:+1:

PR #655, opened 2015-05-30
remove name of function.

phiggins42:
* none of the other functions do? or did?
* now that everything is Extension.extend, no named functions are used as ctors
* somehow, someway, the duplicate `var Toolbar; (function(){ Toolbar = function Toolbar(instance){} )();`
  is causing uglify2 to output `var x; (function() e = function e(e){})();` which is clearly not right

### DIFF
--- a/src/js/toolbar.js
+++ b/src/js/toolbar.js
@@ -5,7 +5,7 @@ var Toolbar;
 (function () {
     'use strict';
 
-    Toolbar = function Toolbar(instance) {
+    Toolbar = function (instance) {
         this.base = instance;
         this.options = instance.options;
 


phiggins42:
i'm trying to make a minimal repro. eg:

```
(function(){
    var Foo;
    (function(){
        'use strict';
        Foo = function Foo(bar){
            this.baz = bar;
        }
    })();
})();
```

outputs

```
'!function(){var n;!function(){"use strict";n=function(n){this.baz=n}}()}();'
```

but I'm experiencing the issue in a larger built file that includes medium-editor.js and passes that through uglify2 ... getting an error after new MediumEditor(...) when Toolbar is initialized, because the prototype is being set to a wrong variable:

screen shot 2015-05-30 at 12 18 34 am

such weird. I'm going to keep trying to reproduce minimally, but have tracked down the exception I'm seeing ("this.initThrottleMethods undefined" during init of editor) to the bad var assignment by uglify2 and confirmed success after removing the name part of the function.

phiggins42:
note this is unnecessary if toolbar were converted to an `Extension.extend` too, as the floating named function would be gone there too.

j0k3r:
Good investigation :+1: 
As far as I remember we plan to convert Toolbar as an extension.

PR #656, opened 2015-05-31
CSS/SASS refactor

daviferreira:
This PR includes a couple of breaking changes I think, but I want to have a `medium-editor` namespace on all our css classes/attrs.

* `medium-element` becomes `medium-editor-element`
* `sticky-toolbar` becomes `medium-editor-sticky-toolbar`
* `stalker-toolbar` becomes `medium-editor-stalker-toolbar`

Maybe we should still keep the old classes for now? Not sure.

The rest was mainly creating components and organizing editor/toolbar components. There is more room for improvement, but this is a start.


j0k3r:
If it's not too hard maybe we can keep old classes name and remove them when realizing v5

phiggins42:
+1 @j0k3r .. these changes are pretty breaking if you had any use of anything like "the top level element is called `[data-medium-element]` (like a custom traverseup? or basically anything). also any custom themeing goes out the window. I vote keep this out until 5.0, or rather, use this as the first patch set towards 5.0, releaseing whatever is in right now as the last in the 4.x tree (or, even better, use proper branch strategy to allow the two to live simultaneously for a while)

nmielnik:
Yeah, with 5.0 coming soon (hopefully very soon), I think we should either keep the old classes along with the new ones, or hold off on these changes until they can get lumped into the v5.0.0 release.

src/sass/components/_anchor-preview.scss
```diff
@@ -0,0 +1,24 @@
...
+    top: 0;
+    word-break: break-all;
+    word-wrap: break-word;
+    visibility: hidden;
+    z-index: $z-toolbar;
+
+    a {
+        color: #fff;
+        display: inline-block;
+        margin: 5px 5px 5px 10px;
```
nchase: is there any reason these weren't condensed to `margin: 5px 5px 10px`?

src/sass/components/_anchor-preview.scss
```diff
@@ -0,0 +1,24 @@
+.medium-editor-anchor-preview {
+    font-family: $font-sans-serif;
+    font-size: 16px;
```
nchase: in web browsers, the `font-size` is equal to this value. Can this be a `font-size: inherit` instead in the interest of making this an easy drop-in solution on the CSS side?

src/sass/components/_placeholder.scss
```diff
@@ -0,0 +1,12 @@
+.medium-editor-placeholder {
+    position: relative;
+
+    &:after {
+        content: attr(data-placeholder) !important;
```
nchase: what is this `!important` overriding?

nchase:
this is going to be awesome for 5.x.x :thumbsup:

src/sass/components/_placeholder.scss
daviferreira: I have no idea, but there must be a reason :P

src/sass/components/_anchor-preview.scss
daviferreira: Same as toolbar, the main container. I think we should leave it for now, but font-sizes and margin/paddings is something I would like to improve.

PR #664, opened 2015-06-01
Convert Toolbar into an extension

nmielnik:
### Summary
This PR covers converting the toolbar into a built-in extension, that is extendable and override-able.

Now the following options will be passed as sub-options inside the toolbar options group:
```js
new MediumEditor('.editor', {
  toolbar: {
    buttons: // array of button names,
    align: // left | center | right (previously toolbarAlign),
    diffLeft: // left offset of toolbar,
    diffTop: // top offset of toolbar,
    firstButtonClass: // class added to first button in toolbar,
    lastButtonClass: // class added to last button in toolbar,
    static: // static positioning of toolbar (previously staticToolbar),
    sticky: // sticky positioning of toolbar (previously stickyToolbar),
    updateOnEmptySelection: // updating static toolbar on collapsed selection
  }
});
```

Further, the toolbar will be disabled by passing `false` as the `toolbar` option:
```js
new MediumEditor('.editor', {
  toolbar: false
});
```

### Additional Changes
In order to separate the close dependency between MediumEditor and the toolbar, there were some extra changes that were needed:
* Maintain a global array of editor instances to use for generating a unique id for each editor
  * Previously the editor depended on counting the number of toolbars on the page to generate a unique-id
* Expose a new `addBuiltInExtension()` method on MediumEditor.
  * This allows the Toolbar to completely control creating and maintaining the buttons that go in the toolbar, even though they live in the `.extensions` array inside MediumEditor
* Create a new `KeyboardCommands` extension which handles the <kbd>CTRL</kbd> + <kbd>B</kbd>, <kbd>CTRL</kbd> + <kbd>I</kbd>, and <kbd>CTRL</kbd> + <kbd>U</kbd> behavior.
  * This was originally handled by the buttons, but now that the buttons are completely owned by the toolbar, this behavior should be allowed to exist even if the toolbar or buttons don't exist

### Why are there so many changes?
* The toolbar was referenced directly via `editor.toolbar` in an immense amount of places in the test code.  Thus, there's a ton of find-and-replace now that the toolbar has to be retrieved via a call to `.getExtensionByName()`
* Converting how 9 toolbar options are passed into medium editor means updating a ton of the demos and tests as well
* Converting the toolbar into an extendable extension took a decent amount of work!

##### NOTE: Extensive wiki updates for the buttons and extensions documentation will be coming in the v5.0.0 release


src/js/core.js
```diff
@@ -370,81 +373,41 @@ function MediumEditor(elements, options) {
...
             }
         }, this);
 
-        for (name in extensions) {
-            if (extensions.hasOwnProperty(name) && buttons.indexOf(name) === -1) {
-                ext = initExtension(extensions[name], name, this);
-                this.commands.push(ext);
+        // Built-in extensions
+        var builtIns = {
+            paste: true,
```
j0k3r: Shouldn't we have a `isPasteEnabled` ?

nmielnik:
We could, but the tricky thing with paste is that we have 2 methods (`cleanPaste()` + `pasteHTML()`) which require the paste handler to exist.  Without breaking the pasteHandler into 2 extensions (one for inserting HTML, 1 for handling pastes) I wasn't sure of another way to handle this other than always enforcing that a pasteHandler be present...

src/js/core.js
nmielnik: We could, but the tricky thing with paste is that we have 2 methods (`cleanPaste()` + `pasteHTML()`) which require the paste handler to exist.  Without breaking the pasteHandler into 2 extensions (one for inserting HTML, 1 for handling pastes) I wasn't sure of another way to handle this other than always enforcing that a pasteHandler be present...

src/js/core.js
```diff
@@ -158,6 +158,38 @@ function MediumEditor(elements, options) {
...
+            win._mediumEditors = [null];
+        }
+
+        if (this.id) {
+            // If this already has a unique id, re-use it
+            win._mediumEditors[this.id] = this;
+        } else {
+            this.id = win._mediumEditors.length;
+            win._mediumEditors.push(this);
+        }
```
j0k3r: Is that snippet equal to what you've done ?

```javascript
if (!this.id) {
    this.id = win._mediumEditors.length;
}

win._mediumEditors[this.id] = this;
```

src/js/core.js
```diff
@@ -215,57 +247,43 @@ function MediumEditor(elements, options) {
...
-    function shouldAddDefaultAnchorPreview() {
-        var i,
-            shouldAdd = false;
-
-        // If anchor-preview is disabled, don't add
-        if (this.options.anchorPreview === false) {
+    function isToolbarEnabled() {
+        // If any of the elements don't have the toolbar disabled
+        // We need a toolbar
+        if(this.elements.every(function (element) {
```
j0k3r: You should have one space between `if` and `(`

src/js/core.js
```diff
@@ -370,81 +373,41 @@ function MediumEditor(elements, options) {
...
+        if (!toolbarExtension && isToolbarEnabled.call(this)) {
+            toolbarExtension = new MediumEditor.extensions.toolbar(this.options.toolbar);
         }
 
-        if (shouldAddDefaultPlaceholder.call(this)) {
-            var placeholderOpts = (typeof this.options.placeholder === 'string') ? {} : this.options.placeholder;
-            this.commands.push(initExtension(initPlaceholder.call(this, placeholderOpts), 'placeholder', this));
+        if (toolbarExtension) {
+            this.extensions.push(initExtension(toolbarExtension, 'toolbar', this));
         }
```
j0k3r: Does this mean that `new MediumEditor.extensions.toolbar(this.options.toolbar)` can be falsy?

Otherwise, why don't you simplify things to:

```javascript
var toolbarExtension = this.options.extensions['toolbar'];
if (!toolbarExtension && isToolbarEnabled.call(this)) {
    this.extensions.push(initExtension(new MediumEditor.extensions.toolbar(this.options.toolbar), 'toolbar', this));
}
```

src/js/core.js
```diff
@@ -643,6 +596,54 @@ function MediumEditor(elements, options) {
...
+                    extension = new MediumEditor.extensions.keyboardCommands(this.options.keyboardCommands);
+                    break;
+                case 'paste':
+                    extension = new MediumEditor.extensions.paste(this.options.paste);
+                    break;
+                case 'placeholder':
+                    extension = new MediumEditor.extensions.placeholder(this.options.placeholder);
+                    break;
+                default:
+                    if (ButtonsData.hasOwnProperty(name)) {
```
j0k3r: Could you add a short comment saying that every buttons (*bold*, *italic*, etc..) is loaded as an extension?
I says that because it wasn't clear for me at first read

src/js/core.js
```diff
@@ -643,6 +596,54 @@ function MediumEditor(elements, options) {
...
         /**
+         * NOT DOCUMENTED - exposed as a helper for other extensions to use
+         */
+        addBuiltInExtension: function (name) {
+            var extension = this.getExtensionByName(name);
+            if (extension) {
+                return extension;
+            }
+
+            switch(name) {
```
j0k3r: Missed a space between `switch` and `(`

src/js/extensions/auto-link.js
```diff
@@ -30,6 +30,8 @@ LINK_REGEXP_TEXT =
 
     AutoLink = Extension.extend({
         init: function () {
+            Extension.prototype.init.apply(this, arguments);
```
j0k3r: Is this something like *calling the parent* in case that if in the future the default `init` function in the parent may actually do something, it will be automatically done by each children?
Because at the moment, `Extension.prototype.init` doesn't do anything: `init: function () {},`

src/js/extensions/keyboard-commands.js
```diff
@@ -0,0 +1,71 @@
...
+                if (!this.keys[keyCode]) {
+                    this.keys[keyCode] = [];
+                }
+                this.keys[keyCode].push(command);
+            }, this);
+        },
+
+        handleKeydown: function (event) {
+            var keyCode = Util.getKeyCode(event);
+            if (this.keys[keyCode]) {
```
j0k3r: I like to avoid non-necessary extra level of indentation by avoiding single if.

I mean, I prefer :
```javascript
if (!this.keys[keyCode]) {
    return;
}
```

Instead of a big `if`.

src/js/extensions/keyboard-commands.js
```diff
@@ -0,0 +1,71 @@
...
+                meta: true,
+                shift: false
+            },
+            {
+                command: 'underline',
+                key: 'u',
+                meta: true,
+                shift: false
+            }
+        ],
```
j0k3r: Is it easy to override these options when creating an instance of MediumEditor ?

src/js/extensions/keyboard-commands.js
```diff
@@ -0,0 +1,71 @@
...
+            var keyCode = Util.getKeyCode(event);
+            if (this.keys[keyCode]) {
+                var isMeta = Util.isMetaCtrlKey(event),
+                    isShift = !!event.shiftKey;
+                this.keys[keyCode].forEach(function (data) {
+                    if (data.meta === isMeta &&
+                        data.shift === isShift) {
+                        event.preventDefault();
+                        event.stopPropagation();
+                        this.execAction(data.command);
```
j0k3r: I was wondering if we can check that `data.command` isn't falsy before calling `execAction`?

I mean, what if I want to completely disable <kbd>CTRL</kbd> + <kbd>U</kbd> ?
If I can override default commands options, I will only write *bold* and *italic*. But browsers may handle <kbd>CTRL</kbd> + <kbd>U</kbd> by default (like they did for bold and italic – *at least Chrome*). By giving a `command: false` I can manually disable this shortcut.

What do you think?

src/js/selection.js
```diff
@@ -77,14 +77,14 @@ var Selection;
 
         // http://stackoverflow.com/questions/4176923/html-of-selected-text
         // by Tim Down
-        getSelectionHtml: function getSelectionHtml() {
+        getSelectionHtml: function getSelectionHtml(doc) {
```
j0k3r: If someone else use `getSelectionHtml` before your change, since the signature of the function has changed, it will breaks.
Could you fallback to `this.options.contentWindow` in case of `doc` is `undefined`?

src/js/core.js
nmielnik: Indeed it is, I'll condense it.

src/js/core.js
nmielnik: Fixed (I'll see if I can update the jscs to catch this)

src/js/core.js
nmielnik: The if statement can indeed be removed, but the call to `this.extensions.push()` and `initExtension()` will still need to happen for both the case where we create the Toolbar, or if the toolbar was passed in as a custom extension.

j0k3r: Oh I see, I didn't understand why you did that `if`. It's because you can pass a custom extension called `toolbar`. All right :)
You should add a comment about that

src/js/extensions/keyboard-commands.js
nmielnik: Yeah, it should be.  I added a test case that does just that (see `keyboard-commands.spec.js`)

src/js/extensions/auto-link.js
nmielnik: Yeah, I added this just about everywhere for that exact reason.  I realized this because I needed the anchor extension to be able to override `init()` so it could listen for the keyboard shortcut, but I still wanted to make sure init got called in `FormExtension` in case something got added in there in the future (instead of calling straight to `Button.init()` which would skip a level of 'inheritance')

src/js/extensions/keyboard-commands.js
j0k3r: And you already added doc in `README.md` my bad!

src/js/core.js
nmielnik: :+1:

src/js/core.js
nmielnik: :+1:

src/js/core.js
nmielnik: I've updated our jscs config so now it complains about this as well.  There were only 2 issues and you managed to find both of them, so nice work :smile:

j0k3r: :grin: :+1:

src/js/extensions/keyboard-commands.js
nmielnik: :+1:

src/js/extensions/keyboard-commands.js
nmielnik: I think this is a great idea and this idea has been brought up by other people as well.  I don't think it should be part of this specific PR though, perhaps a separate PR?

j0k3r: I'm ok with that !

src/js/selection.js
nmielnik: We're heading toward 5.0.0, so we can break it :broken_heart:

j0k3r:
Good to merge for me :+1:

daviferreira:
:+1: 

you guys rock :guitar:

PR #669, opened 2015-06-04
Remove Unused Methods

nmielnik:
This PR removes some old methods that we aren't currently using and which would be good to remove since we're prepping v5.0.0 right now.

* `MediumEditor.callExtensions()`
* `MediumEditor.util.getObject()`
* `MediumEditor.util.setObject()`

### DIFF
--- a/spec/extension.spec.js
+++ b/spec/extension.spec.js
@@ -26,29 +26,6 @@ describe('Extensions TestCase', function () {
             expect(editor.options.extensions).toBe(extensions);
         });
 
-        it('should call methods on all extensions with callExtensions is used', function () {
-            var Extension = function () {},
-                ext1 = new Extension(),
-                ext2 = new Extension(),
-                editor = this.newMediumEditor('.editor', {
-                    extensions: {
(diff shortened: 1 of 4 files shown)


phiggins42:
we've removed all usage of `util.derives` internally, I think? The .extend method can be placed on any contstructor from the ref at `MediumEditor.Extension.extend` eg:

```
var MyCtor = function(options){};
MyCtor.extend = MediumEditor.Extension.extend
````

I would almost prefer that the `.extend` function exist on MediumEditor ctor itself. so the above would be:

```
function MyCtor(){}
MyCtor.extend = MediumEditor.extend;
````

while also adding the benefit of allowing

```
var MyCustomEditor = MediumEditor.extend({
// perhaps I want to duckpunch a core editor method?
});
```

nmielnik:
Good point, I've removed `uilt.derives()`

For adding extend to MediumEditor, seems like it should be a different PR and it could possibly wait until after 5.0.0...I agree it's a good idea and shouldn't have much impact on anything else.

PR #670, opened 2015-06-04
Expose button default config (replace MediumEditor.statics.ButtonsData)

nmielnik:
Before version 5.0.0, the set of default button configuration was exposed via `MediumEditor.statics.ButtonsData`.  This contained all the config options to differentiate the various built-in MediumEditor buttons (ie `bold`, `italic`, `underline`, etc.)

To give users the ability to access the data, but still align with how our other options and defaults are exposed, I've exposed this set of data via the button prototype (`MediumEditor.extensions.button.prototype.defaults`).

Also, to help reduce the cases where this data might be needed, I've added 2 additional features:

1. The button extension constructor can be invoked and passed the name of a built-in button (instead of an object containing all the button options).  For that case, it will initialize itself with the default config data for that button.
2. A `isBuiltInButton()` method is now exposed on the button extension constructor itself (statically).  Now, calls to `MediumEditor.extensions.button.isBuiltInButton(name)` will return true or false if the text describes a built-in button:
  * `MediumEditor.extensions.button.isBuiltInButton('bold')` returns `true`
  * `MediumEditor.extensions.button.isBuiltInButton('blah')` returns `false`

I've also added documentation that describes what all of the options for the button extension do.

### DIFF
--- a/src/js/extensions/button.js
+++ b/src/js/extensions/button.js
@@ -2,9 +2,98 @@ var Button;
 (function () {
     'use strict';
 
-    /*global Extension */
+    /*global Extension, buttonDefaults */
 
     Button = Extension.extend({
+
+        /* Button Options */
+
+        /* action: [string]
+         * The action argument to pass to MediumEditor.execAction()
+         * when the button is clicked
+         */
+        action: undefined,
+
+        /* aria: [string]
+         * The value to add as the aria-label attribute of the button
+         * element displayed in the toolbar.
+         * This is also used as the tooltip for the button
+         */
+        aria: undefined,
+
+        /* tagNames: [Array]
+         * NOTE: This is not used if useQueryState is set to true.
+         *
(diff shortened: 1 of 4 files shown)


daviferreira:
Nice :beers:

src/js/extensions/button.js
```diff
@@ -2,10 +2,99 @@ var Button;
...
+        // (ie 'bold', 'italic', etc.)
+        // When the name of a button is passed, it will initialize itself with the
+        // configuration for that button
+        constructor: function (options) {
+            if (Button.isBuiltInButton(options)) {
+                Extension.call(this, this.defaults[options]);
+            } else {
+                Extension.call(this, options);
+            }
+        },
```
j0k3r: Is there a case in specs where you test this behavior?

nmielnik: There wasn't before...but there are now!

PR #671, opened 2015-06-05
Redesign Header Buttons

nmielnik:
This PR is a proposal on how to handle configuration for the header buttons in MediumEditor.

screen shot 2015-06-05 at 8 54 44 am

Proposal:
* There are up to 3 different header buttons which can be enabled in the toolbar (this aligns with what medium actually supports)
* These 3 buttons are configured by passing the corresponding button names into the `buttons` array of the `toolbar` options:
  * `'h1'` for the <kbd>H1</kbd> Button
  * `'h2'` for the <kbd>H2</kbd> Button
  * `'h3'` for the <kbd>H3</kbd> Button
* By default, the 3 header tags will generate the following when clicked (this is what the actual medium.com editor does as well):
  * The <kbd>H1</kbd> Button will generate a `<h2>` tag
  * The <kbd>H2</kbd> Button will generate a `<h3>` tag
  * The <kbd>H3</kbd> Button will generate a `<h4>` tag
* The html tags generated by the 3 header buttons are configurable via a new `headerTags` option.  This option is an array of strings that are passed in via the `toolbar` options.
  * Example of changing the generated tags to be `<h1>`, `<h3>`, and `<h5>`:
```js
new MediumEditor('.editor', {
  toolbar: {
    buttons: ['bold', 'italic', 'quote', 'h1', 'h2', 'h3'],
    headerTags: ['h1', 'h3', 'h5']
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ var editor = new MediumEditor('.editable', {
 * __diffLeft__: value in pixels to be added to the X axis positioning of the toolbar. Default: `0`
 * __diffTop__: value in pixels to be added to the Y axis positioning of the toolbar. Default: `-10`
 * __firstButtonClass__: CSS class added to the first button in the toolbar. Default: `'medium-editor-button-first'`
-* __headerTags__: the types of html elements to create when the <kbd>H1</kbd>, <kbd>H2</kbd>, or <kbd>H3</kbd> toolbar buttons are clicked.  The first element in the array (index 0) indicates which tag name (ie `<h3>`) will be used when the <kbd>H1</kbd> button is clicked.  The second element in the array (index 1) corresponds to the <kbd>H2</kbd> button and the last element (index 2) corresponds to the <kbd>H3</kbd> button. Default: `['h2', 'h3', 'h4']`
+* __headerTags__: the types of html elements to create when the <kbd>H1</kbd>, <kbd>H2</kbd>, or <kbd>H3</kbd> toolbar buttons are clicked.  The first element in the array (index 0) indicates which tag name (ie `<h2>`) will be used when the <kbd>H1</kbd> button is clicked.  The second element in the array (index 1) corresponds to the <kbd>H2</kbd> button and the last element (index 2) corresponds to the <kbd>H3</kbd> button. Default: `['h2', 'h3', 'h4']`
 * __lastButtonClass__: CSS class added to the last button in the toolbar. Default: `'medium-editor-button-last'`
 * __standardizeSelectionStart__: enables/disables standardizing how the beginning of a range is decided between browsers whenever the selected text is analyzed for updating toolbar buttons status. Default: `false`
 * __static__: enable/disable the toolbar always displaying in the same location relative to the medium-editor element. Default: `false`
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ var editor = new MediumEditor('.editable', {
         diffLeft: 0,
         diffTop: -10,
         firstButtonClass: 'medium-editor-button-first',
-        headerTags: ['h3', 'h4', 'h5'],
+        headerTags: ['h2', 'h3', 'h4'],
         lastButtonClass: 'medium-editor-button-last',
         standardizeSelectionStart: false,
         static: false,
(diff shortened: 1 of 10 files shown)


phiggins42:
why 3? by default, sure. an artificial limit on doing `headerTags:[h1,h2,h3,h4,h5,h6]` seems lame?

nmielnik:
So medium has exactly 3 header options by default, and it doesn't change as far as I can tell, so this keeps the editor in line with what it's named after :smile: 

However, if you look at the code, there isn't actually anything restricting you from putting as many headers as you want...the only thing restricting you is the documentation.

daviferreira:
So we can actually use the `headerTags` option to create any tag supported by `formatBlock` :P

README.md
```diff
@@ -115,10 +113,11 @@ var editor = new MediumEditor('.editable', {
     toolbar: {
         /* These are the default options for the toolbar,
            if nothing is passed this is what is used */
-        buttons: ['bold', 'italic', 'underline', 'anchor', 'header1', 'header2', 'quote'],
+        buttons: ['bold', 'italic', 'underline', 'anchor', 'h1', 'h2', 'quote'],
         diffLeft: 0,
         diffTop: -10,
         firstButtonClass: 'medium-editor-button-first',
+        headerTags: ['h2', 'h3', 'h4'],
```
j0k3r: `h4` isn't necessary here, right? Since we only enabled `h1` & `h2` in buttons

README.md
```diff
@@ -131,10 +130,11 @@ var editor = new MediumEditor('.editable', {
 });
 ```
 
-* __buttons__: the set of buttons to display on the toolbar. Default: `['bold', 'italic', 'underline', 'anchor', 'header1', 'header2', 'quote']`
+* __buttons__: the set of buttons to display on the toolbar. Default: `['bold', 'italic', 'underline', 'anchor', 'h1', 'h2', 'quote']`
 * __diffLeft__: value in pixels to be added to the X axis positioning of the toolbar. Default: `0`
 * __diffTop__: value in pixels to be added to the Y axis positioning of the toolbar. Default: `-10`
 * __firstButtonClass__: CSS class added to the first button in the toolbar. Default: `'medium-editor-button-first'`
+* __headerTags__: the types of html elements to create when the <kbd>H1</kbd>, <kbd>H2</kbd>, or <kbd>H3</kbd> toolbar buttons are clicked.  The first element in the array (index 0) indicates which tag name (ie `<h3>`) will be used when the <kbd>H1</kbd> button is clicked.  The second element in the array (index 1) corresponds to the <kbd>H2</kbd> button and the last element (index 2) corresponds to the <kbd>H3</kbd> button. Default: `['h2', 'h3', 'h4']`
```
j0k3r: Index 0 isn't `h2` instead of `h3`? And index 1 is `h3`, finally index 2 is `h4`. Or I'm totally wrong ? :confused:

nmielnik: The previous defaults would have mapped to this:
`headerTags: ['h3', 'h4']`
I've updated the behavior to match medium, so now it's:
`headerTags: ['h2', 'h3', 'h4']`

Is that what you are asking about?

README.md
nmielnik: If we agree we want to have 3 header buttons by default, I'll update the `buttons` array to include a 'h3' as well.  If we don't want 3 header buttons, I think we should leave this as-is, because these defaults don't necessarily have to be tied to each other.  A user could choose to enable all 3 header buttons, and if they do, it'd be nice if it worked by default (without requiring them to override the `headerTags` option).  What do you think?

README.md
j0k3r: I was talking about the text instead of the default value.

README.md
j0k3r: Well, then why don't we define all available `headerTags` by default :smile:

README.md
nmielnik: Are you talking about the example?
> (ie `<h3>`)

I can fix that to say `<h2>` if that's what you're asking about

j0k3r: Yeah, that's it!

README.md
nmielnik: The reason why I wouldn't want to put all available header tags is that we want the first header to be a `<h2>` and not a `<h1>`.  Also, it'd be good if the 'H4' button was NOT supported OOB, but was still possible.

To match medium, we need the 'H1' button to produce a `<h2>`, which would mean we wouldn't want to put all the headers in there.  Using `headerTags: ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']` wouldn't match medium since we'd have to start with a `<h1>`.  

By putting all 3 header tags in the array, that means the 'H3' button will work OOB, the user just needs to add 'h3' to the buttons array.  

Also, but only putting these 3 options, an 'H4' button won't work unless the user changes the `headerTags` array somehow.  Based on this reasoning I landed on supplying exactly 3 default `headerTags`...

j0k3r: Ok all right, let `headerTags` like this.

nmielnik: Of course, someone else could always weigh in on this...ehm...@daviferreira

daviferreira:
I'm starting to question if we really need a firstHeader/headerTags etc. option... just let them be buttons?

nmielnik:
So you're saying just have any of the 6 buttons be supported, but they always reflect the kind of header they are?  So you could have just <kbd>H2</kbd>, <kbd>H4</kbd>, and <kbd>H6</kbd> buttons in your toolbar which produce `<h2>`, `<h4>`, and `<h6>` element respectively?

daviferreira:
Yes, and you enable whatever you want... I think I lost that battle :)

nmielnik:
Intersting... @j0k3r @nchase @phiggins42 , would you want to abandon the idea of trying to match medium and just go with this approach where the buttons actually reflect the type of header element you'll be adding?

j0k3r:
Well, I know I will have to hack that to display something else as label. We display "Inter" to create a "h3".
Maybe I can create my own button .. 
But it's sad to lose the customisation between the label and the action.

daviferreira:
You have that problem right now for all other buttons, why should it be different for headings?

j0k3r:
My problem is just about heading atm.



Bises

--
Jeremy

nmielnik:
@j0k3r with the new extension model, it actually will be fairly easy to do what you described:

```js
// HeaderOne is a button that creates/removes a `<h1>`
// with the button text being 'Inter'
var HeaderOne = MediumEditor.extensions.button.extend({
  name: 'header1',
  action: 'append-h1',
  aria: 'h1',
  tagNames: ['h1'],
  contentDefault: '<b>Inter</b>'
});

new MediumEditor('.editor', {
  toolbar: {
    buttons: ['header1']
  },
  extensions: {
    'header1': new HeaderOne()
  }
});
```

daviferreira:
I meant you can't have custom labels for default buttons. I think we should fix that instead of having a specific option for headings (if we want, of course :))

nmielnik:
So, I would like to remove the `firstHeader` and `secondHeader` options, are we ok with this current approach for now?

j0k3r:
:+1:

nmielnik:
Closing this for now, I'm gonna try another approach to handling headers

PR #674, opened 2015-06-08
Update README to link to Wiki for options + API documentation

nmielnik:
There is now more detailed and thorough documentation in place via the Wiki, which is a superset of the details in the README.  I was thinking it might be better to have a shorter README and provide links to the pages which have documentation. 

Do you all agree or do you think it's better to have a longer, one-stop-shop style README which include detailed documentation and which is maintained via source control?

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,171 +92,13 @@ MediumEditor also supports textarea. If you provide a textarea element, the scri
 
 People have contributed wrappers around MediumEditor for integrating with different frameworks and tech stacks.  Take a look at the list of existing [Wrappers and Integrations](https://github.com/yabwe/medium-editor/wiki/Wrappers-and-Integration) that have already been written for MediumEditor!
 
-## Initialization options
+## MediumEditor Options
 
-### Core options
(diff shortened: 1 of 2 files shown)


j0k3r:
I like to keep doc in the repo since it's easier to maintain an keep tracks on what change related to which PR. I mean, It's hard to track change in the repo which can affect the wiki.

Wiki on Github is great, but I think we should have a proper markdown documentation. I know there is [Couscous](http://couscous.io/) for PHP project but it might have the same kind of documentation builder for JS project.

phiggins42:
I like keeping docs as close to the source as possible.Wikis are great for tutorials and longwinded examples, but they so often fall stale. A proper auto-generated MD tree (or a manually maintained for that matter) in-source way easier to maintain and track changes on.

nmielnik:
Ok, makes sense to me.  I've added the documentation on the options and API methods back into the README.  

I've added some updates to the buttons documentation in the README, as well as rearranged the list of API methods (as well as added 1 or 2 that were missing).

PR #677, opened 2015-06-10
Custom Button Options + Deprecate Header Options

### LINKED ISSUE #69: Support for Heading Tags <h1> through <h6>
While the current `header1` and `header2` for heading tags makes sense in situations where I have to give options for users to create headings but want them to use only few headings.

But, there are some limitations to current approach.
* Currently only two headers are supported
* It nags me everytime I click on H1 knowing its really `<h3>` that I'm entering

While I actually think `header1` and `header2` approach has its advantages, I have two suggestions that may improve current situation

1. Increase the number of `headers` supported to about four.
2. Give access to low level heading tags with different set of option labels, probably like `<h1>` `<h2>` `<h3>` etc.

PS: I love this project. Thanks a ton for making this wonderful editor.

nmielnik:
This PR solves two button-related issues in one.

1. The `firstHeader` and `secondHeader` options were not ideal as options on MediumEditor.  They also restricted the amount of header buttons that could be easily configured.
2. Making small tweaks to existing buttons required the creation of a new Extension which would need to extend from `MediumEditor.extensions.button` and would need to know how to use the same set of default options that the built-in button would use.

Changes in this PR:
* Add built-in button support for all header types (`h1`, `h2`, `h3`, `h4`, `h5`, and `h6`)
* Remove support for the `firstHeader` and `secondHeader` options.
* Allow for overriding built-in button options by accepting both strings and objects into the `options.toolbar.buttons` array passed in during initialization
* Add `classList` property to button for adding custom classes to built-in button elements
* Add `attrs` property to button for adding custom attributes to built-in button elements

Example of usage:
```javascript
var editor = new MediumEditor('.editable', {
    toolbar: {
        buttons: [
            'bold', 
            'italic',
            {
                name: 'h1',
                action: 'append-h2',
                aria: 'header type 1',
                tagNames: ['h2'],
                contentDefault: '<b>H1</b>',
                classList: ['custom-class-h1'],
                attrs: {
                    'data-header-type': 'h1'
                }
            },
            {
                name: 'h2',
                action: 'append-h3',
                aria: 'header type 2',
                tagNames: ['h3'],
                contentDefault: '<b>H2</b>',
                classList: ['custom-class-h2'],
                attrs: {
                    'data-header-type': 'h2'
                }
            },
            'justifyCenter',
            'quote',
            'anchor'
        ]
    }
});
```

Should fix #69

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ MediumEditor also supports textarea. If you provide a textarea element, the scri
 ### Core options
 * __activeButtonClass__: CSS class added to active buttons in the toolbar. Default: `'medium-editor-button-active'`
 * __allowMultiParagraphSelection__: enables the toolbar when selecting multiple paragraphs/block elements. Default: `true`
-* __buttonLabels__: type of labels on the buttons. Values: 'fontawesome', `{'bold': '<b>b</b>', 'italic': '<i>i</i>'}`. Default: `false`
+* __buttonLabels__: type of labels on the buttons. Values: `false` | 'fontawesome'.  Default: `false`
 #### NOTE:
 Using `'fontawesome'` as the buttonLabels requires version 4.1.0 of the fontawesome css to be on the page to ensure all icons will be displayed correctly
 * __delay__: time in milliseconds to show the toolbar or anchor tag preview. Default: `0`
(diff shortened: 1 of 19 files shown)


README.md
```diff
@@ -151,6 +154,66 @@ var editor = new MediumEditor('.editable', {
...
+            'italic',
+            {
+                name: 'h1',
+                action: 'append-h2',
+                aria: 'header type 1',
+                tagNames: ['h2'],
+                contentDefault: '<b>H1</b>',
+                classList: ['custom-class-h1'],
+                attrs: {
+                    'data-header-type': 'h1'
```
j0k3r: Is it only for the example or does it have any action related to this specific attribute?

daviferreira:
:+1:

README.md
nmielnik: It is just an example, but it's a good point, I don't want to be confusing.  I'll make it more obvious that it's an example.

PR #678, opened 2015-06-10
Auto link: fix bug when entering link at end of paragraph

patrick-webs:
Fix the following issue with auto-link:

* Typing in an auto link at the end of the paragraph and then hitting enter over and over would never result in getting to the next paragraph (requiring use of the cursor keys or the mouse), because the selection would be saved and restored each time due to the SPAN[data-auto-link] getting removed from the new paragraph, and the current Medium selection import/export utilities cannot index a cursor into an empty paragraph that contains no text.


Gruntfile.js
```diff
@@ -125,7 +125,7 @@ module.exports = function (grunt) {
                     consolidate: true
                 },
                 keepRunner: true,
-                template: require('grunt-template-jasmine-istanbul'),
+                //template: require('grunt-template-jasmine-istanbul'),
```
j0k3r: ?

j0k3r:
Could you remove dist files from your PR ? Thanks

Gruntfile.js
patrick-webs: Oops, sorry. Didn't think I committed that.

src/js/extensions/auto-link.js
```diff
@@ -125,17 +128,61 @@ LINK_REGEXP_TEXT =
...
+                                break;
+                            }
+                            whiteSpaceText += textContent[j];
+                        }
+                        spans[i].parentNode.insertBefore(this.document.createTextNode(whiteSpaceText), spans[i].nextSibling);
+                        this.removeTrailingCharacters(spans[i], whiteSpaceText.length);
+                    } else {
+                        // Some editing has happened to the span, so just remove it entirely. The user can put it back
+                        // around just the href content if they need to prevent it from linking
+                        while (spans[i].childNodes.length > 0) {
```
nmielnik: You can use `Util.unwrap()` for this I believe

PR #679, opened 2015-06-10
Create composer.json

youri-g:
Its useful to create a composer.json file in order to reference the repository in packagist, so that it can be easily downloadable for PHP projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,19 @@
+{
+    "name": "medium-editor/medium-editor",
+    "description": "Medium.com WYSIWYG editor clone.",
+    "homepage": "https://github.com/yabwe/medium-editor",
+    "repositories": [
+        {
+            "type": "package",
+            "package": {
+                "name": "medium-editor/medium-editor",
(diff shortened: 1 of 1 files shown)


daviferreira:
I don't think that makes much sense, you should either use bower or npm for front-end stuff :)

j0k3r:
Well, I'm against that :-1: 

I know Composer & Packagist, but do you know Bower?
You should use a proper solution to handle your assets instead of relying on Composer/Packagist/Satis for everything. Composer is dedicated to PHP project, Bower is dedicated to JS projects (to make it simple).

You just need 2 files to handle that.

`bower.json`, to retrieve deps:

```json
{
  "name": "My PHP Project",
  "version": "1.0.0",
  "private": "true",
  "dependencies": {
    "medium-editor": "^4.12.0"
  }
}
```

`.bowerrc`, to defined where to write JS deps:

```json
{
  "directory": "web/components/",
  "analytics": false
}
```

Then run `bower up` and you're done.

PR #680, opened 2015-06-10
Use uppercase version of shortcut

j0k3r:
We handle shortcut by listening to onKeyDown. There is also onKeyPress but it is only fired when something is inserted in the content. For example, when you hit the shift key, keyPress is never fired because it does not insert something in the content.

onKeyPress & onKeyDown doesn't handle the key code in the same way.
When you hit `b` letter, onKeyDown will receive 66 but onKeyPress will receive 98.
When you hit `B` letter, onKeyDown will receive 66 and onKeyPress will receive 66.
The way we transcript `b` shortcut to a number is by using: `'b'.charCodeAt(0)` which is 98

As you might guess, it can't work. Because we will receive 66 when we are waiting for 98 to bold the content;

Using uppercase letter fix the problem because `'B'.charCodeAt(0)` is equal to 66.

You can test that: http://www.asquare.net/javascript/tests/KeyCode.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,19 +274,19 @@ var editor = new MediumEditor('.editable', {
         commands: [
             {
                 command: 'bold',
-                key: 'b',
+                key: 'B',
                 meta: true,
                 shift: false
             },
             {
                 command: 'italic',
-                key: 'i',
+                key: 'I',
                 meta: true,
                 shift: false
             },
             {
                 command: 'underline',
-                key: 'u',
+                key: 'U',
                 meta: true,
                 shift: false
             }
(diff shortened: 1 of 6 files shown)


phiggins42:
another thing to note, because we use keydown vs keypress, we are unable to bind to other keys. 

for example, google docs uses ctrl-/ for removeformatting shortcut, but the default hotkey:"/" won't work because of similar issues.

nmielnik:
Yeah, my thoughts on that were we could add another option to the config that accepted a straight keycode instead of a character.

So you could pass either `key` or `keyCode`, and we would update the key handler to check for either.

PR #682, opened 2015-06-10
Ability to disable a shortcut

j0k3r:


spec/keyboard-commands.spec.js
```diff
@@ -110,5 +110,30 @@ describe('KeyboardCommands TestCase', function () {
             });
             expect(editor.execAction).not.toHaveBeenCalled();
         });
+
+        it('should not execute when the command key are false', function () {
```
nmielnik: Can you add another test case (or another assertion in this test case) where if `event.shiftKey` is `true`, it should NOT block the code from executing?

j0k3r: Like that ?

```js
fireEvent(editor.elements[0], 'keydown', {
    keyCode: 'J'.charCodeAt(0),
    ctrlKey: true,
    metaKey: true,
    shiftKey: true
});
expect(editor.execAction).not.toHaveBeenCalled();
```

nmielnik: Actually, you could look at the result of `fireEvent()` and see if `preventDefault()` had been called on it.

That way you could:
```js
var result = fireEvent(... {
    ...
    shiftKey: false
});
expect(result).toBe(true); // true == cancelled

result = fireEvent(... {
    ...
    shiftKey: true
});
expect(result).toBe(false); // false == not cancelled
```

I think that's a better way to check if the functionality does what you expect it to do, what do you think?

j0k3r: It's a good idea :bulb:

PR #683, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


j0k3r:
Can you also update the Gruntfile about `gruntConfig.bump` so it doesn't update the bower.json anymore ? Thanks

PR #688, opened 2015-06-16
Use nodeName instead of tagName + Fix busted code in toolbar

nmielnik:
I propose we begin to use `nodeName` everywhere instead of `tagName`.  `nodeName` is defined on the `Node` class for all nodes, whereas `tagName` only exists on `Element`.  Thus, we can always use `nodeName` safely, whereas `tagName` might not exists.  This simplifies the code, avoids unnecessary exceptions, and since `nodeName` and `tagName` are identical on `Element`s, the behavior of our code should remain the same.

I've also fixed a line of code in toolbar that was broken and always returning true. I've also ensured that when we move up the DOM hierarchy while checking the state of elements, we're not bailing out early if we start at a text node (which I don't believe could have happened anyway).

Lastly, I've reused the new `Util.isMediumEditorElement()` helper everywhere in the code where it was explicitly checking for the `data-medium-editor-element` attribute.

### DIFF
--- a/src/js/extensions/toolbar.js
+++ b/src/js/extensions/toolbar.js
@@ -469,6 +469,13 @@ var Toolbar;
 
             parentNode = Selection.getSelectedParentElement(selectionRange);
 
+            // Make sure the selection parent isn't outside of the contenteditable
+            if (!this.getEditorElements().some(function (element) {
+                    return Util.isDescendant(element, parentNode, true);
+                })) {
+                return;
+            }
+
             // Climb up the DOM and do manual checks for whether a certain extension is currently enabled for this node
             while (parentNode) {
                 manualStateChecks.forEach(updateExtensionState);
--- a/src/js/extensions/toolbar.js
+++ b/src/js/extensions/toolbar.js
@@ -449,8 +449,6 @@ var Toolbar;
                 return;
             }
 
-            parentNode = Selection.getSelectedParentElement(selectionRange);
-
             // Loop through all extensions
             this.forEachExtension(function (extension) {
                 // For those extensions where we can use document.queryCommandState(), do so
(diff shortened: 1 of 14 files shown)


src/js/extensions/toolbar.js
```diff
@@ -469,12 +467,21 @@ var Toolbar;
...
+
+            // Make sure the selection parent isn't outside of the contenteditable
+            if (!this.getEditorElements().some(function (element) {
+                    return Util.isDescendant(element, parentNode, true);
+                })) {
+                return;
+            }
+
             // Climb up the DOM and do manual checks for whether a certain extension is currently enabled for this node
-            while (parentNode.tagName !== undefined && Util.parentElements.indexOf(parentNode.tagName.toLowerCase) === -1) {
```
nmielnik: This is the busted line of code.  Our array of `parentElement`s array of tag names will never contain an instance of the `String.toLowerCase` function (big surprise).

src/js/extensions/toolbar.js
```diff
@@ -469,12 +467,21 @@ var Toolbar;
                 manualStateChecks.push(extension);
             });
 
+            parentNode = Selection.getSelectedParentElement(selectionRange);
+
+            // Make sure the selection parent isn't outside of the contenteditable
+            if (!this.getEditorElements().some(function (element) {
```
nmielnik: The previous loop was verifying that `parentNode` had a tag name, which would prevent the code from passing the element into the extensions if the `parentNode` was either a text node, or `document` (which apparently doesn't have a `tagName` either).  Since I've remove the `tagName` check, make sure the `parentNode` is either an editor element, or a descendant of one, then we don't have to worry about hitting the `document`

src/js/util.js
```diff
@@ -614,7 +614,7 @@ var Util;
         },
 
         isMediumEditorElement: function (element) {
-            return element && element.nodeType !== 3 && !!element.getAttribute('data-medium-editor-element');
+            return element && element.getAttribute && !!element.getAttribute('data-medium-editor-element');
```
j0k3r: Is it equivalent ?

j0k3r:
Good idea :+1:

src/js/util.js
nmielnik: I had to add this line because apparently `.getAttribute` does not exist on `document`, and so this would throw an error if you passed in the document.  Who knew?

daviferreira:
:+1:

PR #697, opened 2015-06-19
Ignore all source files

j0k3r:
Why do people needs `sass` file in Bower ? (see #147)

ping @seifsallam @daviferreira @nchase


daviferreira:
Maybe to use with their own sass, but I agree with you, I think it should not be included on the bower package. :+1:

PR #700, opened 2015-06-22
Creating a link should create only one anchor if possible

patrick-webs:
When selecting text across inline element boundaries, the built-in
browser createLink command will create 2 or more anchor tags. This is a
weird behavior that the user will find strange when they go to delete or
modify one of the anchor tags and find that their changes don't affect
the other adjacent tags that appear to be part of a single link.

This fixes this problem for selections within a single paragraph or
other parent (block-level) element. If the user has selected text across
multiple parent elements, we fall back to the built-in browser behavior
as there is no way to ensure only a single link is created in this
situation.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -923,14 +923,38 @@ function MediumEditor(elements, options) {
                 i;
 
             if (opts.url && opts.url.trim().length > 0) {
-                this.options.ownerDocument.execCommand('createLink', false, opts.url);
+                var currentSelection = this.options.contentWindow.getSelection();
+                if (currentSelection) {
+                    var exportedSelection = this.exportSelection(),
+                        currentEditor,
+                        startContainerParentElement,
+                        endContainerParentElement,
+                        textNodes;
 
-                if (this.options.targetBlank || opts.target === '_blank') {
-                    Util.setTargetBlank(Selection.getSelectionStart(this.options.ownerDocument), opts.url);
-                }
+                    currentEditor = Util.traverseUp(currentSelection.getRangeAt(0).startContainer, function (node) {
+                        return node.getAttribute('data-medium-editor-element');
+                    });
+                    startContainerParentElement = Util.findParentBlockContainer(
+                        currentSelection.getRangeAt(0).startContainer);
+                    endContainerParentElement = Util.findParentBlockContainer(
+                        currentSelection.getRangeAt(0).endContainer);
 
-                if (opts.buttonClass) {
-                    Util.addClassToAnchors(Selection.getSelectionStart(this.options.ownerDocument), opts.buttonClass);
+                    if (startContainerParentElement === endContainerParentElement) {
+                        textNodes = Util.findOrCreateMatchingTextNodes(this.options.ownerDocument,
+                                currentEditor,
+                                exportedSelection);
+                        Util.createLink(this.options.ownerDocument, textNodes, opts.url.trim());
+                        this.restoreSelection();
+                    } else {
+                        this.options.ownerDocument.execCommand('createLink', false, opts.url);
+                    }
+                    if (this.options.targetBlank || opts.target === '_blank') {
+                        Util.setTargetBlank(Selection.getSelectionStart(this.options.ownerDocument), opts.url);
+                    }
+
+                    if (opts.buttonClass) {
+                        Util.addClassToAnchors(Selection.getSelectionStart(this.options.ownerDocument), opts.buttonClass);
+                    }
                 }
             }
 
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -941,8 +941,6 @@ function MediumEditor(elements, options) {
                             fragment = this.options.ownerDocument.createDocumentFragment();
                         exportedSelection = this.exportSelection();
                         fragment.appendChild(parentElement.cloneNode(true));
-                        this.options.contentWindow.getSelection().removeAllRanges();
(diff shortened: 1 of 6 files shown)


j0k3r:
Could you add a description as comment on every method you added in Util? So we know how they are supposed to be used and what they do (even is the function name seems to be explicit).

src/js/util.js
```diff
@@ -218,6 +284,12 @@ var Util;
             return false;
         },
 
+        findParentBlockContainer: function (node) {
```
nmielnik: Could you use `Util.getTopBlockContainer()` instead of this?

src/js/core.js
```diff
@@ -923,14 +923,38 @@ function MediumEditor(elements, options) {
...
+
+                    currentEditor = Util.traverseUp(currentSelection.getRangeAt(0).startContainer, function (node) {
+                        return node.getAttribute('data-medium-editor-element');
+                    });
+                    startContainerParentElement = Util.findParentBlockContainer(
+                        currentSelection.getRangeAt(0).startContainer);
+                    endContainerParentElement = Util.findParentBlockContainer(
+                        currentSelection.getRangeAt(0).endContainer);
+
+                    if (startContainerParentElement === endContainerParentElement) {
```
nmielnik: Wasn't the goal of the code to target the case where the selection spans multiple block elements?  It seems like `startContianerParentElement === endContainerParentElement` would be for the most common case (where the selection is all within 1 paragraph) or have I misunderstood something?

src/js/core.js
```diff
@@ -923,14 +923,38 @@ function MediumEditor(elements, options) {
...
-                }
+                var currentSelection = this.options.contentWindow.getSelection();
+                if (currentSelection) {
+                    var exportedSelection = this.exportSelection(),
+                        currentEditor,
+                        startContainerParentElement,
+                        endContainerParentElement,
+                        textNodes;
+
+                    currentEditor = Util.traverseUp(currentSelection.getRangeAt(0).startContainer, function (node) {
```
nmielnik: `Selection.getSelectionElement()` will do this same thing for you :smile:

src/js/core.js
```diff
@@ -923,14 +923,38 @@ function MediumEditor(elements, options) {
...
 
             if (opts.url && opts.url.trim().length > 0) {
-                this.options.ownerDocument.execCommand('createLink', false, opts.url);
-
-                if (this.options.targetBlank || opts.target === '_blank') {
-                    Util.setTargetBlank(Selection.getSelectionStart(this.options.ownerDocument), opts.url);
-                }
+                var currentSelection = this.options.contentWindow.getSelection();
+                if (currentSelection) {
+                    var exportedSelection = this.exportSelection(),
```
nmielnik: An optimization could be to have both this call to `exportSelection()` as well as the call to get the `currentEditor` happen inside of the `if` statement that actually uses them instead of for all cases.

src/js/core.js
patrick-webs: The goal of this code is to target the most common case (selection within 1 paragraph) and fix the issues that occur when the selection partially overlaps common inline elements such as SPAN, STRONG, and EM.

It isn't possible to fix these issues if the selection spans multiple parent (block) elements as HTML doesn't allow a single anchor tag to begin within one block element and continue into partial content of a subsequent block element, there has to be individual anchor tags inside each block element. E.g. `<p>Hello <a href="foo">world,</p><p>This is a test.</a> But this isn't</p>` is not allowed in HTML. So in that case we just use the browser's built-in `createLink` as there isn't much special we can do to fix stuff.

src/js/util.js
patrick-webs: No, because we want to find the closest block container to the element, not the top container. I'll change the name to `Util.getClosestBlockContainer` to make the difference between these two methods clear.

For instance in the following html:

    <div id=medium-editor>
    <blockquote>
        <pre>Here is some pre-formatted text</pre>
        <p>And here is some non preformatted text</p>
    </blockquote>
    </div>

If the user has selected such that `startContainer` is inside the `pre` tag and `endContainer` inside the `p` tag (e.g. a selection of "pre-formatted text And here is some", we want to know this and fall back to the browser `createLink` command rather than attempting to split up tags, which would result in changing the format of the text (due to duplicating `pre` or `p` tags which have implicit line break semantics that the user won't want us to make).

src/js/core.js
patrick-webs: OK, made this change.

src/js/core.js
patrick-webs: Thanks. done.

patrick-webs:
@j0k3r I added a description to the `Util` methods.

src/js/core.js
```diff
@@ -923,14 +923,75 @@ function MediumEditor(elements, options) {
...
+
+                    if (startContainerParentElement === endContainerParentElement) {
+                        var currentEditor = Selection.getSelectionElement(this.options.contentWindow),
+                            parentElement = (startContainerParentElement || currentEditor),
+                            fragment = this.options.ownerDocument.createDocumentFragment();
+                        exportedSelection = this.exportSelection();
+                        fragment.appendChild(parentElement.cloneNode(true));
+                        this.options.contentWindow.getSelection().removeAllRanges();
+                        var range = this.options.ownerDocument.createRange();
+                        if (currentEditor === parentElement) {
```
patrick-webs: add a helper function that provides the four lines:

# remove all ranges
# create range
# set start
# set end
# add range

PR #703, opened 2015-06-24
Ensure all extra attributes are removed on destroy

### LINKED ISSUE #702: .destroy() question
Hi,

the .destroy() call should remove all extra stuff the editor bind to the editable DOM element?
Because for me the
role="textbox"
aria-multiline="true" 
medium-editor-index="0"

Stays inthe code (I tried on the demo page via console and there I had the same result).

Is it a feature ( because of course I can remove them manually) or some minor bug?

Thx.

John

j0k3r:
Fix #702

### DIFF
--- a/spec/setup.spec.js
+++ b/spec/setup.spec.js
@@ -31,6 +31,10 @@ describe('Setup/Destroy TestCase', function () {
             spyOn(MediumEditor.prototype, 'setup').and.callThrough();
             editor.setup();
             expect(editor.setup).toHaveBeenCalled();
+            expect(document.querySelector('[data-medium-editor-element]')).toBeTruthy();
+            expect(document.querySelector('[aria-multiline]')).toBeTruthy();
+            expect(document.querySelector('[medium-editor-index]')).toBeTruthy();
+            expect(document.querySelector('[role]')).toBeTruthy();
         });
 
         it('should know about defaults', function () {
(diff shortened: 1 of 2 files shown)


src/js/core.js
```diff
@@ -520,9 +520,14 @@ function MediumEditor(elements, options) {
                     element.innerHTML = element.innerHTML;
                 }
 
+                // cleanup extra added attributes
                 element.removeAttribute('contentEditable');
                 element.removeAttribute('spellcheck');
+                element.removeAttribute('data-medium-focused');
                 element.removeAttribute('data-medium-editor-element');
+                element.removeAttribute('role');
```
nmielnik: I would vote for only removing attributes that we knowingly added. I'm not sure why `role` and `aria-multiline` are getting added by some browsers, but if someone purposefully added those attributes before initializing and destroying MediumEditor, I don't think we should remove them.

So yeah, can we reduce this to only remove `data-medium-focused` and `medium-editor-index`?

spec/setup.spec.js
```diff
@@ -31,6 +31,10 @@ describe('Setup/Destroy TestCase', function () {
             spyOn(MediumEditor.prototype, 'setup').and.callThrough();
             editor.setup();
             expect(editor.setup).toHaveBeenCalled();
+            expect(document.querySelector('[data-medium-editor-element]')).toBeTruthy();
+            expect(document.querySelector('[aria-multiline]')).toBeTruthy();
+            expect(document.querySelector('[medium-editor-index]')).toBeTruthy();
+            expect(document.querySelector('[role]')).toBeTruthy();
```
nmielnik: Could we just check for all of the attributes here (including `spellcheck` and `contentEditable`)?

spec/setup.spec.js
```diff
@@ -44,6 +48,12 @@ describe('Setup/Destroy TestCase', function () {
             expect(document.querySelector('.medium-editor-toolbar')).toBeTruthy();
             editor.destroy();
             expect(document.querySelector('.medium-editor-toolbar')).toBeFalsy();
+
+            expect(document.querySelector('[data-medium-editor-element]')).toBeFalsy();
+            expect(document.querySelector('[data-medium-focused]')).toBeFalsy();
+            expect(document.querySelector('[aria-multiline]')).toBeFalsy();
+            expect(document.querySelector('[medium-editor-index]')).toBeFalsy();
+            expect(document.querySelector('[role]')).toBeFalsy();
```
nmielnik: If we did this check here, we could ensure we don't add any other stray attributes in the future:

```js
expect(this.el.getAttribute('class')).toBe('editor');
expect(this.el.attributes.length).toBe(1); // 'class' should be the only remaining attribute
```

spec/setup.spec.js
j0k3r: Done

PR #707, opened 2015-06-27
Some Events Cleanup

nmielnik:
This PR covers cleaning up some code in the Events object.

1. The Events object maintains setting the `data-medium-focused` attribute on the currently focused editor element. Since Events has a destroy method, Events should be reponsibile for removing this attribute, not the core editor.
2. There were many repeat lines where Event looped through each editor element and attached an event.  I've created a helper method to avoid all of this repeat code.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -572,6 +572,8 @@ function MediumEditor(elements, options) {
                 }
             }, this);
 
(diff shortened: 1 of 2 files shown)


j0k3r:
:+1:

PR #708, opened 2015-06-27
Fix allowMultiParagraphSelection option

nmielnik:
This PR covers two issues with the `allowMultiParagraphSelection` option.

1. When determining whether the selection spanned multiple block elements, the regular expression used to remove empty elements would remove multiple valid blocks if it found any empty ones: http://stackoverflow.com/questions/3129738/remove-empty-tags-using-regex

2. The `allowMultiParagraphSelection` option should have been moved into the toolbar options for v5.0 but it was missed.  The option has now been moved into the toolbar options, and the `allowMultiParagraphSelection` option on the global options is deprecated and will be removed in v6.0.0


j0k3r:
Good catch

Issue #666, opened 2015-06-02
Create list inside of list


IonicaBizau:
Creating lists inside of other lists work fine, but the HTML is wrong:

Instead of having:

```html
<ul>
    <li>item 1</li>
    <ul>
        <li>foo</li>
    </ul>
    <li>item 2</li>
    <li>item 3</li>
</ul>
```

The correct option is to have `<li>` element containing the nested `<ul>`:

```html
<ul>
    <li>item 1</li>
    <li>
      <ul>
        <li>foo</li>
      </ul>
    </li>
    <li>item 2</li>
    <li>item 3</li>
</ul>
```

> [![](https://cloud.githubusercontent.com/assets/2864371/7934928/cc65d772-0934-11e5-9486-ad8a8bf5c3ff.png)](http://ionicabizau.github.io/medium-editor-markdown/)

Here are the [W3 specifications](http://www.w3.org/wiki/HTML_lists#Nesting_lists).

> Note how the nested list starts after the <li> and the text of the containing list item (“Chapter One”); then ends before the </li> of the containing list item. 

However:

> Theoretically you can nest as many lists as you like, although in practice it can become confusing to nest lists too deeply

This is important in [`medium-editor-markdown`](https://github.com/IonicaBizau/medium-editor-markdown/) extension since `to-markdown` doesn't handle this way of list nesting. I'm not really sure if @domchristie should fix this in `to-markdown` or you, here in medium-editor. I would choose the second  option. :sparkle: 

/cc https://github.com/IonicaBizau/medium-editor-markdown/issues/6

mphasize:
:+1:

IonicaBizau:
@daviferreira Any feedback on this?

j0k3r:
@IonicaBizau We are focusing on the v5 release at the moment. We are cleaning stuff, removing all deprecated things, etc .. Since it's a major release, we are _allowed_ to break things so we need to be really focus.

After the v5 release, we plan to focus on closing issues and fixing bugs. Since we are keeping a branch for the v4.12.x, fixed issues & bugs will be included in both v4 & v5 (we'll do our best for that).

We plan to release v5 in the coming weeks.
Meanwhile, if you find time to dig into that list problem, we'll be happy to review your PR!

IonicaBizau:
[...]

PR #710, opened 2015-06-29
Fix for #666: Creating and removing nested lists by indenting/outdenting

mphasize:
This PR aims to correctly handle indenting and outdenting lists (i.e. nesting lists). (see #666 for discussion)

`medium-editor` triggers the action through a contenteditable browser `execCommand`, so we would expect the browser to produce valid DOM as a result of that command, but unfortunately popular browsers like Chrome in fact create invalid DOM. So instead of relying on the browser, we need to execute the command and follow up with inspecting the created DOM and fixing it when the structure is not correct. Since there already is a `cleanListDOM` function in the code, we can build on that and include these use cases.

W3 spec for nested lists for reference: http://www.w3.org/wiki/HTML_lists#Nesting_lists

This PR:
- [x] fixes the content test case, using valid DOM structures in the test case
- [x] runs `cleanListDOM` after indenting or outdenting a list
- [x] fix position of the cursor after indenting / outdenting
- [x] the newly created `ul` needs to be moved inside the previous `li` when indenting
- [x] when outdenting, the remaining `li` is stranded inside it's previous parent `li` and needs to be moved to the new position right after the previous parent `li`
- [x] handle the edge case when trying to indent the first list item
- [x] add test for edge case
- [ ] fix outdenting list items with siblings present


mphasize:
This PR is still work in progress, but it would be great to discuss potential implications already.

One thing I noticed is that the correct nesting can create a weird visual artifact with a two bullet points in front of the nested list item.

src/js/util.js
```diff
@@ -844,6 +855,26 @@ var Util;
             } else {
                 el.parentNode.removeChild(el);
             }
+        },
+        wrap: function (el, doc, tag) { // see https://stackoverflow.com/questions/3337587/wrapping-a-set-of-dom-elements-using-javascript
```
j0k3r: Could you add some tests on that method ?

Also, add an extra line before the function and move the comment about SO before the function name.

src/js/util.js
```diff
@@ -476,11 +476,22 @@ var Util;
                 return;
             }
 
-            var list = element.parentElement;
+            var parent = element.parentElement;
 
-            if (list.parentElement.nodeName.toLowerCase() === 'p') { // yes we need to clean up
-                this.unwrap(list.parentElement, ownerDocument);
+            if ((parent.nodeName.toLowerCase() === 'ul' || parent.nodeName.toLowerCase() === 'ol') && (parent.parentElement.nodeName.toLowerCase() === 'ul' || parent.parentElement.nodeName.toLowerCase() === 'ol')) { // yes, we need to fix list nesting (valid is: ul/ol > li > ul/ol )
```
j0k3r: Could you move comments at the end of each `if` just before the `if` instead ? It'll be more easy to read

mphasize:
Fixing this issue reveals some other issues. If you look at the W3 spec mentioned above, the example suggests that you should be able to write text inside a higher level list item and then create a nested list with further items related to the top-level list item. With the current behavior of the toolbar buttons and the keyboard commands, it seems impossible to navigate the cursor into a position to write content for a top-level list item and then create a second order list inside that list item.

- cursor cannot be moved into the space between a higher order list item and the beginning of a lower order list
- clicking a list button inside a list item always removes the top-level list, there seems to be no way to create further indentations

What would be the desired way / approach to create nested lists?

mphasize:
@j0k3r I'll fix the comments right now. The `unwrap` function in utils.js also seems to be missing tests. Is there any template for unit testing this? But apart from a specific test, the behavior of the function is implicitly tested in `content.spec.js` with the tests for indenting/outdenting an ordered list.

j0k3r:
@mphasize Even if the function is implicitly tested, it stills good to have dedicated tests on it just in case we update it in the future.
And yes sadly, `unwrap` function doesn't have tests ..

You already have a playground in `util.spec.js` you can follow the [test of `isListItem`](https://github.com/yabwe/medium-editor/blob/master/spec/util.spec.js#L280) for example.

mphasize:
@j0k3r I'll see what I can do about it later (need to move on now). One more question though: The contributing guideline says not to include anything from the `/dist` folder, but the files in there are obviously tracked in the repo... so do I need to add them?

j0k3r:
@mphasize Yeah it's an issue that we didn't fix yet. Dist files are tracked because npm need them for the release. We don't want them in PR because it often generates conflicts with other PR (because of the minification, etc ..).

And each time we prepare a new release, we have a task in the Gruntfile that handles that (in addition to increment number, etc ..).

if you have a suggestion or a workaround about that with npm, tell us :wink:

IonicaBizau:
> Dist files are tracked because npm need them for the release.

@j0k3r Why not generate them on `npm install` then?

j0k3r:
@IonicaBizau you mean by using postinstall script? In fact, we don't want to pollute npm with all our files.

We recently remove sass files from bower and I tihnk it's better to send only dist files and undist but concat files instead of everything.

But maybe we shouldn't pollute the @mphasize PR and discuss that on a separate issue.

IonicaBizau:
@j0k3r Yep, makes sense! :+1:

mphasize:
So I updated the PR as discussed and the behavior looks good to me.

When in the first item of a list, I'm currently ignoring a users attempt to indent as I'm convinced that a second order list needs a first order list in order to be.
However, ignoring a user keypress is never the best solution and there might be better things to do instead of ignoring, but I feel that's for another issue to discuss and figure out.

Final touch on this PR would be to implement a test for the first list item edge case, but I would love your feedback on this before I put in the effort.

src/js/util.js
```diff
@@ -476,11 +476,30 @@ var Util;
...
 
-            var list = element.parentElement;
+            var parent = element.parentElement;
 
-            if (list.parentElement.nodeName.toLowerCase() === 'p') { // yes we need to clean up
-                this.unwrap(list.parentElement, ownerDocument);
+            // do we need to fix list nesting? (valid is: ul/ol > li > ul/ol )
+            if ((parent.nodeName.toLowerCase() === 'ul' || parent.nodeName.toLowerCase() === 'ol') && (parent.parentElement.nodeName.toLowerCase() === 'ul' || parent.parentElement.nodeName.toLowerCase() === 'ol')) {
+                // does a previous list item exist? (a second order list needs to be inside a previous list item)
+                if (parent.previousElementSibling.nodeName.toLowerCase() === 'li') {
```
nmielnik: This reference to `parent.previousElementSibling` can be null, which causes an error here.

nmielnik:
So I pulled this down and I must say this is pretty promising, it feels like we're really close.

Some things I noticed:

1. No matter what your selection is, when you tab in the list-item, the cursor is moved to the end of the line. Regardless of whether the cursor is are the beginning of the word, middle of the word, or if you've actually selected text, the cursor moves to the end.

1. In firefox, if you tab-indent in a list-item, and then shift-tab-unindent it back out, it gets merged into the previous list item with multiple `<br>` inbetween.  This makes it look like the bullet for that list item has disappeared: (it works fine in IE and Chrome)

screen shot 2015-06-30 at 7 44 15 am

I'll let you know if I find anything else, but I'm excited about this, thank you sir!

nmielnik:
I'm going to make a change now to make `importSelection` and `exportSelection` into utility methods so you can easily preserve the existing selection and restore it, which should address issue 1 that I found,

nmielnik:
@mphasize once #716 is merged in, you can call `Selection.exportSelection()` before you start your DOM manipulation, and `Selection,importSelection()` after you finish, and that should help preserve selection / cursor position.

mphasize:
@nmielnik Thanks for the feedback! The `export/importSelection` looks promising, so I will definitely use that. I need to check Firefox and see what's going on there. The error with the previousSibling being `null` shouldn't occur, as we do not indent when there's no previous `li`... interesting, so let's figure this out. Does this error specifically occur in Firefox? I've seen an interesting comment in the tests in `content.spec.js`, talking about ignoring Firefox in testing because of a similar issue. Probably best to figure out what's happening there and solve it.

src/js/core.js
```diff
@@ -42,8 +42,13 @@ function MediumEditor(elements, options) {
             // If Shift is down, outdent, otherwise indent
             if (event.shiftKey) {
                 this.options.ownerDocument.execCommand('outdent', false, null);
+                Util.cleanListDOM(this.options.ownerDocument, Selection.getSelectionStart(this.options.ownerDocument));
             } else {
-                this.options.ownerDocument.execCommand('indent', false, null);
+                // If it's the first element in the list, ignore this command
+                if (node.previousSibling !== null) {
+                    this.options.ownerDocument.execCommand('indent', false, null);
```
mphasize: I tried using `exportSelection` and `importSelection` here, calling export before the command is triggered and import after the call to `cleanListDOM`, but unfortunately that did not work as hoped. In fact, the browser behavior jumped back to the original uncorrected behavior sending the cursor to the end of the visual line.

mphasize:
Okay, I fixed the possible error when previousElement is `null` and I added a test case for ignoring an indentation attempt in the first item of a list.

@nmielnik I had a look at what Firefox does when outdenting and Firefox seems to destroy the `li` node it's outdenting. That makes it pretty hard to fix the behavior as it's difficult to correctly identify the original content for moving it around. (At least that's what I think at the moment, maybe there's an easy way? )

I also found another interesting bug while testing everything a bit more.
Consider the following list:

- item A
  - subitem 1
  - subitem 2
- item B

When the user has the selection in `subitem 1` and want's to outdent, `subitem 1` ends up below `subitem 2` although it's supposed to stay on top of it. So it looks like I will need to move `subitem 1` to the top-level list and then move `subitem 2` from `item A` to `subitem 1`.

mphasize:
I did some more research on the topic and the side-effects of fixing the browser implementation. What I've found, is pretty sad.

The browser implementation of `indent` / `outdent` actually handles quite a lot of different situations and edge case "correctly", it just produces invalid DOM. Trying to fix the DOM breaks all the goodness and smartness already built into the browser implementation. Worse, since every browser is broken in a different way, you need to clean-up in different ways, making the fix more and more complex.

I'm at the point where I would argue to *ditch the broken browser command* and instead just build a clean implementation from scratch.

Please let me know what you think on the matter.

nmielnik:
This sad news indeed :disappointed: 

I in favor of any system that would work, so if handling `indent` / `outdent` / `orderedlists` / `unorderedlists` requires a custom implementation, I would definitely not be against it.  It seems like it could be a lot of work, and would require a pretty solid amount of testing, but so does every piece of innovative software :smile:

nmielnik:
Closing this PR for now, definitely let us know if there's anything we can do to help, and feel free to open another PR for this anytime.

PR #715, opened 2015-06-30
Fixes #713 adds target='_blank' for auto links if the option is set

### LINKED ISSUE #713: targetBlank = true not working for automatic links (when autoLink = true)
When links are created automatically and the targetBlank is set to true, the automatically created links will still not have any target set. 

Similar to #508

filipkis:


src/js/extensions/auto-link.js
```diff
@@ -205,7 +205,7 @@ LINK_REGEXP_TEXT =
 
         createAutoLink: function (textNodes, href) {
             href = Util.ensureUrlHasProtocol(href);
-            var anchor = Util.createLink(this.document, textNodes, href),
+            var anchor = Util.createLink(this.document, textNodes, href, this.getEditorOption('targetBlank')),
```
j0k3r: I was wondering why autoLink extension use `Util.createLink` instead of `Core.createLink` ? Because, `Core.createLink` already handle `targetBlank` option ([see](https://github.com/yabwe/medium-editor/blob/master/src/js/core.js#L1054-L1056)).

Any idea @patrick-webs ?

nmielnik: The purpose of `Util.createLink` is to build an anchor tag around a set of textnodes, and it will actually extract the text nodes from the DOM, and insert them into the anchor that is returned (presumably to be inserted somewhere by the user).

`Core.createLink` is your generic use case, where you just build an anchor around the user's current selection.

Now that you bring it up, I'm thinking we should rename `Util.createLink` or move it out of `Util` altogether. We can do that in a follow-up to this.

nmielnik: Also, it looks like both the auto-link extension and `Core.createLink` call `Util.createLink`, so I guess it shouldn't be moved out of Util...and maybe it does make sense to be called this...have to think on it.

src/js/util.js
```diff
@@ -106,10 +106,13 @@ var Util;
          * descendants of the same closest block container. If the preconditions are not met, unexpected
          * behavior will result.
          */
-        createLink: function (document, textNodes, href) {
+        createLink: function (document, textNodes, href, blank) {
```
nmielnik: Could you change this argument from `blank` (`boolean`) to `target` (`string`)?  It will be easier to maintain the helper function in the future if you can always pass in the desired target and not just a boolean specifying a specific target.

src/js/extensions/auto-link.js
j0k3r: Yeah, it shouldn't be removed.
I think it could lead to a confusion since both method _seem_ to the same action and they have the same name

j0k3r:
Everything seems ok :+1:

PR #718, opened 2015-07-02
Add Meteor.js support, http://meteor.com

djhi:
Hi,

I'm a Meteor.js dev and volunteer package maintainer for Meteor. I've been using MediumEditor with Meteor and would like to add it to Meteor's ecosystem which will expose MediumEditor to many new devs (Meteor has 20k+ stars on GitHub), likely bringing in new testers and contributors.

This PR will let you directly publish updated versions of the library as they become available. All you have to do is create an account at https://meteor.com/ (click SIGN IN, then Create account). After you've done that, please let me know the name of the account, and I'll add you as a maintainer.

I've already published the current version of the package on Atmosphere (Meteor's package directory). You can see it at https://atmospherejs.com/mediumeditor/mediumeditor. When you release new versions, you can publish them to Atmosphere with `grunt meteor`.

Thanks & best regards,
Gildas

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -300,6 +300,29 @@ module.exports = function (grunt) {
         }
     };
 
+    gruntConfig.exec = {
+        'meteor-init': {
+            command: [
+                // Make sure Meteor is installed, per https://meteor.com/install.
+                // The curl'ed script is safe; takes 2 minutes to read source & check.
+                'type meteor >/dev/null 2>&1 || { curl https://install.meteor.com/ | sh; }',
+                // Meteor expects package.js to be in the root directory
+                // of the checkout, so copy it there temporarily
+                'cp meteor/package.js .'
+                ].join(';')
+        },
+        'meteor-cleanup': {
+            // remove build files and package.js
+            command: 'rm -rf .build.* versions.json package.js'
+        },
+        'meteor-test': {
+            command: 'node_modules/.bin/spacejam --mongo-url mongodb:// test-packages ./'
+        },
+        'meteor-publish': {
+            command: 'meteor publish'
+        }
+    };
+
     grunt.initConfig(gruntConfig);
 
     require('time-grunt')(grunt);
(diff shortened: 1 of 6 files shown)


j0k3r:
Bonjour Gildas!

I wasn't aware that Meteor were using its own package directory instead of using npm like every one else.

I'm not agaisnt adding MediumEditor to _Atmosphere_ but why do we need all these files? I mean, can we have only one command to build every thing Atmosphere needs and then just run `meteor publish` by ourself?

djhi:
Bonjour Jeremy :)

Rather than explaining here the reason of atmosphere, I invite you to read this which explain why they don't use npm: https://github.com/meteor/meteor/pull/516#issuecomment-12919473

I'm afraid those files are needed for meteor packaging but they won't change so they should not be a problem for you in the future (you won't have to maintain them).

nchase:
is there some way to integrate with meteor _without_ including/maintaining the stuff you added here?

I noticed that `less` is one of the most popular packages in the meteor package repository, but it doesn't have any of the files that you added here. How is that done?

djhi:
The less package is maintained by meteor here:
https://github.com/meteor/meteor/tree/devel/packages/less

As I said before, you won't haev to maintain those files as they are not
going to change. If that's still too much troubles, I will just maintain my
fork and try to keep it updated

Le jeu. 2 juil. 2015 à 17:21, Noah Chase <notifications@github.com> a
écrit :

> is there some way to integrate with meteor *without*
> including/maintaining the stuff you added here?
>
> I noticed that less is one of the most popular packages in the meteor
> package repository, but it doesn't have any of the files that you added
> here. How is that done?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/yabwe/medium-editor/pull/718#issuecomment-118066001>.
>
-- 
Gildas Garcia

nmielnik:
After spending the weekend thinking about it, I think it's going to be best that we don't take these changes into our code base.  I appreciate the work you and others are doing here, but I don't think having all of these files on our code base is the right way to go about this in general.

In the future, as the community grows and the process for adding/maintaining packages in atmosphere improves, we will reconsider this.  Until then, I will have to decline these changes.

Apologies, and thanks a ton for supporting medium-editor!

djhi:
Ok now prb, I'm working on a little app to keep my fork synchronized. Thx for the great lib :)

Issue #722, opened 2015-07-03
imageDragging: false option not working on FF


schnittchen:
I was able to drag&drop images into the editable even with imageDragging: false. I checkt both agains the minified and unminified version from the dist/ directory on master.

Here's my config:

```js
var initEditor = function(){
  var options = {
    toolbar: {
      buttons: ['bold', 'italic', 'underline', 'anchor', 'h2', 'h3', 'h4', 'removeFormat']
    },
    imageDragging: false,
    placeholder: { text: 'hier formatierten Text eingeben' }
  }
  new MediumEditor("#editable", options);
};
```

j0k3r:
On which browser are you testing ?

I've setup a jsfiddle to reproduce your issue but it works for me: http://jsfiddle.net/j0k/jhvzzp1c/3/
 - Put your cursor into the first contenteditable, then drag an image, it works
 - Put your cursor into the second contenteditable, then drag an image, it doesn't works (actually it tells the browser to open the image)

Could you provide a jsfiddle to reproduce?

schnittchen:
I can drag an image into both areas of your fiddle.

I'm using Iceweasle 38.0.1.

j0k3r:
Could you try on a different browser ? I've no idea how _bad_ contenteditable is supported by Iceweasel (which I didn't know few seconds ago).

schnittchen:
Same on FF 38.0.5 and 39.0 on windows (Iceweasle was on debian testing).
On IE 11.0.9.6, the tab is replaced with the image (file://...).

j0k3r:
Ok, I've tried on Firefox 38.0.5 on my Mac (which seems to be almost the same as Iceweasel) and I can drag image on both areas too.
I previously tried on Chrome and it worked.

In fact, `imageDragging: false,` doesn't disable drag'n'drop feature, it does not enable our custom extension to handle dragging. Maybe, when imageDragging is disabled we should add some listeners on  `dragover` and `dragleave` event to cancel them..

j0k3r:
This seems to work for me on Firefox:

```js
editor.subscribe('editableDrag', function(event){
    event.preventDefault();
    return false;
});
editor.subscribe('editableDrop', function(event){
    event.preventDefault();
    return false;
});
```

Could you try on Iceweasel too ?

schnittchen:
With that snippet, dragging has no effect both with imageDragging == true and == false.

j0k3r:
[...]

PR #723, opened 2015-07-06
Disable drag-and-drop on MediumEditor (except for images)

nmielnik:
This PR addresses #722 where when the `imageDragging` option explicitly set to false, you can still drag images into the editor, which causes the browser to navigate away and open the file itself.

This change will create a new `fileDragging` extension, which disables drag-and-drop on all editor elements in general.  If `imageDragging` is true (default), it will attempt to read the file and insert the image into the text.  If the `imageDragging` option is `false`, it will just disabled the image drag-and-drop the same way it disables drag-and-drop for all other files.

This change will not break users who are already overriding and passing the `imageDragging` extension, but passing a `imageDragging` extension into the editor will disable the new `fileDragging` extension and any other changes built into that moving forward.


oyeanuj:
@nmielnik As I understand it, this PR prevents non-image files to be dragged and dropped on the editor. To allow for other files to be dragged like video, docs, etc, changing the `allowedTypes` in the `fileDragger`extension is the right way to go?

nmielnik:
@oyeanuj this whole file-dragging extension is really confusing, I wish I had come up with a better solution, but this is what I remember:

* By default, when you drag and drop a file into a browser, it will open a new tab containing just that file.
* To prevent the browser from opening a new tab, the file-dragging extension will attach to all drag & drop events over the editor, and prevent the default action. This will happen to ALL files whenever the file-dragging extension is being used.
* The `allowedTypes` list (member of file-dragging extension) is just a way to allow some code to run for certain file types.  Currently, `images` is the only entry in the `allowedTypes`.  However, if you look [here](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/file-dragging.js#L51), event if you add another file type into the `allowedTypes`, you'll still need to add some code to actually handle inserting the dragged file into the editor...

So, in addition to adding more entries into the `allowedTypes` array, there will also need to be logic added to handle exactly how to insert those files into the editor.

I could see if you wanted to override the fileDragging extension for your specific use case, and I could also see making modifications to the medium-editor source code to handle additional file types in a nice way.  It's up to you.

Apologies again, the `allowedTypes` property doesn't seem all that useful, and this fileDragging extension is hard to follow.

oyeanuj:
Not at all, I totally understand how things evolve. This context is super helpful as I was confused why imageDragger was overwritten by fileDragger while only accepting images. 

My current thinking was your latter suggestion: modifying fileDragger source code to create markup for different file types (video, audio, document) and if it seems to work smoothly, I will definitely try to issue a PR.

Finally, one more question based on your comment above:
>To prevent the browser from opening a new tab, the file-dragging extension will attach to all drag & drop events over the editor, and prevent the default action.

Will this interfere with drag and drop events to reorder paragraphs or files as well?

Again, thank you for the detailed response!

nmielnik:
I'm not sure if it will interfere or not.  The more I think about it, the more I'm less confident about what will happen.  The best thing would be to give it a shot, and if you aren't able to drag and/or drop, you'll know it's probably the file-dragging extension and can try to workaround it or modify how the extension works so that you can accomplish drag-and-drop.

Not a great answer, I know :stuck_out_tongue_winking_eye:

PR #735, opened 2015-07-20
Fix for custom extension not suppressing built in anchor preview

commandline:
We realized in our application that our custom extension was not suppressing the built in anchor preview so we were seeing unstyled, extra and unwanted preview nodes in our application. The unit test should exercise this bug and the couple of lines changes are all that should be needed to fix it, without introducing any breakage for existing users of the built-in.


nmielnik:
:+1:

Issue #405, opened 2015-02-06
Update contents programatically?


DylanPiercey:
Is it possible to update the contents of a medium-editor programatically? (Without finding the element and overriding the html manually).

hellocosmin:
Yes, though you need to do it yourself. You need to deactivate(), set the new properties for whatever options you want to change and activate() back.

See https://github.com/daviferreira/medium-editor/issues/385 for an example with dynamic buttons.

daviferreira:
@DylanPiercey the problem is what to do when you have multiple editors using the same selector. Right now you can use the insertHTML method, but it will update the focused element only.

nmielnik:
It seems like we would just want a helper method that allows you to set the content of the editor's element(s).

Something like:

```js
setContent: function (html, index) {
    this.elements[index].innerHTML = html;
    // We probably want to default index to either 0, or to loop through all elements and set the content for all elements
}
```

nmielnik:
As pointed out in #731 we would want this code to also trigger the `editableInput` event so any listeners would know about the change to the content.

nmielnik:
Fixed by #736

gsouf:
Hi, is there an alternative to setContent that does not trigger ``editableInput`` event?

PR #736, opened 2015-07-21
Expose setContent method on medium

ayobami14:
Adds a `setContent` method to set the `innerHTML` of an element at a provided index.
https://github.com/yabwe/medium-editor/issues/405

Return early if `contentCache` not defined in `updateInput`


nmielnik:
:+1:

PR #746, opened 2015-07-27
Cleanup only pasted elements

j0k3r:
When you use `cleanPaste` with a multiline pasted text it will do some custom stuff related to paragraph.

But it will also cleanup *block elements*. The problem is that it cleans block element of the WHOLE content (if you do not have a selection, the selected element will be the editor), not only the pasted one. So I moved the code that performs this cleanup inside `pasteHTML` where I have a node so I can cleanup ONLY the pasted content.

Problem appear when using external library that update MediumEditor content adding custom html, like medium-editor-insert-plugin. It adds div, figure, etc .. so if the pasted content was multiline, the cleanup will remove all this tags (according to my `cleanTag` attributes).

Also, we previously added a new line after pasted content. Why ? I think when user paste content it doesn't want a new paragraph to be created right after the pasted content.

*PS: don't take a closer look at the branch name, I was fixing the wrong issue at first :)*

### DIFF
--- a/src/js/extensions/paste.js
+++ b/src/js/extensions/paste.js
@@ -1,4 +1,4 @@
-/*global Util, Selection, Extension */
+/*global Util, Extension */
 var PasteHandler;
 
 (function () {
@@ -138,8 +138,7 @@ var PasteHandler;
         },
 
         cleanPaste: function (text) {
-            var i, elList, workEl,
-                el = Selection.getSelectionElement(this.window),
+            var i, elList,
                 multiline = /<p|<br|<div/.test(text),
                 replacements = createReplacements().concat(this.cleanReplacements || []);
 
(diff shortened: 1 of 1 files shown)


src/js/extensions/paste.js
```diff
@@ -155,31 +154,6 @@ var PasteHandler;
             elList = text.split('<br><br>');
 
             this.pasteHTML('<p>' + elList.join('</p><p>') + '</p>');
-
-            try {
-                this.document.execCommand('insertText', false, '\n');
```
nmielnik: The only thing I can think of why we might do this:

1. Something with undo/redo maybe?  Did you check to see how undo/redo worked with pasting before/after you removed this line?
1. If you paste at the end of chunk of text, does something weird happen?  Maybe the extra newline gave users a place to put their cursor so they can add more text without inheriting the formatting from the pasted text?

Issue #738, opened 2015-07-22
Sometimes adding class to links , doesn't work


l2aelba:
Demo : http://jsfiddle.net/8jrod5c9/

![a3aebhtxxh](https://cloud.githubusercontent.com/assets/983458/8824755/0807eb6a-3079-11e5-924d-fd009cf690cf.gif)
Found bugs :
- Button checkbox still checked
- Sometimes adding class does not work (work perfect just first time)
- You can see more on GIF

nmielnik:
Yeah, there are a bunch of things going wrong here.  I've checked in a fix for the css so the toolbar doesn't look so horrific with checkboxes in it.

Figuring out why things aren't working properly, that will take some investigation.

nmielnik:
@l2aelba I've updated your jsfiddle to include the latest version of MediumEditor: http://jsfiddle.net/8jrod5c9/3/ a long with a few minor tweaks.

1. The issue where the checkboxes are showing up looking terrible in the toolbar has been fixed as of https://github.com/yabwe/medium-editor/commit/d40c6ed9bdb55df9b1ca9e9344d95008f1d121f0 and was released in version 5.5.1
1. I've opened #747 to make sure the checkboxes in the toolbar checked/unchecked correctly based on the anchor you're editing

I also discovered that there's a weird side-effect caused by your `.button` class having a `padding: 10px`.  Because of this padding, if you create a button link, and then un-link it, you are left with a single character of black-background text directly before the link.  Removing the padding from that button makes the issue go away, I removed that padding in the updated jsfiddle link.

Regarding the other problem you ran into in your repo gif, where you're unable to create a 2nd link.  It seems like it's limited to just when you:
1. Create a button link
2. Hit <kbd>ENTER</kbd> a few times to insert some line breaks
3. Start typing text directly after the line breaks, and attempt to create a link with that text.

If you insert any other character (other than new lines) between the previous link and the new link you're trying to create, the issue doesn't happen and the new link is created correctly.

I believe the issue may be related to when ONLY line-breaks separate a new link you're trying to create and a previously created link.  Let me know if you're able to reproduce it another way.

Hopefully now this scopes down the severity of the issue a bit.

l2aelba:
@nmielnik Thanks , i tested https://github.com/yabwe/medium-editor/pull/747

Checkboxes (checked, unchecked) are work.

Now just about addling 2nd link (class/link) left :D

nmielnik:
[...]

PR #752, opened 2015-07-29
Fix #751 - issue with finding anchor when clicking anchor button

### LINKED ISSUE #751: Editor refuse to create links if there is already a link positioned in the very beginning.
Hello!

The bug can be reproduced here with latest Chrome : https://yabwe.github.io/medium-editor/
Create a link on "My father" and then try to create another one after. It fails and refuse to create new links.

screen shot 2015-07-29 at 10 42 27

The script think that there is already a link so refuse to create a new one. 
Here is the part of the code that seems to be wrong :

```js
            var selectedParentElement = Selection.getSelectedParentElement(Selection.getSelectionRange(this.document)),
                firstTextNode = Util.getFirstTextNode(selectedParentElement);

            if (Util.getClosestTag(firstTextNode, 'a')) {
                return this.execAction('unlink');
            }
```

nmielnik:
This fixes #751 

When the user clicks the 'anchor' button, we attempt to find an existing anchor within the selection.  If there is one, we 'unlink' that anchor, otherwise we display the anchor form an allow them to create a link.

There was a bug in the code that tries to find the existing link.  It was incorrectly looking for the first text node within the parent selection node, which manifested itself as being unable to create a link when a link appeared as a previousSibling of a text node which was being selected.

### DIFF
--- a/src/js/extensions/anchor.js
+++ b/src/js/extensions/anchor.js
@@ -60,17 +60,11 @@ var AnchorForm;
             event.preventDefault();
             event.stopPropagation();
 
-            var linkSelected = false,
-                range = Selection.getSelectionRange(this.document);
+            var range = Selection.getSelectionRange(this.document);
 
             if (range.startContainer.nodeName.toLowerCase() === 'a' ||
-                range.endContainer.nodeName.toLowerCase() === 'a') {
-                linkSelected = true;
-            } else if (Util.getClosestTag(Selection.getSelectedParentElement(range), 'a')) {
-                linkSelected = true;
-            }
-
-            if (linkSelected) {
+                range.endContainer.nodeName.toLowerCase() === 'a' ||
+                Util.getClosestTag(Selection.getSelectedParentElement(range), 'a')) {
                 return this.execAction('unlink');
             }
 
--- a/src/js/extensions/anchor.js
+++ b/src/js/extensions/anchor.js
@@ -60,10 +60,9 @@ var AnchorForm;
             event.preventDefault();
             event.stopPropagation();
 
-            var selectedParentElement = Selection.getSelectedParentElement(Selection.getSelectionRange(this.document)),
-                firstTextNode = Util.getFirstTextNode(selectedParentElement);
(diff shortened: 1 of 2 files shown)


nmielnik:
I believe this also fixes issue #738

src/js/extensions/anchor.js
```diff
@@ -60,10 +60,9 @@ var AnchorForm;
             event.preventDefault();
             event.stopPropagation();
 
-            var selectedParentElement = Selection.getSelectedParentElement(Selection.getSelectionRange(this.document)),
-                firstTextNode = Util.getFirstTextNode(selectedParentElement);
+            var selectedParentElement = Selection.getSelectedParentElement(Selection.getSelectionRange(this.document));
```
patrick-webs: this should also handle the case where the start container of the selection is the anchor tag.

src/js/extensions/anchor.js
```diff
@@ -60,10 +60,17 @@ var AnchorForm;
...
-                firstTextNode = Util.getFirstTextNode(selectedParentElement);
+            var linkSelected = false,
+                range = Selection.getSelectionRange(this.document);
+
+            if (range.startContainer.nodeName.toLowerCase() === 'a' ||
+                range.endContainer.nodeName.toLowerCase() === 'a') {
+                linkSelected = true;
+            } else if (Util.getClosestTag(Selection.getSelectedParentElement(range), 'a')) {
+                linkSelected = true;
+            }
```
j0k3r: Why not a single `if` ?

```js
if (range.startContainer.nodeName.toLowerCase() === 'a' ||
    range.endContainer.nodeName.toLowerCase() === 'a' ||
    Util.getClosestTag(Selection.getSelectedParentElement(range), 'a')) {
    return this.execAction('unlink');
}

nmielnik: I was trying to make things more readable since these are 2 different kinds of checks, but I agree it'll be easier to just make it one if statement

j0k3r:
Maybe you can add an extra comment to explain the "combination" conditions then


--
Jeremy

PR #753, opened 2015-07-29
Fix for #738 issue with importSelection after empty paragraphs

### LINKED ISSUE #738: Sometimes adding class to links , doesn't work

nmielnik:
There is logic which handles the case where we import/export the current selection, but the selection occurs after 1 or more empty paragraphs.  We determine where to position the cursor by tracking how many empty paragraphs there are between the last piece of text, and where the selection begins.

However, this logic was assuming that in the case where we need to import the selection, the selection is always collapsed (just a cursor).  Because of this, we were unable to import/export selections correctly at the beginning of paragraphs.  

This also manifested itself such that we were unable to create links at the beginning of paragraphs (ie #738).

This PR removes the code which collapsed the selection when importing selections that occur after empty paragraphs.

### DIFF
--- a/spec/selection.spec.js
+++ b/spec/selection.spec.js
@@ -68,6 +68,24 @@ describe('Selection TestCase', function () {
             expect(node.getAttribute('id')).toBe('1');
         });
 
+        it('should import an exported non-collapsed selection after an empty paragraph', function () {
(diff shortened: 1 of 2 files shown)


j0k3r:
Interesting !

Issue #754, opened 2015-07-31
Replacing already anchored content's link with a new one creates new element inside of first anchor


guneycob:
First of all, I'm using medium-editor v5.5.1. When you have an already anchored text in your editable content and you want to change its link via clicking on the tooltip that showing the current link, the editor never deletes or replaces the old anchor but adds new anchor inside of that.

This causes really serious problems when serializing the content or getting the updated content with some other way.

How is this working?
http://i.imgur.com/uTmSMm9.gif

How this should be working?
http://i.imgur.com/3sVbAAA.gif

PR #755, opened 2015-07-31
Fix for #754, issue with replacing the anchor that already has one

guneycob:
If we unlink the anchor before adding a new one, it works as intended.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -906,6 +906,7 @@ function MediumEditor(elements, options) {
                 i;
 
             if (opts.url && opts.url.trim().length > 0) {
+	            this.execAction('unlink');
                 var currentSelection = this.options.contentWindow.getSelection();
                 if (currentSelection) {
                     var exportedSelection,
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -906,7 +906,7 @@ function MediumEditor(elements, options) {
                 i;
 
             if (opts.url && opts.url.trim().length > 0) {
-	            this.execAction('unlink');
+                this.execAction('unlink');
                 var currentSelection = this.options.contentWindow.getSelection();
                 if (currentSelection) {
                     var exportedSelection,


j0k3r:
Good catch, could you add a test for that case ? Thanks

guneycob:
I've never used Jasmine or something like that before. I've tried something like that in anchor-preview.spec.js based your written test codes:

```javascript
it('should remove the old link and add new instead of that', function () {
		    var editor = this.newMediumEditor('.editor'),
			    anchorPreview = editor.getExtensionByName('anchor-preview'),
			    anchor = editor.getExtensionByName('anchor'),
			    toolbar = editor.getExtensionByName('toolbar'),
                input;

		    // show preview
		    fireEvent(document.getElementById('test-link'), 'mouseover');

		    // load into editor
		    jasmine.clock().tick(1);
		    fireEvent(anchorPreview.getPreviewElement(), 'click');
		    jasmine.clock().tick(200);

            anchor.getInput('http://test.com').value = 'newvalue';
            fireEvent(anchor.getForm('http://test.com').querySelector('a.medium-editor-toolbar-save'), 'click');

            expect(this.el.innerHTML.indexOf('<a href="newvalue">ipsum</a>')).toBe(6);
	    });

It passes but with a removed the id tag of initial 'test-link', and it passes even if the line in my commit is commented. There are something wrong with these test code of mine.

I guess it's better if you do. I'm sure that it will be verified by Jasmine too.

j0k3r:
@guneycob ok no problem, I'll take care of the test then.

j0k3r:
After checking the problem, it seems that this kind of *non-*unlink previous link appears only in Chrome. That's why you failed to reproduce it using Jasmine.

I won't add a test about that, but a comment described why we are unlink things.

And by the way, this bug was introduced in #700 since the whole link creating was rebuild instead of using `this.options.ownerDocument.execCommand('createLink', false, opts.url);`.

j0k3r:
I've merged your stuff manually.

Thanks !

PR #756, opened 2015-08-02
Recreate all attributes when using textarea

### LINKED ISSUE #711: Textarea attribute not carried over to Div
When the `textarea` is hidden and `div` element is created, we are losing a custom attribute that styles the element.  Is there anyway we can attach custom attributes to the `div`?

screen shot 2015-06-28 at 8 58 41 pm

j0k3r:
This should fix #711

All attributes from the textarea are now re-created when the medium-editor is created, except attributes that are already attached to the div. Should I remove this condition?

Should we enable that using an option or by default it's fine?

### DIFF
--- a/spec/textarea.spec.js
+++ b/spec/textarea.spec.js
@@ -15,6 +15,7 @@ describe('Textarea TestCase', function () {
         this.el.setAttribute('data-disable-double-return', false);
         this.el.setAttribute('data-disable-preview', false);
         this.el.setAttribute('spellcheck', true);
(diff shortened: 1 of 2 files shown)


j0k3r:
It seems that Saucelabs & Coveralls are drunk at the moment.

nmielnik:
:+1: 

Also, the tests are passing in saucelabs, it just seems like the reporting back to Travis isn't working right.

Coveralls is all screwed up too, but your changes are fine.

PR #758, opened 2015-08-03
Fix #757, issue where selection restored too large

### LINKED ISSUE #757: Buggy restore selection when creating link on first word in paragraph
Steps to reproduce: 

1. Create a new paragraph
2. Select first word in paragraph
3. Create a link
4. this.base.restoreSelection() doesn't restore selection on link correctly (what is visually seen). MediumEditor.selection.getSelectionStart(this.base.options.ownerDocument) returns previous paragraph, not anchor node from current paragraph. It is problematic when you want to apply custom class on link or, in my case, add custom data attributes to link.

mediumeditor___demo

patrick-webs:
Fix issue #757 where the selection import would ignore the embedded
information about empty paragraphs before it and would import a
selection as multiline that should have been contained at the beginning
of a paragraph.

Also restore a `range.collapse(true)` that was used for an MSIE-specific
fix unrelated to restoring these types of selections.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -262,6 +262,40 @@ describe('Anchor Button TestCase', function () {
             expect(link.classList.contains('btn')).toBe(true);
             expect(link.classList.contains('btn-default')).toBe(true);
         });
+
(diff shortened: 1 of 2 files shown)


nmielnik:
Gracias!

PR #760, opened 2015-08-05
Cleanup comment created by Chrome

j0k3r:
Chrome add comments when you paste content with html enabled.

It's not really a problem if you paste your content once. But if you past it twice, everything after the pasted content will be removed. I guess Chrome find its _fragment_ comments again and erase them ...

Here is a more revelant gif of the problem:

medium-photo-firefox

Should I add some tests specific to Chrome?

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -324,11 +324,6 @@ var Util;
         insertHTMLCommand: function (doc, html) {
             var selection, range, el, fragment, node, lastNode, toReplace;
 
-            // cleanup fragment comment created by Chrome
-            html = html
-                .replace('<!--EndFragment-->', '')
-                .replace('<!--StartFragment-->', '');
-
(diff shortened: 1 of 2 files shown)


src/js/util.js
```diff
@@ -324,6 +324,11 @@ var Util;
         insertHTMLCommand: function (doc, html) {
             var selection, range, el, fragment, node, lastNode, toReplace;
 
+            // cleanup fragment comment created by Chrome
+            html = html
```
nmielnik: Should this be in the paste code?  We have a set of replacements RegExp in there it could be added to, just curious why you added it here.

j0k3r: Yeah, it should be.
I was a bit in a hurry when debugging this stuff I missed the replacement options part.

PR #761, opened 2015-08-07
Fix #757 Track beginning of paragraph in exportSelection

### LINKED ISSUE #757: Buggy restore selection when creating link on first word in paragraph

nmielnik:
This fixes an issue when saving selection if the cursor is at beginning of a paragraph, but not preceded by empty paragraphs.

We track whether the cursor is at the beginning of a paragraph that has empty paragraphs before it by counting the empty paragraph and storing in the selection data.  However, if the paragraph did not have empty paragraphs in front of it, we didn't store anything, which made the selection get set to the end of the paragraph before it, rather the beginning of the paragraph it should be within.

The fix is to always track if the cursor is at the beginning of a paragraph, and set `emptyBlocksIndex` to `0` if there are no empty paragraphs before it.  Before, this case would have just left `emptyBlocksIndex` as undefined.

This manifested itself in the re-opened issue for #757 and should now be fixed.

### DIFF
--- a/src/js/selection.js
+++ b/src/js/selection.js
@@ -180,9 +180,11 @@ var Selection;
             return range;
         },
 
-        // Returns 0 unless the cursor is within or preceded by empty paragraphs/blocks,
-        // in which case it returns the count of such preceding paragraphs, including
-        // the empty paragraph in which the cursor itself may be embedded.
+        // Returns -1 unless the cursor is at the beginning of a paragraph/block
+        // If the paragraph/block is preceeded by empty paragraphs/block (with no text)
+        // it will return the number of empty paragraphs before the cursor.
+        // Otherwise, it will return 0, which indicates the cursor is at the beginning
+        // of a paragraph/block, and not at the end of the paragraph/block before it
         getIndexRelativeToAdjacentEmptyBlocks: function (doc, root, cursorContainer, cursorOffset) {
             // If there is text in front of the cursor, that means there isn't only empty blocks before it
             if (cursorContainer.nodeType === 3 && cursorOffset > 0) {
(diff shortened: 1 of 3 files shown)


src/js/selection.js
```diff
@@ -180,23 +180,24 @@ var Selection;
...
+        // If the paragraph/block is preceeded by empty paragraphs/block (with no text)
+        // it will return the number of empty paragraphs before the cursor.
+        // Otherwise, it will return 0, which indicates the cursor is at the beginning
+        // of a paragraph/block, and not at the end of the paragraph/block before it
         getIndexRelativeToAdjacentEmptyBlocks: function (doc, root, cursorContainer, cursorOffset) {
             // If there is text in front of the cursor, that means there isn't only empty blocks before it
-            if (cursorContainer.nodeType === 3 && cursorOffset > 0) {
-                return 0;
+            if (cursorContainer.textContent.length > 0 && cursorOffset > 0) {
+                return -1;
```
patrick-webs: this will change the behavior if the cursor was after the IMG in the following case. Whether this is important I don't know, since in either case the old code also would restore the cursor the wrong place, just to a place that is slightly less wrong:

    <p>Hello world</p>
    <p></br></p>
    <p>The quick brown fox</p>

nmielnik: So this is not actually a change in behavior, the cursor is restored to the same location as the previous code did.

PR #763, opened 2015-08-07
Fire editableInput only once, at the end of link creation

patrick-webs:
I found that editableInput could be fired twice in the situation where you create a link with a target=_blank attribute. It fires once when the document execCommand `insertHTML` is called, and a second time after the link is modified.

This also had the result of editableInput being fired before medium restored the selection after calling `insertHTML`, which put the selection out of sync with what it should have been.

This PR changes the createLink command to suppress firing of the `editableInput` event until after the HTML is inserted and the selection is restored, plus after any changes to the link attributes have been made in the DOM.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -263,6 +263,49 @@ describe('Anchor Button TestCase', function () {
             expect(link.classList.contains('btn-default')).toBe(true);
         });
 
+        it('should fire editableInput only once when the user creates a link open to a new window,' +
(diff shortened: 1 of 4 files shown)


nmielnik:
If anyone wants to look at this, I would definitely recommend using this link instead, github did a terrible job dealing with the indentation changes:

https://github.com/yabwe/medium-editor/pull/763/files?w=1

j0k3r:
@nmielnik I wasn't aware of this parameter `?w=1`, thanks !

PR #769, opened 2015-08-10
Placeholder text should only show when empty

geofflane:
Integrating this into my app (using Angular ngModel), and I found the placeholder text would show up when I rendered the content of the ngModel into the body. Using the psuedo-class :empty ensures that the placeholder only shows if there is nothing in the element.

### DIFF
--- a/src/sass/components/_placeholder.scss
+++ b/src/sass/components/_placeholder.scss
@@ -1,7 +1,7 @@
 .medium-editor-placeholder {
     position: relative;
 
-    &:after {
+    &:empty:after {
         content: attr(data-placeholder) !important;
         font-style: italic;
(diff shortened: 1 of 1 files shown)


nmielnik:
Hey @geofflane thanks for the pull request.  I wasn't quite familiar with this selector, you learn something new every day.

Unfortunately this doesn't seem to work in the general case.  the `:empty` psuedo-class only works if the element has no children, but if you delete all of the text out of a `contenteditable` there will still be a `<p><br/></p>` left behind.  All of our supported browsers put some version of this html into the empty `contenteditable`, so the `contenteditable` element will never be empty except at the very beginning when the element is created.

PR #775, opened 2015-08-12
Enable Saucelabs tests for Windows 10

nmielnik:
Enable saucelabs tests to run for:
* Latest Chrome on Windows 10
* Latest Firefox on Windows 10
* IE11 on Windows 10

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -49,29 +49,29 @@ module.exports = function (grunt) {
             version: '11',
             platform: 'WIN8.1'
         }, {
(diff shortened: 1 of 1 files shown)


j0k3r:
Wow, this mean medium-editor works fine on Windows 10, without checking Edge :)

nmielnik:
Yup, it was pretty sweet that everything passed when I tried it out.

There are still lots of failures in Edge, but we can't enable Edge in SauceLabs yet.

PR #780, opened 2015-08-14
Fix #776 Remove Global Variables and References

### LINKED ISSUE #773: MediumEditor global variable
_From @gerherz on July 27, 2015 21:57_

Since v0.4.0, MediumEditor is needed as a global variable. I'm instanciating the editor inside another object, making the variable local to that object and unavailable globally.
Shouldn't MediumEditor be imported from its peer dependency so as not to depend from global variables?

```javascript
import MediumEditor from 'medium-editor';
// OR
var MediumEditor = require('medium-editor');
```

_Copied from original issue: yabwe/medium-editor-tables#14_

### LINKED ISSUE #776: Remove references to global variables from extensions
Currently our built-in extensions reference global vars that only internal parts of MediumEditor have access to (ie `Util` or `Selection`).

We can explore dependency injection approaches later, but for now we should at least update all of these references to refer to `MediumEditor.util`, `MediumEditor.selection`, etc.  This will allow the extensions to be extended upon without issue of bad references, which a few people have complained about (including in #773 )

nmielnik:
This fixes #776, and may be sufficient enough to resolve #773 

Move all global variables into members of the `MediumEditor` variable and update their references throughout the code. This includes:
* Utilities
  * `Util`, `Selection`
* Extension
  * `extensionDefaults`, `Extension`, `Button`, `FormExtension`, `Toolbar`, and all the other built-in extensions
* Other Globals
  * `Events`, `editorDefaults`, `buttonDefaults`

In addition, there is now a new `globals.js` file which defines `MediumEditor` and `MediumEditor.extensions` as the first file included in the bundle, which makes these available to all other source code.

Lastly, `MediumEditor`, `setupTestHelpers`, and all jasmine global methods are added as exclusions in `.jshintrc` so references to these do not require a `/*global */` comment at the beginning of the file.


src/js/selection.js
```diff
@@ -1,18 +1,15 @@
...
-        if (Util.isBlockContainer(node)) {
+        if (MediumEditor.util.isBlockContainer(node)) {
             return NodeFilter.FILTER_ACCEPT;
         } else {
             return NodeFilter.FILTER_SKIP;
         }
     }
 
-    Selection = {
+    var Selection = {
```
j0k3r: Why don't you define directly `MediumEditor.selection = {` ?

src/js/util.js
```diff
@@ -50,7 +48,7 @@ var Util;
          * See #591
          */
         isMetaCtrlKey: function (event) {
-            if ((this.isMac && event.metaKey) || (!this.isMac && event.ctrlKey)) {
+            if ((Util.isMac && event.metaKey) || (!Util.isMac && event.ctrlKey)) {
```
j0k3r: `Util` isn't yet created so you should keep `this`, nope? Like you did in `Selection`.

j0k3r:
Everything looks great and more understandable. Also I was wondering if it will require a new major version or not ...

src/js/selection.js
nmielnik: The thought here was to keep the references within the file short (`Selection` instead of `MediumEditor.selection`) in both `Selection` and `Util`.  Just define a local variable for references within the file, and at the end write it to the global variable.  Kind of like how you do things with `define()` or `exports`.

Do you think there's more value in using the global reference?

src/js/util.js
nmielnik: Since this is within a function call, the code won't execute until some other code actually invokes the function, so it's safe to refer to `Util` here.

nmielnik:
I think this just should be a patch.  We're not changing any part of the API since `MediumEditor.util`, `MediumEditor.selection`, `MediumEditor.Extension`, and `MediumEditor.extensions` are still the same.  No external code could actually access any of these global vars we were declaring within our scope, so it's definitely safe.

src/js/selection.js
j0k3r: In fact it was to keep consistency with other files.
For example, you defined:

```js
MediumEditor.extensions.paste = MediumEditor.Extension.extend({
```

Why don't you do?

```js
var PasteHandler = MediumEditor.Extension.extend({
    // ...
});

MediumEditor.extensions.paste = PasteHandler;
```

src/js/util.js
j0k3r: Ok allright :+1:

src/js/selection.js
nmielnik: Ok yeah, that's a fair point and it's a pretty standard pattern. I'll make that change.

PR #782, opened 2015-08-16
Multiple medium instances cannot be used in a single page with textareas

alexmcmanus:
When medium replaces textareas with divs, the medium-editor-textarea-id attributes used to link them together is not unique across the whole page. If you need to use multiple editor instances (for example, if you have a dynamic page, or if your are binding the editor inside some other component framework), the syncing between textarea and editor doesn't work if there are multiple editors.

This change simply re-uses the existing unique div id to make the medium-editor-textarea-id value unique in the page.

### DIFF
--- a/spec/textarea.spec.js
+++ b/spec/textarea.spec.js
@@ -77,6 +77,31 @@ describe('Textarea TestCase', function () {
         });
     });
 
+    it('should create unique medium-editor-textarea-ids across all editor instances', function () {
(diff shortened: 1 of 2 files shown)


nmielnik:
Good catch, thanks for the fix!

j0k3r:
Yeah thanks for the fix and to provide a test to reproduce the bug :+1:

PR #784, opened 2015-08-18
Fix #748 - Don't delete editor inside util.insertHTMLCommand

### LINKED ISSUE #748: Pasting into Medium-Editor in IE removes the editor element from the DOM and replaces it with a text node
If I open the Medium-Editor demo here: https://yabwe.github.io/medium-editor/ in IE11 and clear it's contents before pasting in text, the Medium-Editor itself is removed from the DOM and no longer available.

For example, say this is the HTML for my editor:
```html
<h1 class="title">My article title</h1>
<div class=".editable .my-article">
    <p>Here is my article text...</p>
</div>
```

and Medium-Editor is initialized:
```js
var myEditor = MediumEditor('.editable');
```

This causes my HTML above to be slightly modified like this:
```html
<h1 class="title">My article title</h1>
<div class="editable" contenteditable="true" spellcheck="true" data-medium-editor-element="true" role="textbox" aria-multiline="true" medium-editor-index="0" data-placeholder="Type your text" data-medium-focused="true">
    <p>Here is my article text...</p>
</div>
```

However, if I now select all the text in my editor instance and delete it, and then paste in text from the clipboard (copied from Word, Notepad, wherever). This is what happens to the DOM:
```html
<h1 class="title">My article title</h1>
<p>This is my pasted text...</p>
```

You'll notice that the entire `.editable` div that `myEditor` was bound to is now gone from the DOM. Highlighting text no longer causes the toolbar to appear, the editor instance is completely wiped out.

As I mentioned, this is reproducible on IE11 using the Medium-Editor public demo at https://yabwe.github.io/medium-editor/.  I also tried with the latest version 5.5.1 (the demo is at 5.1.0) and the same issue was still present.

nmielnik:
This fixes #748 which was a bug within the custom `insertHTMLCommand()` utility method which runs on browsers which don't support `insertHTML` natively through `document.execCommand()`.

The custom code for `insertHTMLCommand()` assumed that the editor would always have some content.  In the case where the editor is completely empty (no child nodes), the code will now create a temporary empty textnode which will get replaced by the incoming HTML.

### DIFF
--- a/spec/util.spec.js
+++ b/spec/util.spec.js
@@ -378,4 +378,31 @@ describe('MediumEditor.util', function () {
             MediumEditor.util.isIE = origIsIE;
         });
     });
+
+    describe('insertHTMLCommand', function () {
+        it('should not remove the contenteditable element when calling insert into an empty contenteditable element', function () {
+            var el = this.createElement('div', 'editable', ''),
+                origQCS = document.queryCommandSupported;
+            // Force our custom implementation to run
+            spyOn(document, 'queryCommandSupported').and.callFake(function (command) {
+                if (command === 'insertHTML') {
+                    return false;
+                }
+                return origQCS.apply(document, arguments);
+            });
+            // Mimic an editor element
+            el.setAttribute('contenteditable', true);
+            el.setAttribute('data-medium-editor-element', true);
+
+            // Make sure the element has 0 child nodes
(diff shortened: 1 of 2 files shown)


j0k3r:
Good catch and good job :+1:

PR #785, opened 2015-08-18
Fix auto-link within list items

nmielnik:
This fixes an issue where auto-link was not finding valid urls within list items.

Auto-Link was iterating over `<p>` tags and finding text within them which could be a link.  By instead iterating over all possible block elements, it was able to find valid urls within any of our defined block elements, including list items.


j0k3r:
Good idea !

PR #786, opened 2015-08-19
Handle backward compatibility with <= 5.4.0

j0k3r:
`alt` options was added in 5.4.0 (see #727).

Every custom keyboardCommands config since this version will get a `data.alt` as `undefined` when testing if <kbd>alt</kbd> is pressed. Config example before 5.4.0:

```js
{
  command: 'bold',
  key: 'B',
  meta: true,
  shift: false
},
```

And then `undefined === false` is falsy so the command will never be fired.

I don't know if we need this extra `if` or an extra check in the following `if`.
If you guys agree about that solution, I'll add some tests after.

### DIFF
--- a/spec/keyboard-commands.spec.js
+++ b/spec/keyboard-commands.spec.js
@@ -93,18 +93,35 @@ describe('KeyboardCommands TestCase', function () {
                             key: 'p',
                             meta: true,
                             shift: false
+                        },
+                        {
+                            command: 'subscript',
+                            key: 'r',
+                            meta: true,
+                            shift: false
                         }
                     ]
                 }
             });
             selectElementContentsAndFire(editor.elements[0]);
             jasmine.clock().tick(1);
+            // If alt-key is not pressed, command should still be executed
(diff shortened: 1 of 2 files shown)


nmielnik:
Good catch! I think the solution you're proposing is good, just add a 'ToDo: Deprecate' comment so we can remove it in 6.0.0

j0k3r:
Yeah I'll do that and add a test too.

j0k3r:
Wow, Travis seems to have some glitch ..

nmielnik:
So I thought about it some more, and I'm thinking it might be more backward compatible if we go the route of not even checking the value of `altKey` on the event if they didn't specify whether the command cares about <kbd>alt</kbd> or not.

So, if when configuring the keyboard command, if `altKey` isn't specified, the action will be executed regardless of whether `event.altKey` is true or false.

I submitted a commit and added a bit to the test case, let me know what you think.

j0k3r:
Yeah good idea !

PR #787, opened 2015-08-20
Fix #768 + Fix #783 + Simplify and Fix Placeholder Extension

### LINKED ISSUE #768: remove the placeholder from UI when setting the content throught setContent method
i'm using the function setContent and it, in my opinion, should call
yourElement.classList.remove('medium-editor-placeholder')

@nmielnik
> I agree that setting the content should trigger the placeholder to go away, but we'll have to think a bit about the way to implement it. We wouldn't want the setContent method to remove the placeholder attribute directly, since the placeholder logic is in it's own extension and we want to make sure we keep separation of concerns. We'll just need to change around the placeholder extension a bit to know whenever the content changes.

suggestion capture an event on setContent

### LINKED ISSUE #783: Static Toolbar - Click Button - Placeholder still visible with content insertion
I am thinking this is related to issue #194 

When I click an h1 button in the static toolbar it adds content to the editor. The problem is the placeholder text is still visible until you type. The screenshot shots the blinking cursor and the custom placeholder text.

screen shot 2015-08-17 at 10 17 16 am

Live example: http://jsfiddle.net/u0L6xcuu/ , the h1 button is labeled "Lead" in the fiddle.

nmielnik:
This PR covers a bit of a re-write of how the placeholder extension handles hiding/showing the placeholder text.

The new placeholder extension leverages the `editableInput`, `focus`, and `blur` custom-events provided by the internal MediumEditor code to better detect when the placeholder text should be hidden or shown.

By listening to `editableInput`, the extension is also able to detect when calls to `setContent()` happen, which fixes #768 .

By listening to `focus`, `blur`, and `editableInput`, and checking the value of the `placeholder.hideOnClick` option, the extension is also able to prevent showing the placeholder incorrectly while clicking buttons on the toolbar, which fixes #783 

Other Issues Resolved:
* Since the placeholder extension now uses `editableInput`, MediumEditor will now always monitor for the `editableInput` event whenever it's instantiated.  This required tweaking some of the tests around `editableInput` which assumed that the event detection wouldn't happen automatically.
* Since the placeholder extension is using `editableInput`, a bug was found inside link creation when it came time to manually trigger the custom event.  In some cases, the reference to the editable element passed to the custom event was `null`, so this is now fixed as well.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -930,8 +930,6 @@
                         // we want to make sure we create a single link, and not multiple links
                         // which can happen with the built in browser functionality
                         if (commonAncestorContainer.nodeType !== 3 && startContainerParentElement === endContainerParentElement) {
-
-                            currentEditor = MediumEditor.selection.getSelectionElement(this.options.contentWindow);
                             var parentElement = (startContainerParentElement || currentEditor),
                                 fragment = this.options.ownerDocument.createDocumentFragment();
 
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -889,7 +889,13 @@
         },
 
         createLink: function (opts) {
-            var currentEditor, customEvent, i;
+            var currentEditor = MediumEditor.selection.getSelectionElement(this.options.contentWindow),
+                customEvent = {};
+
+            // Make sure the selection is within an element this editor is tracking
+            if (this.elements.indexOf(currentEditor) === -1) {
+                return;
+            }
 
             try {
                 this.events.disableCustomEvent('editableInput');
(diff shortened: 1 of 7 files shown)


j0k3r:
Interesting !
IE doesn't seems to like your changes

src/js/core.js
```diff
@@ -889,7 +889,13 @@
         },
 
         createLink: function (opts) {
-            var currentEditor, customEvent, i;
+            var currentEditor = MediumEditor.selection.getSelectionElement(this.options.contentWindow),
```
j0k3r: Then you should remove the previous assignation at line 934

nmielnik: Good point, I've added a commit.

PR #792, opened 2015-08-24
Make util.isDescendant() more efficient when browser allows for it

nmielnik:
Browsers expose a `Node.contains()` method ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains)) which checks if one node is contained with another.

We should use this function inside our `MediumEditor.util.isDescendant()` helper method as it will be more efficient to rely on the browser for this check.

However, browsers don't consistently return `true` for this if the child node being checked for is a text-node, so we'll need to feature detect, and use our old while-loop fallback in cases where it's not supported.

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -22,6 +22,18 @@
         return dest;
     }
 
+    // https://developer.mozilla.org/en-US/docs/Web/API/Node/contains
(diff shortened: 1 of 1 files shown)


j0k3r:
Interesting feature!

Issue #499, opened 2015-03-23
Need a way to control where the toolbar element is appended


Pajamaman:
Given the following code:

    <div id="container" style="background-color: #ccc; height: 50px;"></div>
    <h3>Edit Page</h3>
    <form>
        <p>Type in the area below:</p>
        <div id="editor">Blah blah blah</div>
    </form>
    <script>
        var editor = new MediumEditor('#editor', {
            elementsContainer: document.getElementById('container')
        });
    </script>

The toolbar should show up in the grey area at the top of the page. But it floats over the top of the selected text.

Demo: http://jsfiddle.net/ydun0hjs/

![screenshot from 2015-03-23 13 18 59](https://cloud.githubusercontent.com/assets/1323271/6788486/51443956-d15f-11e4-967a-6c6c486e4519.png)

Pajamaman:
I finally realized the elementsContainer option only specifies where in the DOM to add the toolbar element. It doesn't specify where it should appear. How would I go about getting the toolbar to appear in the container though?

nmielnik:
We'd need to introduce another option like `toolbarContainer` which could be passed and then the toolbar could be displayed inside.  You'd likely need to either use the `staticToolbar` option, or just fork the code for positioning the toolbar when the `toolbarContainer` option is present, so that we don't attempt to position it somewhere.

The code for appending the toolbar can be found in [`core.js`](https://github.com/daviferreira/medium-editor/blob/master/src/js/core.js#L292)

The code for positioning the toolbar is in [`toolbar.js`](https://github.com/daviferreira/medium-editor/blob/master/src/js/toolbar.js#L438)

marvwhere:
+1 (is there any progress on this?)

nmielnik:
Looks like you made some good progress on this yourself :wink:

nmielnik:
I think we have a workable solution via #796 , which as been merged to master. Closing this issue for now.

Pajamaman:
@marvwhere nice!

PR #796, opened 2015-08-30
Added toolbar-relative-container feature

marvwhere:
In relation to  this issue: https://github.com/yabwe/medium-editor/issues/499

I needed this feature for my current project, so i tried to make it kind of "official" would be happy if it makes it into master.

Not sure about specs, had no other ideas what to test, because all other features should be already covered.


nmielnik:
This looks great, thanks a ton for contribution!

Issue #248, opened 2014-06-23
disableDoubleReturn doesn't work when inside a P on Firefox


sahandsepanta:
Setting 
<code>
disableDoubleReturn : true
</code>
 works on Chromium browser but not in Firefox 30.

hellocosmin:
Confirmed to still not work in Firefox (35.0.1), <em>if</em> the editable is a paragraph. Here's a [playground](http://jsfiddle.net/gfqzr2gd/) for reference.
It <em>can</em> work in Firefox, as long as the editable is a div. [Demo](http://jsfiddle.net/gfqzr2gd/3/)

Maybe the description is incomplete or a bit misleading?
For example, I understood this can prevent double returns when cursor is anywhere <em>inside</em> the editable, not just when you're at the end of it.

Maybe having it prevent double returns triggered from anywhere inside the editable, could be considered a feature request? :)

Issue #207, opened 2014-04-18
How to use disableDoubleReturn for different behaviour?


cdolek:
Hi,

Thank you for the medium editor.

```
"disableDoubleReturn: allows/disallows two (or more) empty new lines. 
You can also set specific element behavior by using setting a 
data-disable-double-return attribute. Default: false"
```

Is this in the works or already implemented? When I set this attribute, it merely disables double returns.

Thanks again

j0k3r:
@sahandsepanta Could this be what is defined in #248 ?

nmielnik:
Seems like a feature to allow for custom html insertion when double-return occurs could be something to consider if we implement a fix for #70

csarven:
What I'm experiencing in Ubuntu Firefox 40 and Chromium 44,  where disableDoubleReturn is set to false, and double return is pressed somewhere inside a paragraph other than at the end of the paragraph:
```
<p>Hello world</p>
```
becomes
```
<p>Hello</p>
<p><br/></p>
<p>world</p>
```

I think the expected behaviour should be the same as if the caret is at the end of the line i.e., allow only a single return which results in:
```
<p>Hello</p>
<p>world</p>
```

Issue #70, opened 2013-10-30
Add option for return key adding line break, and double enter adding new paragraph


sathishmanohar:
Currently pressing the return key once creates new paragraph, which I feel is counter intuitive. By defaulting to line breaks on single return and paragraphs on double returns, It'll feel lot more intuitive, since many other editors and word processors use the same pattern.

bbnnt:
@litera is shift+enter is adding a br tag when using medium editor ?

hellocosmin:
@bbnnt yes it does, you can check the demo using a browser inspection tool.

Since someone brought this up, I have another (related) concern. Here's the scenario:

- I have a .editable on which I use ME; this element is a block level (h1, div...)
- using <kbd>Shift+Enter</kbd> correctly adds a `<br>`
- hitting <kbd>Backspace</kbd> to undo/remove that `<br/>` wraps the entire .editable contents inside a `<p></p>`

This results in incorrect formatting, since I now have a paragraph inside my h1, for example. Might be a smaller issue for web view, but in email design this is a nightmare.
I know there are ways of dealing with this (like unwrapping the contents or using something like jQuery htmlclean() plugin ), but could we consider a possible option for ME, by which we can turn paragraph wrapping on/off?

ME can be used in so many applications and it would be a really useful option to have (especially in email layout editors), instead of having to manually clean HTML or removing some code in the core.

litera:
@bbnnt As @hellocosmin pointed out, the answer to your question is **yes, <kbd>Shift + ENTER</kbd> creates a `<br />`**.

@hellocosmin: This issue you're having is very likely due to editable behaviour of specific browser. You can check individual browsers if they all act the same. but in order to solve your issue you can always *unwrap* your content if wrapping is inside a `p` element.

I've had an issue with Chrome browser creating `span` elements with inline styles that was particularly annoying because it was setting `font-size` style. So I added my code that unwrapped `span` content and everything's ok now. You can do the same for your paragraph wrapping.

hellocosmin:
@litera Checked in both Chrome and Firefox, it does exactly the same thing.

Commenting out this line, makes it work fine (i.e. not wrapping the contents in a paragraph):
https://github.com/daviferreira/medium-editor/blob/master/dist/js/medium-editor.js#L1799

bbnnt:
@litera @hellocosmin  thanks a lot lot for your hints
unfortunately, the shift+enter does creates a br :/
hope I'll get over this
[...]

PR #799, opened 2015-09-01
If disableDoubleReturn is true, prevents returns from caret position 0

csarven:
This patch lets `disableDoubleReturn` work as expected. If the caret position is at 0, there is no need to create a new line, paragraph or list item.

It should resolve one or more of these issues: https://github.com/yabwe/medium-editor/issues/70 , https://github.com/yabwe/medium-editor/issues/207 , https://github.com/yabwe/medium-editor/issues/248

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -11,7 +11,8 @@
 
             // if current text selection is empty OR previous sibling text is empty
             if ((node && node.textContent.trim() === '') ||
-                (node.previousElementSibling && node.previousElementSibling.textContent.trim() === '')) {
+                (node.previousElementSibling && node.previousElementSibling.textContent.trim() === '') ||
+                (MediumEditor.selection.getCaretOffsets(node).left === 0)) {
                 event.preventDefault();
             }
         }


nmielnik:
Would you be able to add any tests to verify this?  Maybe some test(s) that cover some of the existing issues that this may resolve so that we can close them?

csarven:
Will this do http://jsfiddle.net/fnk2tku6/ (using master dist/js/medium-editor.js with the patch applied) or is there a preferred way to create tests?

It primarily resolves this https://github.com/yabwe/medium-editor/issues/207#issuecomment-136153264 and confirms the "expected" behaviour of the disableDoubleReturn based on my understanding. I think #248 may be the only one which can be closed. This patch may be a precursor to resolving #70 and #207 .

nmielnik:
Apologies for being vague, when I mention adding test I'm suggesting something like in [`content.spec.js`](https://github.com/yabwe/medium-editor/blob/master/spec/content.spec.js#L125).  These tests are run each time you run `grunt` from the terminal.

After running grunt, you can also run all the tests by loading this in a browser:
`<medium-editor-root>/_SpecRunner.html`

nmielnik:
@csarven any progress on this?

csarven:
@nmielnik Unfortunately, I haven't had the chance to create the tests. I will have to come back to it at a later date. If you or someone else would like to pick that up, that's great.

Issue #543, opened 2015-04-08
Unwanted <span style="font-size: 18px;"><br></span> added


VictorienTardif:
Sometimes when return key is pressed, a `<span style="font-size: 18px;"><br></span>` is added.
How can I prevent that?

Thank you.

bborn:
+1

phiggins42:
@davi added the tag, but this is a contenteditable "core" issue where if you have css rules that are styling things these `span[style]` tags are created when two conflicting font-rules end up bumping up against each other. (i have a similar issue with custom styled headings having a different line-height than the following paragraph, and backspacing the paragraph into the heading leaves the remainer of the paragraph less-font-weight'd and more-line-height'ed (because the p and h3 have different font settings)

bborn:
I believe this has been discussed, but @daviferreira any more thoughts on using Scribe instead of native contenteditable?

nchase:
@bborn: is Scribe tested against IE or Safari yet, or is it still just Firefox and Chrome?

bborn:
@nchase not sure - according to [this](https://github.com/guardian/scribe/issues/151) it seems like they're not supporting older browsers (but it's not clear exactly what they mean buy that). Looks like Safari 6 is at least partially supported.

nchase:
@bborn: they run automated tests against Chrome and Firefox, but not Safari or IE.

If they aren't running-and-maintaining tests that exercise the code in Safari and IE, they aren't actually supporting them – a change to the code could break important things without anyone knowing until the code gets used in the wild, which is not safe.

bborn:
Yup - however it may just be due to their testing setup. See here: https://github.com/guardian/scribe/issues/249 (bad Selenium support for Safari)

VictorienTardif:
Thank you for your answers, and your good explication @phiggins42. 
So I assume I can't do anything about that!

I'll try to figure out what's in conflict in my case. 
Thank you!

bborn:
Here's my hack to fix it:

```
medium = new MediumEditor(.. your config here ...);
el = medium.elements[0];
$(el).on("input", function() {
  $(el).find("span[style]").contents().unwrap();
});
```

This basically just removes all spans with a style attribute.

phiggins42:
I have a similar hack in place, but it causes all sorts of strange behavior to undo stack doing that dom manipulation that way.

bborn:
[...]

PR #801, opened 2015-09-02
WIP – Handle bad span creation

j0k3r:
I ran into a bug (that might be the same as https://github.com/yabwe/medium-editor/issues/543) when merging two elements together (like, paragraphs and headings), a span with inline styles will be inserted around the contents of what was the second element.

Let's say you have:

``` html
<p>So, I called myself Pip, and came to be called Pip.</p>
<h3>I give Pirrip as my father’s family name</h3>
```

If you move your caret to the end of the first paragraph and hit <kbd>del</kbd> key, the `h3` will be merged in the paragraph and replaced by a span with ugly attributes:

``` html
<p>So, I called myself Pip, and came to be called Pip.<span style="font-size: 32px; line-height: 42px;">I give Pirrip as my father’s family name</span></p>
```

Using MutationObserver, we listen to every dom change. When it affect a paragraph and inserted node is a span, we unwrap it inside the paragraph. This works great :smile: 

 :warning: **BUT** it breaks the undo/redo. Undoing the previous action will generate that html:

``` html
<p>So, I called myself Pip, and came to be called Pip.I give Pirrip as my father’s family name<br></p>
<h3>I give Pirrip as my father’s family name</h3>
```

As you can see, it recreate the heading but leave the text of the heading inside the paragraph.

---

Todo:
- [ ] Fix tests on IE 11
- [ ] Cleanup code
- [ ] Add tests ? (even if PhantomJS < 2.0.0 doesn't handle `MutationObserver`)
- [ ] Add an option to enable this feature

---

_Related links for inspiration:_
- http://www.neotericdesign.com/blog/2013/3/working-around-chrome-s-contenteditable-span-bug
- https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver
- https://developers.google.com/web/updates/2012/02/Detect-DOM-changes-with-Mutation-Observers
- https://github.com/rafaelw/mutation-summary
- http://stackoverflow.com/q/15015019/569101


j0k3r:
While updating to Phantom 2, I ran into 3 problems.

1. related to `unwrap` method.
It seems that `el.parentNode` isn't defined in some case. Should this be related to the dom we are using in tests ?

 That's why I added this to avoid an error:

 ```diff
 +            if (null === el.parentNode) {
 +                return;
 +            }
 ```
These were the errors:

  ```
  - should not remove the <p> container when adding a link inside a top-level <p> with a single text-node child...>> TypeError: null is not an object (evaluating 'el.parentNode.removeChild') at
  >> .grunt/grunt-contrib-jasmine/src/js/util.js:9 unwrap
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> :0 forEach
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> TypeError: null is not an object (evaluating 'el.parentNode.removeChild') at
  >> .grunt/grunt-contrib-jasmine/src/js/util.js:9 unwrap
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> :0 forEach
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> TypeError: null is not an object (evaluating 'el.parentNode.removeChild') at
  >> .grunt/grunt-contrib-jasmine/src/js/util.js:9 unwrap
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> :0 forEach
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> TypeError: null is not an object (evaluating 'el.parentNode.removeChild') at
  >> .grunt/grunt-contrib-jasmine/src/js/util.js:9 unwrap
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> :0 forEach
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9


  - should call init when instantiated...>> TypeError: null is not an object (evaluating 'el.parentNode.removeChild') at
  >> .grunt/grunt-contrib-jasmine/src/js/util.js:9 unwrap
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  >> :0 forEach
  >> .grunt/grunt-contrib-jasmine/src/js/core.js:9
  ```

2. Problem with `window.console.error`
It seems that auto-link extension generate an [a console error](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/auto-link.js#L73) in some bad case and it generates an error in the tests suite, which doesn't kill the test (that's odd...).

  ```
  - should stop attempting to auto-link on keypress if an error is encountered...>> Failed to perform linking Error: DOM ERROR at
  >> .grunt/grunt-contrib-jasmine/src/js/extensions/auto-link.js:9
  ```

3. IE :disappointed: 
All IE tests are now failing ...

nmielnik:
@j0k3r for better stack traces, try commenting out [this line](https://github.com/yabwe/medium-editor/blob/master/Gruntfile.js#L136) locally to disable the istanbul code coverage.  It will give you actual line numbers in the stack trace.

> It seems that auto-link extension generate an a console error in some bad case and it generates an error in the tests suite, which doesn't kill the test (that's odd...).

This is actually by design.  The code was written s.t. if any error happens while attempting to auto-link, the extension will disable itself (since the code will attempt to repeat the same operation on every keystroke or modification to the editor).  The test case was written to intentionally throw an error and verify that auto-link stopped listening for events.

> IE :disappointed: 
All IE tests are now failing ...

Can you try merging in the Edge branch and seeing if those fail too?  I'll try to take a look at your branch when I get a minute and see if I can help with any of the failures.

j0k3r:
> The test case was written to intentionally throw an error and verify that auto-link stopped listening for events.

Oh ok I just saw that the error was already fired *and* logged in a previous build. It just displayed red instead of being a simple log using Phantom 2.

> Can you try merging in the Edge branch

I’ve rebased my branch with your Edge fixes

PR #811, opened 2015-09-12
Fix #790 - Don't auto-link across multiple list-items

### LINKED ISSUE #790: Auto Link creates links with parts of different <li> tags
I've found if you have an `li` that ends in `.`, and the next `li` starts with a TLD, then auto link creates a link with the end of the first tag, and the start of the second.

[Codepen](http://codepen.io/anon/pen/LVwYXO) to illustrate the issue.

nmielnik:
This fixes #790 where auto-link was creating links that spanned across consecutive list-items.

The root cause is the way auto-link looks for elements with text that may contain valid urls which we should convert into links.  Previously it was searching for all elements in the `MediumEditor.util.blockContainerElementNames` array, which included elements like `ol`, `ul`, and `dl`. However, the text content of these elements is just the text within each of their child list-items concatenated together without new-lines.  The result was if concatenating two or more list-items together resulted in a valid url, the code would attempt to create a link which spanned those multiple list-items.

A simple fix for this is to filter out the specific elements we want to ignore when checking for auto-linkable text, since we will be checking their child elements anyway (`'ol', 'ul', 'dl', 'table', 'tbody', 'tfoot', 'tr'`)

### DIFF
--- a/demo/auto-link.html
+++ b/demo/auto-link.html
@@ -33,7 +33,10 @@
     <script>
         var editor = new MediumEditor('.editable', {
             buttonLabels: 'fontawesome',
-            autoLink: true
+            autoLink: true,
+            toolbar: {
+                buttons: ['bold', 'italic', 'unorderedlist', 'orderedlist', 'anchor']
+            }
         }),
         cssLink = document.getElementById('medium-editor-theme');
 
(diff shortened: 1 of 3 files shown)


j0k3r:
:+1:

PR #817, opened 2015-09-18
More autolink fixes

nmielnik:
My original attempt to fix auto-link issues within nested block elements (ie `<li>` elements inside of `<ol>` elements) was very targeted towards specific common cases where we know we don't want to auto-link the `textContent` of the element, but rather auto-link the content of its children (`'ol', 'ul', 'dl', 'table', 'tbody', 'tfoot', 'tr'` elements).

However, there are still many other cases where we don't want to auto-link the `textContent` of a block element, since it can have other block elements:

```html
<div>
  <blockquote>
    <ol>
      <li>www</li>
      <li>.example.com</li>
    </ol>
  </blockquote>
<div>
```
In this example, the auto-link code would try to create a link around the entire `<ol>` element, since the code would first find the `<blockquote>` and then see that its textContent ('www.example.com') is a valid URL. 

The fix is to do a better job of finding the unique pieces of the content which contain no block elements.

So there is now a `MediumEditor.util.splitByBlockElements(element)` method which will find these pieces and return them as an array.  Calling this helper on the above example would return `[<li>www</li>, <li>.example.com</li>]` elements.  Thus, when auto-link calls this, it would check each of the `<li>` elements to see if they individually have auto-linkable text, and never look at the `<blockquote>` or `<ol>` for textContent.

### DIFF
--- a/spec/auto-link.spec.js
+++ b/spec/auto-link.spec.js
@@ -465,6 +465,15 @@ describe('Autolink', function () {
                 expect(lastLi.firstChild).toBe(links[0]);
                 expect(lastLi.textContent).toBe('www.example.com');
             });
+
+            it('should not create a link which spans multiple list items', function () {
+                this.el.innerHTML = '<blockquote><ol><li>www</li><li>.example.com</li></ol></blockquote>';
+
+                selectElementContentsAndFire(this.el);
+                triggerAutolinking(this.el);
+                var links = this.el.getElementsByTagName('a');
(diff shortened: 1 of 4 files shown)


j0k3r:
Why don't you create a test with the case you described in this PR ?

nmielnik:
The change in `auto-link.spec.js` was to account for the case I called out here :smile:

j0k3r:
Great :+1:

Issue #803, opened 2015-09-04
Changing link on images doesn't work


l2aelba:
![zdvyj1ivrh](https://cloud.githubusercontent.com/assets/983458/9678791/134b7b12-52e9-11e5-8251-f0046b328e69.gif)

I dont know what happen

Demo : http://jsfiddle.net/ucdp9cqo/

l2aelba:
+ On normal text too

j0k3r:
I guess we check that the `a` node contains text and then update the link. If you add some text inside the link and then update the link you'll get that (the default link was `google.com` and I updated it to `google.fr`):

Before:

```html
<a href="http://google.com">test <img src="https://placeholdit.imgix.net/~text?txtsize=33&txt=350%C3%97150&w=350&h=150"></a>
```

After:

```html
<a href="http://google.fr">test </a>
<a href="http://google.com"><img src="https://placeholdit.imgix.net/~text?txtsize=33&amp;txt=350%C3%97150&amp;w=350&amp;h=150"></a>
```

l2aelba:
@j0k3r  No problem, i hope this can be fixed ASAP :D Please + Thanks

l2aelba:
Any fix ?

j0k3r:
@l2aelba not yet, but if you have time to find a way to fix this bug, you're welcome!

l2aelba:
@j0k3r  thanks :D

anilmaurya:
@l2aelba Can you confirm if your issue is resolved or not ?

l2aelba:
@anilmaurya  Nope sorry :( Not even Chrome

anilmaurya:
@l2aelba Merged fixed is not released, hence its not present in `dist/js/medium-editor.js` . Your js fiddle is using `dist/js/medium-editor.js` therefore its not working .

j0k3r:
I've just bumped 5.8.3 to include your fixes @anilmaurya

anilmaurya:
@j0k3r thanks :+1:

l2aelba:
@j0k3r , @anilmaurya , Seems like still doest fixed ?

http://jsfiddle.net/ucdp9cqo/

l2aelba:
Seems like it doesnt work now again :(

nmielnik:
Bummer, looks like this got broken again somehow...what's an even bigger bummer is that the tests aren't failing.

[...]

PR #819, opened 2015-09-21
Fix changing link on images, #803

anilmaurya:


src/js/core.js
```diff
@@ -981,6 +981,14 @@
                                     editableElementIndex: exportedSelection.editableElementIndex
                                 }
                             );
+                            // If textNodes are not present, when changing link on images
+                            // ex: <a><img src="http://image.test.com"></a>, change fragment to currRange.startContainer
+                            // and set textNodes array to [imageElement, imageElement]
+                            if (textNodes.length === 0) {
+                                fragment = this.options.ownerDocument.createDocumentFragment();
+                                fragment.appendChild(currRange.startContainer.cloneNode(true));
+                                textNodes = [fragment.firstChild.firstChild, fragment.firstChild.lastChild];
```
nmielnik: Could `currRange.startContainer` ever be a text node?  If so, the reference to `fragment.firstChild.firstChild` and `fragment.firstChild.lastChild` could be undefined, in which case I'm not sure what the rest of the code would do.  It's also possible that this is handled ok, but I wanted to at least make sure you know what would happen in that case.

nmielnik:
@anilmaurya Thanks a ton for the pull request and the test case, this is great!

Since you've linked to issue #803, have you verified that once an image is wrapped in a link, that you're able to actual change the url of the link?  Do you think it'd be useful to have a test case that attempts to change the href of an existing link, perhaps in `anchor-preview.spec.js` or `anchor.spec.js`?

src/js/core.js
anilmaurya: If `currRange.startContainer` is text node than textNodes.length will never equal to 0 .

anilmaurya:
@nmielnik Yes, I checked that once an image is wrapped in a link than I am able to change its url.

I updated demo/auto-link.html , to check this case. I think my test case is sufficient for this scenario, in case you think its not enough and requires more test case, Let me know.

nmielnik:
Thanks!

nmielnik:
@anilmaurya looks like the tests are failing on Firefox and IE11, would you be able to take a look?

anilmaurya:
@nmielnik ok, I will check it.

nmielnik:
### Firefox:
Firefox Failure

### IE 11:
![IE 11 Failure]
(https://assets.saucelabs.com/jobs/2912feaf41ca48949bc7dd1cadd68b96/0004screenshot.png)

PR #820, opened 2015-09-23
Fix changing links on images for firefox, #803

anilmaurya:
fix anchor specs to select aTag when changing links

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -434,8 +434,10 @@ describe('Anchor Button TestCase', function () {
             spyOn(MediumEditor.prototype, 'createLink').and.callThrough();
             var editor = this.newMediumEditor('.editor'),
                 toolbar = editor.getExtensionByName('toolbar'),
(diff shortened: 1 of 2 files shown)


anilmaurya:
@nmielnik How do I check if this is running fine on IE11 . I don't have windows machine and I don't know how to check it on browserstack. Can you help ?

nmielnik:
Unfortunately I'm not quite sure of a way to trigger the tests to run against SauceLabs via an external fork.  There is a way to use SauceLabs to grab an OS + Browser session and run it against your local machine via sauce connect, but you have to setup sauce connect properly, and I'm not sure if you'd be able to leverage our sauce labs account.

So, you can always install something like Virtual Box and download a Windows + IE11 VM to run the tests in, which is what I'm going to use to run the tests again.  I'll pull down your fork and run it in my VM and let you know.

anilmaurya:
@nmielnik Thanks for the info.

PR #836, opened 2015-10-09
Fix font-awesome.css link in demo/auto-link.html

anilmaurya:
screen shot 2015-10-09 at 10 28 35 pm

### DIFF
--- a/demo/auto-link.html
+++ b/demo/auto-link.html
@@ -4,7 +4,7 @@
     <meta charset="UTF-8">
     <title>medium editor | demo</title>
     <link rel="stylesheet" href="css/demo.css">
-    <link rel="stylesheet" href="//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.css">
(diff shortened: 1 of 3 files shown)


daviferreira:
This is actually on purpose, to make use of the current protocol. If we are going to change it, we need to change on all demo files.  But I'm not sure about it :)

anilmaurya:
@daviferreira I didn't understand your point. Is this working for you ? are you serving demo files with http server ?

nmielnik:
@daviferreira we use a completely different file for the demo page that we host right?  If we leave that hosted page using the relative protocol URLs, but change the internal demo files within the repo to use a hard-coded protocol, we should still be ok right?

I might be forgetting another reason why we changed the urls to be relative-protocol in the actual source code...

daviferreira:
@anilmaurya yeap, actually I always use https://simbco.github.io/httpster/ :)

But I agree, let's change it. If you could update this PR changing al demo files please.

anilmaurya:
@daviferreira sure , I will update this PR with all demo files change.

anilmaurya:
@daviferreira only 2 other demo files had this issue. Updated PR

daviferreira:
Thanks!

Issue #824, opened 2015-09-28
Embedded links allow trailing whitespaces


pranthiks:
When I embed links using the medium editor toolbar, I can create links that have trailing white spaces at the end. This becomes a problem when I'm pasting links that I've copied from somewhere else and I've unknowingly copied some trailing white spaces at the end.

I think it would be good to trim any trailing white spaces before and after any links entered into the url field. 

Are there plans to fix this? Also willing to submit a pull request for this.

<img width="665" alt="screen shot 2015-09-28 at 3 38 07 pm" src="https://cloud.githubusercontent.com/assets/7840137/10150932/80756092-65f7-11e5-89c2-4ad5762c9524.png">

j0k3r:
Good point, feel free to submit a PR about that. It's a good candidate for an _easy first bug_. :)

nmielnik:
Fixed via #838

PR #838, opened 2015-10-10
removed trailing white spaces from the link and undid the changes in dist files made in earlier commit

rishijain:
Fix for https://github.com/yabwe/medium-editor/issues/824

1. Squashed multiple commits into 1 commit.
2. trimmed the url.
3. Removed unnecessary conditional check with trimmed values as now we are initialising  it with the trimmed values itself.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -72,6 +72,43 @@ describe('Anchor Button TestCase', function () {
             expect(this.el.innerHTML.indexOf('<a href="test">lorem ipsum</a>')).toBe(0);
         });
 
+        it('should remove the extra white spaces in the link when user presses enter', function () {
+            spyOn(MediumEditor.prototype, 'createLink').and.callThrough();
+            var editor = this.newMediumEditor('.editor'),
+                toolbar = editor.getExtensionByName('toolbar'),
+                button, input;
+
+            selectElementContents(editor.elements[0]);
+            button = toolbar.getToolbarElement().querySelector('[data-action="createLink"]');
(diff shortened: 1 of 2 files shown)


src/js/core.js
```diff
@@ -899,7 +899,7 @@
 
             try {
                 this.events.disableCustomEvent('editableInput');
-                if (opts.url && opts.url.trim().length > 0) {
+                if (opts.url && opts.url.length > 0) {
```
nmielnik: Why would we want to allow for the creation of a link with a url which is only whitespace here?

rishijain: No.
See the change I made in src/js/extensions/anchor.js file. I am trimming down the url before setting in opts.url variable. So opts.url.length would be equal to 0 if it is only whitespace there.

rishijain: I have added a test case testing the scenario you asked the question for.

nmielnik: So `createLink()` is a public method that can (and is) used by other users externally.  Would leaving the 2 calls to `trim()` from within this `createLink()` method still allow for you to fix the issue with the anchor extension?

rishijain: I missed this part "users can call this function externally". So I guess I need to add those 2 trim() calls back because otherwise it would allow users to add "empty strings" as url which is not correct.

I will update the PR.

rishijain: I have updated the PR with required calls to trim() . @nmielnik

nmielnik:
Awesome, thanks for the PR!

Issue #825, opened 2015-09-29
Anchor preview only working when editor is not focused


gerherz:
I'm having a problem with the anchor preview extension. By using the inspector, I can tell it only works when the editor is not focused. I'm using the 5.8.2 version with this config:

```javascript
{
	elementsContainer: document.getElementById('cont'+this.props.name),
	buttonLabels: 'fontawesome',
	spellcheck: false,
	imageDragging: false,
	anchor: {
		placeholderText: "http://",
		targetCheckbox: true,
		targetCheckboxText: "Abrir en una nueva ventana"
	},
	toolbar: {
		buttons: ['bold', 'italic', 'underline', 'strikethrough', 'subscript', 	'superscript', 'anchor', 'table', 'h2', 'orderedlist', 'unorderedlist', 'justifyLeft', 'justifyCenter', 'justifyRight', 'justifyFull', 'fontcolor', 'embed', 'imagen', 'removeFormat'],
		static: true,
		updateOnEmptySelection: true
	},
	extensions: {
		table: new MediumEditorTables(),
		fontcolor: new MediumEditorForeColor(),
		embed: new MediumEditorEmbed(),
		imagen: new MediumEditorInsertImage()
	},
	autoLink: true,
	paste: {
		cleanPastedHTML: true,
		cleanAttrs: ['class'],
		forcePlainText: false
	}
}
```

j0k3r:
Could you provide a jsfiddle to reproduce the bug ?

gerherz:
Sure: http://jsfiddle.net/ujfev0aj/
It's happening both on Firefox and Chromium.

DylanPiercey:
+1 to this issue.

nmielnik:
So as far as I remember, there was intent around not having the anchor-preview show up when the toolbar is visible. You can see the code that does this [here](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/anchor-preview.js#L168-L172) within the anchor-preview extension.

The goal of this code was to prevent both the toolbar and the anchor-preview from showing up at the same time, which could be pretty messy in the UI.  Normally, if you focus into the editor but don't select any text, the toolbar will not show up, thus you can hover over an anchor and see the anchor preview.

In this case, you're using both the `toolbar.static: true` and `toolbar.updateOnEmptySelection: true` options.  Because of that, the toolbar will actually be displayed whenever any focus is inside the editor, and since the toolbar is visible, this code prevents the anchor-preview from displaying.

[...]

PR #839, opened 2015-10-10
New anchor preview option to show when toolbar is visible

gerherz:
New option as discussed here: https://github.com/yabwe/medium-editor/issues/825

### DIFF
--- a/OPTIONS.md
+++ b/OPTIONS.md
@@ -105,7 +105,7 @@ CSS class added to active buttons in the toolbar.
 #### `buttonLabels`
 **Default:** `false`
 
-Custom content for the toolbar buttons. 
+Custom content for the toolbar buttons.
 
 **Valid Values:**
 * `false`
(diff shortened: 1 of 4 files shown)


nmielnik:
Awesome, the change looks great!  Were you able to verify that this fixed #825 for your use case?

2 Things:
1. Could you add a couple tests?
2. Could you update the documentation (README.md & OPTIONS.md) to describe the few option?

Thanks again!

gerherz:
Yes, my use case is fixed. I have just updated the documentation on both files.

As for the tests, I couldn't find a test that checks that the preview is not visible when the toolbar is. Is there any, so that I can work on top of it?

nmielnik:
You can see a test case that is similar to what you'd be testing [here](https://github.com/yabwe/medium-editor/blob/master/spec/anchor-preview.spec.js#L220-L236) (although the description of the test case isn't quite accurate).  You can also check `toolbar.spec.js` for cases that check for the existence/visibility of the toolbar.

gerherz:
Thanks, I've added two tests. Please take a look at them to make sure that they are OK.

nmielnik:
Awesome, thanks @gerherz !

Issue #805, opened 2015-09-09
Support for font type/family?


sqzr1:
Medium editor is a great alternative to the regular editors out there. Are there any plugins/addons for changing the Font-Family of an element? Ie, like the Font-Size addon?

j0k3r:
Nope, there is no such extension at the moment.
But feel free to provide this one ! You can take inspiration from fontSize extension and stuff in the wiki

anilmaurya:
I think this issue should be closed now , @j0k3r is anything pending in this issue ?

j0k3r:
@anilmaurya :+1: 

@sqzr1 You can now see the [Fond Name Extension](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/fontname.js).

PR #845, opened 2015-10-17
Font Name Extension

gerherz:
https://github.com/yabwe/medium-editor/issues/805
Based on the font size extension file, I've created a new font face extension. On line 94, the different font options are listed. How should this array be populated? Which font faces should be included?

### DIFF
--- a/src/js/extensions/fontname.js
+++ b/src/js/extensions/fontname.js
@@ -1,0 +1,170 @@
+(function () {
+    'use strict';
+
+    var FontNameForm = MediumEditor.extensions.form.extend({
+
+        name: 'fontname',
+        action: 'fontName',
+        aria: 'change font name',
+        contentDefault: '&#xB1;', // ±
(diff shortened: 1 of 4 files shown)


src/js/extensions/fontname.js
```diff
@@ -0,0 +1,183 @@
...
+
+        // form creation and event handling
+        createForm: function () {
+            var doc = this.document,
+                form = doc.createElement('div'),
+                select = doc.createElement('select'),
+                close = doc.createElement('a'),
+                save = doc.createElement('a'),
+                option,
+                fonts = ['', 'Arial', 'Helvetica', 'Times New Roman'];
```
j0k3r: Shouldn't this be defined as an option of the extension instead?

gerherz:
OK, that's why I asked. It's changed now to use the "fontName -> fonts" option.

PR #846, opened 2015-10-18
Embed extension

gerherz:
I coded a new extension for inserting embeds. I know there's medium-editor-insert-plugin, but I don't like it being dependent on jQuery.

This extension presents the user with an input and then inserts its content to the editor. In between, several transformations can be defined to clean, transform or validate the content. These transformations can be overriden through the embed:transforms option as an array of objects containing the properties "pattern" (a regex) and "embed" (a string). I've added a default where a Youtube URL is converted to an embed:
```javascript
transforms: [
            {
                pattern: /https?:\/\/www\.youtube\.com\/watch\?.*?v=([a-zA-Z0-9]{11})/,
                embed: '<iframe width="420" height="315" src="https://www.youtube.com/embed/$1" frameborder="0" allowfullscreen></iframe>'
            }
        ],
```

I had to code this extension for a project I'm working on and though you could use it. If, for any reason, it's not something you would like to add to the project or think it should live on a separate repository, feel free to discard it.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -24,6 +24,7 @@ module.exports = function (grunt) {
             'src/js/extensions/anchor.js',
             'src/js/extensions/anchor-preview.js',
             'src/js/extensions/auto-link.js',
+            'src/js/extensions/embed.js',
             'src/js/extensions/file-dragging.js',
             'src/js/extensions/keyboard-commands.js',
             'src/js/extensions/fontsize.js',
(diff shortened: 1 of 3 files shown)


j0k3r:
Well, for having played around with medium-editor-insert-plugin, I can say that handling embed isn't as simple as a regex to convert an url to an iframe.
For example, your Youtube pattern doesn't with the short domain `youtu.be`. I know you can improve the regex but I don't think this the right way to handle the case.

Maybe instead of having that extension into the core, it could a really great example in the wiki on how to create a simple extension to handle _basic_ embed.

PR #858, opened 2015-10-29
Update CUSTOM-EVENTS.md

sohje:
Some typos


nmielnik:
:disappointed: oops!

Thanks for spotting this, could you fix line 144 as well?  You'll need to update that so that the table of contents link will actually scroll to the correct location...

sohje:
Done ;]

nmielnik:
Thanks!

PR #859, opened 2015-10-29
Remove carriage returns

stvkas:
In Unix vim, each line ends with a carriage return character, 0xD in Windows, which vim displays as ^M

Before
screen shot 2015-10-29 at 3 30 18 pm

After
screen shot 2015-10-29 at 3 57 06 pm


nmielnik:
Was this issue present in the other demo files?

stvkas:
I only saw it in pass-instance.html, all the other demo files were fine.

j0k3r:
Thanks!

PR #863, opened 2015-10-30
Fix table extension demo

anilmaurya:


nmielnik:
Thanks! had forgotten all about this after we implemented the new extension model in v5.0.

PR #870, opened 2015-11-05
Singleenterblockelement

rishijain:
For https://github.com/yabwe/medium-editor/issues/70

Note:
1. Default value of the option ```singleEnterBlockElement``` is ```true```.
2. pressing enter first time adds a ```<br>``` tag.
3. pressing enter again removes the ```<br>``` tag and adds a ```<p>``` element as sibling of the selected node.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -22,15 +22,25 @@
                 textRight,
                 newEl;
 
+            if (p.nodeName.toLowerCase() !== 'p') {
+                return;
+            }
+
             textLeft = nodeHtml.substring(0, caretPositions.left);
             textRight = nodeHtml.substring(caretPositions.left, nodeHtml.length);
+
+            event.preventDefault();
 
             //if enter is being pressed twice
             if (textRight.substring(0, 4) === '<br>') {
                 //need to remove the last added br tag, and add a new p tag
-                event.preventDefault();
+                p.innerHTML = textLeft;
+                var newPElement = document.createElement('p');
+                textRight = textRight.substring(4, textRight.length);
+                newPElement.innerHTML = textRight;
+                p.parentNode.insertBefore(newPElement, p.nextSibling);
+                MediumEditor.selection.moveCursor(this.options.ownerDocument, newPElement, 1);
             } else {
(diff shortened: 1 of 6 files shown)


src/js/core.js
```diff
@@ -14,16 +14,49 @@
...
             event.preventDefault();
-        } else if (this.options.disableDoubleReturn || element.getAttribute('data-disable-double-return')) {
-            var node = MediumEditor.selection.getSelectionStart(this.options.ownerDocument);
 
-            // if current text selection is empty OR previous sibling text is empty OR it is not a list
-            if ((node && node.textContent.trim() === '' && node.nodeName.toLowerCase() !== 'li') ||
-                (node.previousElementSibling && node.previousElementSibling.nodeName.toLowerCase() !== 'br' &&
-                 node.previousElementSibling.textContent.trim() === '')) {
+            //if enter is being pressed twice
+            if (textRight.substring(0, 4) === '<br>') {
```
nmielnik: Is this consistent across browsers?  I would be surprised if each browser returned `<br>` for this, and some didn't have something like `<br/>` or `<BR>` or something else annoying and inconsistent.

src/js/core.js
```diff
@@ -14,16 +14,49 @@
...
+                //need to remove the last added br tag, and add a new p tag
+                p.innerHTML = textLeft;
+                var newPElement = document.createElement('p');
+                textRight = textRight.substring(4, textRight.length);
+                newPElement.innerHTML = textRight;
+                p.parentNode.insertBefore(newPElement, p.nextSibling);
```
nmielnik: Similar to feedback for #879, it would be great if we could accomplish this "wrap the text in a new `<p>`" via something like `MediumEditor.util.execFormatBlock()` so that we lean on the browser to handle it correctly, and we do it in a way that will work for undo/redo.

nmielnik:
Could you add documentation for this new option into the `README.md` as well as `OPTIONS.md`?

rishijain:
I am closing this PR right now. Will retake it on a new branch on remote.

Issue #874, opened 2015-11-10
Change in text size affects whole paragraph


lizraeli:
When I highlight a certain area of text and click on H2, H3, or blockquote, the entire paragraph that contains my selection changes size. I checked on both Chrome and Firefox.

j0k3r:
Could you provide a jsfiddle to reproduce the bug? Thanks

lizraeli:
http://jsfiddle.net/gh/get/library/pure/lizraeli/medium-editor/tree/master/demo

j0k3r:
Oh I see. It has always worked like that. And medium.com has the same behavior.

When you select a part of a paragraph and change the _main_ tag to "block" element (like h1, h2, blockquote, ul, etc..) it converts the paragraph to the block element. Link, bold, italic, etc .. aren't block element.

This is something that won't be fixed I guess.

nmielnik:
Agreed, this is behavior that is common not just in contenteditable or browser editors, but client editors as well.

PR #877, opened 2015-11-14
Updated checklink function to check if a link is a phone number

lizraeli:
Updated checklink function to check if a link is a valid phone number. The test is quite liberal to allow for US and European phone numbers.
If the string is recognized as a phone number, prepend  'tel:'.
otherwise perform previous https / ftps / rtmpt / mailto check.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -200,7 +200,7 @@ describe('Anchor Button TestCase', function () {
             expect(link).not.toBeNull();
             expect(link.href).toBe('http://test.com/');
         });
- 	it('should add tel: if need be and linkValidation option is set to true', function () {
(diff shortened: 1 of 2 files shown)


j0k3r:
Thanks !

This should fix #874.
Could add a test also?

I don't know if it's better to enable this feature using an option or not. @yabwe/owners any thoughts on that?

lizraeli:
No problem, I'm glad to contribute! I added a test to anchor.spec.js.

nmielnik:
This will only work when the user sets `linkValidation` to `true`, so I'm ok with not making this be behind an additional option.  If we needed to in the future, we could always have it be an option that defaults to true, and not break anyone.

Issue #873, opened 2015-11-09
Specify phone number on anchor select.


DylanPiercey:
Currently it looks like the anchor option will detect emails and automatically add mailto: however does not do the same for phone numbers. I am not sure if this has been resolved (although I couldn't see anything in the change logs) and I am using version 5.0.0

Is there something I am missing or can this be added?

j0k3r:
No, we support `tel:` link in the previous release.
We only support `http(s)`, `ftp(s)`, `rtmp(t)` & `mailto`

:arrow_right: https://github.com/yabwe/medium-editor/blob/76164c997e2cb326ecf035aef63bfb8400ddb881/src/js/extensions/anchor.js#L235

Feel free to add it! :)

ppoulsen:
I added `tel:` as a recognized scheme in #878.

As for @DylanPiercey's original question about automatically adding `mailto:` to emails, I don't believe medium-editor does anything to recognize emails in the anchor and add `mailto:` to them. It will only add `http://` when it doesn't find a matching scheme and `anchor.linkValidation === true`.

nmielnik:
Yeah, if an email address is automatically being prefixed with `mailto:` I don't believe it's our code that's doing it (unless you're using the auto-link feature, but even then I'd be surprised if it was auto adding this).  Some browser may do this automatically, as I know IE had auto-linking built into it for a while.

I think there are 2 ways to think about this:
1. When user enters something into the anchor form that looks like an email or phone number, automatically adding `mailto:` or `tel:`
2. For the auto-link extension, automatically finding anything that looks like an email or phone number, and turning it into a link.

Option 2 would definitely be much trickier, but I still think Option 1 is a simple thing that someone could pick up easily, given they have some reliable regexs.

rishijain:
I think somewhere between https://github.com/yabwe/medium-editor/pull/878 and https://github.com/yabwe/medium-editor/pull/877 this issue got resolved. Closing this. 🔒

PR #878, opened 2015-11-15
Link Validation - Allow More Schemes

ppoulsen:
This uses a far more liberal regular expression when `anchor.linkValidation === true`. Before, `http://`, `https://`, `ftp://`, `ftps://`, `rtmp://`, `rtmpt://`, and `mailto:` were the only allowed schemes (and `mailto:` could be anywhere in the URL). Now, any set of alphabetical characters before `://` is considered valid, as well as the protocol-relative `//` and the external schemes `mailto:`, `tel:`, and `maps:`. The regular expression is also now case-insensitive.

No additional tests added to the existing tests [for adding http:// when no scheme provided](https://github.com/yabwe/medium-editor/issues/873#issuecomment-155103024) and [for not changing when valid](https://github.com/yabwe/medium-editor/blob/bdd19e3153bf07b1434df8506d2aee77e04d11c3/spec/anchor.spec.js#L203).

**Further Reading**
[Generic scheme regular expression retrieved from here.](http://stackoverflow.com/a/19709846)
[Gist on external schemes.](https://gist.github.com/jonathantneal/6301455)


nmielnik:
@ppoulsen could you add a test to verify that some of the urls you're supporting will make it through the `checkLinkFormat()` check?  You could add unit tests that call `checkLinkFormat()` directly with various schemes, or you can add more tests similar to [this test](https://github.com/yabwe/medium-editor/blob/master/spec/anchor.spec.js#L186).

nmielnik:
Apologies @ppoulsen but it looks like #877 conflicted with your changes.  I know your PR came first, but the other PR was ready to go (with a test) so it got merged in first.  If you could fix the merge conflict and add a couple tests, we'll make sure this gets in quickly.

Thanks again for the submission.

ppoulsen:
@nmielnik will do!

ppoulsen:
@nmielnik Alright, should be good to go!

nmielnik:
Awesome, thanks @ppoulsen !

Issue #143, opened 2014-02-11
Delete quote on backspace


websirnik:
When the cursor is at the beginning of the quote, I see users trying to delete the quote by pressing backspace. I think this make sense and this flow should be enabled by default.

![](http://cl.ly/image/0c1j3W2y153Q/Screen_Shot_2014-02-11_at_16_56_05-4.png)

rishijain:
I think it is already fixed for list. I have added a fix for quote in the PR https://github.com/yabwe/medium-editor/issues/879

PR #879, opened 2015-11-17
fixed 143 👊

rishijain:
Fixed https://github.com/yabwe/medium-editor/issues/143

Upon pressing backspace key at the start of the ```blockquote``` tag, changes it to ```p``` tag.
Upon pressing backspace key at any position other than start of the element, default action continues.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -131,23 +131,13 @@
 
             event.preventDefault();
         } else if (MediumEditor.util.isKey(event, MediumEditor.util.keyCode.BACKSPACE) &&
-                (tagName === 'blockquote' || node.parentElement.nodeName.toLowerCase() === 'blockquote') &&
+                (MediumEditor.util.getClosestTag(node, 'blockquote') !== false) &&
                 MediumEditor.selection.getCaretOffsets(node).left === 0) {
 
             // when cursor is at the begining of the element and the element is <blockquote>
             // then pressing backspace key should change the <blockquote> to a <p> tag
             event.preventDefault();
-
-            //if <blockquote> is the parent element, then make it the current node
-            if (node.parentElement.nodeName.toLowerCase() === 'blockquote') {
-                node = node.parentElement;
-            }
-
-            p = this.options.ownerDocument.createElement('p');
-            p.innerHTML = node.innerHTML;
-            node.parentElement.insertBefore(p, node);
-            node.parentElement.removeChild(node);
-            MediumEditor.selection.moveCursor(this.options.ownerDocument, p, 0);
+            MediumEditor.util.execFormatBlock(this.options.ownerDocument, 'p');
         }
     }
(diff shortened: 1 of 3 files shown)


j0k3r:
What happend after hitting backspace key if you <kbd>CTRL</kbd>+<kbd>Z</kbd> ?

rishijain:
I see. It adds another ```<p>``` tag with same innerHTML.

nmielnik:
@rishijain thanks for tackling yet another one of our long-standing issues!

As for ways to attempt to preserve undo/redo history:

The first thing I would try would be selecting the entire `<blockquote>` element (adding it to the current selection via the selection api or medium-editor's selection helper methods), and calling `MediumEditor.util.execFormatBlock(this.options.ownerDocument, 'p')`

That should convert the entire block element into a `<p>` in a way that the browser is ok with, thus undo/redo should work fine.

You can then restore the selection to be at the beginning of the element.

nmielnik:
Also, I know that other code very close to where you added this is potentially problematic for undo/redo as well.  We just hadn't been thinking about that at the time it was added, but ideally we would go back and explorer ways to some of the other code within this `handleBlockDeleteKeydowns()` function in a more undo/redo friendly manor.

rishijain:
Thanks @nmielnik  for the suggestions. It has helped me understand the medium-editor code a little bit more.

src/js/core.js
```diff
@@ -130,6 +130,21 @@
...
+        } else if (MediumEditor.util.isKey(event, MediumEditor.util.keyCode.BACKSPACE) &&
+                (tagName === 'blockquote' || node.parentElement.nodeName.toLowerCase() === 'blockquote') &&
+                MediumEditor.selection.getCaretOffsets(node).left === 0) {
+
+            // when cursor is at the begining of the element and the element is <blockquote>
+            // then pressing backspace key should change the <blockquote> to a <p> tag
+            event.preventDefault();
+
+            //if <blockquote> is the parent element, then make it the current node
+            if (node.parentElement.nodeName.toLowerCase() === 'blockquote') {
```
nmielnik: I think I may have steered you in the wrong direction here.  I believe calling `execFormatBlock()` will work without you needing to actually select the entire block.  So, you might be able to remove this entire conditional and just have these 2 lines:

```js
event.preventDefault();
MediumEditor.util.execFormatBlock(this.options.ownerDocument, 'p');
```

Could you double check to see if that works?  If so, that will make this code nice an concise.  Apologies again for giving some bad info :worried:

rishijain: the conditional is added if ```blockquote``` has child elements inside it. For ex: ```<blockquote><b>lorem</b></blockquote>```. So need to select the whole ```blockquote``` element and then call ```execFormatBlock``` on it.

Thoughts?

PR #885, opened 2015-11-22
Use editableInput instead of input event for textarea syncing

orthes:
Is there a reason why a basic ```input``` event is used for textarea syncing? [My plugin](https://github.com/orthes/medium-editor-insert-plugin) fires your custom ```editableInput``` event everytime an image is added/deleted and this causes a problem when using textareas, because ```input``` event is never fired and textarea is not sycned (https://github.com/orthes/medium-editor-insert-plugin/issues/238).

```editor.serliaze()``` returns correct content, ```editor.subscribe('editableInput')``` is fired, but textarea is not synced.

I also noticed that the editor usually never fires ```input``` event or listens for it. Instead you always use ```editableInput```.

So in this small PR I simply replaced ```input``` event with ```editableInput```.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -330,6 +330,8 @@
     }
 
     function initElements() {
+        var isTextareaUsed = false;
+
         this.elements.forEach(function (element, index) {
             if (!this.options.disableEditing && !element.getAttribute('data-disable-editing')) {
                 element.setAttribute('contentEditable', true);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/js/core.js [hunk 1]
```diff
@@ -339,17 +339,14 @@
             element.setAttribute('role', 'textbox');
             element.setAttribute('aria-multiline', true);
             element.setAttribute('medium-editor-index', index);
+        }, this);
 
-            if (element.hasAttribute('medium-editor-textarea-id')) {
```
j0k3r: I think you shouldn't remove this condition, it will avoid an extra listener when user doesn't use MediumEditor with textarea.

orthes: Yeah, I understand. But previously, ```input``` event was attached to element itself. But now ```editableInput``` is attached to the editor, not element. Maybe I could add a condition, if there is at least one element with textarea, then listen for ```editableInput```.

jesgundy:
For what it's worth, I also solved this problem on my local copy with a slightly different approach. Sharing here in case it's useful:

```js
function initElements() {
    var self = this;

    this.elements.forEach(function (element, index) {
        if (!this.options.disableEditing && !element.getAttribute('data-disable-editing')) {
            element.setAttribute('contentEditable', true);
            element.setAttribute('spellcheck', this.options.spellcheck);
        }
        element.setAttribute('data-medium-editor-element', true);
        element.setAttribute('role', 'textbox');
        element.setAttribute('aria-multiline', true);
        element.setAttribute('medium-editor-index', index);

        if (element.hasAttribute('medium-editor-textarea-id')) {
            this.subscribe('editableInput', function () {
              var textarea = element.parentNode.querySelector('textarea[medium-editor-textarea-id="' + element.getAttribute('medium-editor-textarea-id') + '"]');
                if (textarea) {
                    textarea.value = self.serialize()[element.id].value;
                }
            });
        }
    }, this);
}
```

Tested on IE 9, 10, Edge

j0k3r:
Hum interesting @jesgundy.
I was wondering if solution from @orthes will work on multiple textarea.
Your solution work on that case.

jesgundy:
True - I did test w/ more than one editor instance on the page and it seemed to work fine. 

¯\_(ツ)_/¯

j0k3r:
You tested with your solution or with the current version in the master?

jesgundy:
I tested that patch with the `dist/` version 5.10.0.

orthes:
My solution works with multiple textareas too. The only difference is that in my case, the event subscription is done only once and that is enough, because it is done on the editor itself, not on each element separately. ```this``` in ```this.subscribe``` refers to the editor.

j0k3r:
Thanks for that precision @orthes :+1:

jesgundy:
:+1:

PR #902, opened 2015-12-04
WIP - removeAllRanges IE 11 bug

pnispel:
IE 11.0.96 on Windows 8

I'm getting a error with the library when calling `selectElement`. 

I stumbled across [this issue](https://github.com/Voog/wysihtml/issues/231) while trying to fix it and it ended up fixing it for my purposes.


nmielnik:
There are a few other calls to `removeAllRanges()` in other places in the code.  Should we replace these as well?  If so, can we just make a `selection.replaceRange(selection, range)` helper function that we can use in other places?   I would think this helper method would just do the check that you included in here.

Also, looks like there's a spacing issue with your change, which is causing the build to fail.

src/js/selection.js
```diff
@@ -375,7 +375,11 @@
                 sel = doc.getSelection();
 
             range.selectNodeContents(node);
-            sel.removeAllRanges();
+            
+            if (sel.getRangeAt(0).getClientRects.length > 0) {
+              sel.removeAllRanges();
```
j0k3r: Could you fix indentation here?

pnispel:
Moving this to WIP since the changes caused tests to fail. 

@nmielnik is that what you had in mind for switching the other `removeAllRanges` calls and the helper method?

nmielnik:
@pnispel yeah, that helper is along the lines of what I was thinking of.  I'm not sure if we should use this helper method in all of our tests though, maybe we can just leave the tests as-is?

1. There is an additional reference in `util.js` [here](https://github.com/yabwe/medium-editor/blob/master/src/js/util.js#L453)
2. Could you remove all the dist files from your PR?  We generate new dist files on each release, so our general practice is never include them along with code changes: [CONTRIBUTION.md](https://github.com/yabwe/medium-editor/blob/master/CONTRIBUTING.md)
3. If we are ok with leaving the test code as-is (without using the helper), the helper method could actually be `replaceRange(selection, range)`.  I suggest this, because if you see all the references to `removeAllRanges()` in `selection.js` and `util.js`, we only call `selection.removeAllRanges()` when we want to immediately call `selection.addRange(range)`.  Perhaps a helper method for replacing the range would be more helpful and would be easier to re-use?

pnispel:
@nmielnik I've updated to reflect that. It looks like theres one place that you dont `addRange` immediately after, so I kept the `removeAllRanges` method in there. https://github.com/yabwe/medium-editor/blob/master/src/js/selection.js#L383

src/js/selection.js
```diff
@@ -375,12 +374,11 @@
...
 
             range.selectNodeContents(node);
-            sel.removeAllRanges();
-            sel.addRange(range);
+            MediumEditor.util.replaceRange(sel, range);
         },
 
         select: function (doc, startNode, startOffset, endNode, endOffset) {
-            doc.getSelection().removeAllRanges();
+            MediumEditor.util.removeAllRanges(doc.getSelection());
```
nmielnik: It's safe to get rid of this line and just call `MediumEditor.util.removeAllRanges()` on line 389.  That would allow us to not have 2 new helpers exposed (the reason why I like to try to avoid adding to many exposed helpers is that we can't really change the signature or the behavior of those without doing a major release, so the less we expose externally the better).

nmielnik:
@pnispel sorry for the run around, we really appreciate you finding out about this issue and jumping through all these hoops to get it into the code!

It looks like there are a bunch of [test failures](https://travis-ci.org/yabwe/medium-editor/builds/94969611), are you able to see those locally when you run `grunt`?

pnispel:
Is that (last commit) what you were thinking @nmielnik? I'm looking into the test failures, but its not really clear why they're failing - I dont know a ton about the selection API or the medium-editor code.

nmielnik:
@pnispel so the changes look great, thanks a ton.

It seems like these failures are indicative of there being something wrong with us doing this new check before we call `removeAllRanges()`.  I've tried to look into this a bit, it's not clear exactly why the tests would now be failing.

What I've been able to find is that the changes in `selection.js::129` and in `selection.js::383-388` are the changes that cause the failure.  The other changes in here work fine and don't cause tests to fail on their own.

src/js/util.js
```diff
@@ -965,6 +964,18 @@
             } else {
                 el.parentNode.removeChild(el);
             }
+        },
+
+        // https://github.com/Voog/wysihtml/issues/231
+        // fixes IE 11 bug with removeAllRanges
+        replaceRange: function (sel, replacement) {
+            if (sel.rangeCount > 0 &&
+                sel.getRangeAt(0).getClientRects.length < 0) {
```
nmielnik: As per the link you mentioned here, I believe this should be `sel.getRangeAt(0).getClientRects.length > 0` no?

nmielnik:
I've tried and wasn't able to fix this in a way to prevent a ton of test failures, so closing this.

PR #904, opened 2015-12-05
Fix issue with image-only selections

nmielnik:
Currently, medium-editor does not sufficiently handle exporting/importing selections that contain only an image (it's kind of supported in some browsers, but completely busted in IE). Images are a specific problem because:

1. `` elements are not text-nodes _(the code only handled selections that span 1 or more text nodes)_
2. `` elements are elements that will never have children _(the code ignored any element without children)_
3. `` elements have empty textContent _(the code depends on content having length so it can store the beginning and end of a selection)_

medium-editor handles exporting/importing selections based on the number of characters and paragraphs are before and after the selection.  So, if the selection contains only an image, when attempting to restore the selection the code ends up with a collapse selection.

The fixes here allow for wrapping an image in a link, and changing the url of a link on just an image.  This behavior was partially introduce by #803 but the behavior was broken in phantom (preventing extensive tests from being added) and was also broken in IE11 (causing test failures for the entire project).

By fixing this, this fixes the test failures in IE11.

### DIFF
--- a/src/js/selection.js
+++ b/src/js/selection.js
@@ -55,7 +55,7 @@
                     end: start + range.toString().length
                 };
 
-                if (selectionState.start === selectionState.end && MediumEditor.selection.selectionContainsContent(doc)) {
+                if (selectionState.start === selectionState.end && this.selectionContainsContent(doc)) {
                     selectionState.emptyContentSelection = true;
                 }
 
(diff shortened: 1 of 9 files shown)


src/js/selection.js
```diff
@@ -89,26 +98,55 @@
...
-                        stop = true;
-                    }
-                    charIndex = nextCharIndex;
-                } else {
-                    var i = node.childNodes.length - 1;
-                    while (i >= 0) {
-                        nodeStack.push(node.childNodes[i]);
-                        i -= 1;
+                // Only iterate over elements and text nodes
+                if (node.nodeType <= 3) {
```
j0k3r: I'm not a big fan of _big_ `if` like that.
Can't we do something like:

```js
if (node.nodeType > 3) {
    continue;
}
```

j0k3r:
Big hug 🤗 for the work on fixing failing test !! :+1: :rocket:

src/js/selection.js
nmielnik: doh, I'm not a big fan of it either. Fixing it now

l2aelba:
This still bug

j0k3r:
@l2aelba could you be more clear and give more information in a separate issue?

l2aelba:
Same as https://github.com/yabwe/medium-editor/issues/803 @j0k3r

PR #909, opened 2015-12-15
Trim src when using the image toolbar button

whatthewhat:
This PR trims trailing whitespace and newlines from image urls before inserting them into an `` tag.

**The issue this PR fixes**:
If the src has trailing newlines (it's pretty easy to accidentally select an empty line after the url) browsers display the image just fine, but most email clients do not (i.e. gmail) if the content is used in an email later.

### DIFF
--- a/spec/buttons.spec.js
+++ b/spec/buttons.spec.js
@@ -736,13 +736,13 @@ describe('Buttons TestCase', function () {
                 button = toolbar.getToolbarElement().querySelector('[data-action="image"]');
             spyOn(document, 'execCommand').and.callThrough();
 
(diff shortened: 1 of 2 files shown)


j0k3r:
Could you update test case for that ?
I think you just need to put some extra space inside the span `this.el.innerHTML = '<span id="span-image">http://i.imgur.com/twlXfUq.jpg</span>';` right?

whatthewhat:
@JoTrdl good point! updated

j0k3r:
Thanks :+1:

whatthewhat:
Thanks for merging! :)

PR #913, opened 2015-12-20
Update README.md

xethorn:
I'm not 100% sure how to describe this issue, so here's a first draft. Issue:

```
<div class="editor">
<p>Some texte</p>
</div>

.editor p {
    min-height: 12px;
}
```

Renders into:
screen shot 2015-12-20 at 10 17 26 am

Because of this: it's not possible anymore to go from one paragraph to another.


j0k3r:
Well, you shouldn't push a PR which put the issue in the README file :)
Instead, just create an issue describing the problem and we might find some time to fix it.

I'll close this PR and could you open a new issue with all information you have (include MediumEditor version, browser version and a jsfiddle to reproduce the error).

Thanks !

PR #930, opened 2016-01-15
stickyTopOffset option added

vied12:
I needed this option in order to keep on the screen a sticky toolbar when I have a static page header.


j0k3r:
Interesting. Can you provide a screenshot before / after of your toolbar? I'm just curious to see the difference.

vied12:
## Before
before
## After
after

j0k3r:
Ok thanks !

vied12:
You are welcome :bowtie:

PR #931, opened 2016-01-15
break out of infinite loop in importSelection

kbomb:
It currently hangs on non-text/element nodes, viz. comments.


j0k3r:
Thanks for that. Could you update the non-dist file instead?
Also could you add a test that fail without your changes?

kbomb:
Thanks for the feedback.  Going to close this PR because my colleague has her own working branch with additional fixes:
https://github.com/rachelsison/medium-editor/tree/text-node-loop-fix

j0k3r:
Well, maybe @rachelsison can submit a fix for that too :)

Issue #138, opened 2014-01-31
Misplacement of buttons when selection is near to the right edge


jonathanmv:
If the browser window is narrow and the selection occurs near to the right side, the buttons get misplaced as shown in the screenshot
![image](https://f.cloud.github.com/assets/86024/2052962/45aed2c0-8a9f-11e3-80f0-99dbf4973fb6.png)

This doesn't happen on the left side
![image](https://f.cloud.github.com/assets/86024/2052976/86196424-8a9f-11e3-9fcd-5743ed2d2b76.png)

nmielnik:
This is now fixed in the code itself, it will avoid wrapping the buttons unless the viewport itself is not wide enough to fit the toolbar.

PR #933, opened 2016-01-17
fixed issue 138

LeonRyan:
Issue https://github.com/yabwe/medium-editor/issues/138 is closed but is still occurring.
Fixed it by positioning the toolbar and anchor preview on the rhs when it doesn't fit.


nmielnik:
Thanks, nice fix!

PR #937, opened 2016-01-20
Fix #935 - Fix a variety of issues related to selection imports/exports

### LINKED ISSUE #935: Exception is thrown when I want to create a link from the first word of the first paragraph
Whenever I select the first word of the first paragraph and I want to turn it to a link nothing happens and instead I receive a JS exception.

This is how the editable is instantiated:
```
    $elem.editMeeteorWay({toolbar: {buttons: ['bold', 'italic', 'underline', 'anchor']},
                          autoLink: true,
                          anchorPreview: false,
                          extensions: {
                              'new-anchor-preview': new NewAnchorPreview(),
                              'edit-anchor-on-click' : new EditAnchorOnClick()
                          }
                        });
```

This is the markup:
```
<div class="editable" data-editor-options="{&quot;placeholder&quot;:{&quot;text&quot;:&quot;Add an agenda item by typing here.&quot;}}" data-insert-path="[&quot;agenda_item&quot;,&quot;agenda_item_text&quot;]" data-params="{&quot;agenda_item&quot;:{&quot;user_id&quot;:2,&quot;meeting_id&quot;:995},&quot;project_id&quot;:64,&quot;object_id&quot;:70119390241700}" data-url="/agenda_items/2405" contenteditable="true" spellcheck="true" data-medium-editor-element="true" role="textbox" aria-multiline="true" medium-editor-index="0" data-placeholder="Add an agenda item by typing here.">
      <p>one two<br><a href="transindex.hu">three</a><br></p><p><a href="amazon.com">one</a> two three</p>
</div>
```
This is the js exception trace:
Uncaught TypeError: Cannot read property 'toLowerCase' of undefinedUtil.getFirstSelectableLeafNode @ medium-editor.js?body=1:1258Selection.importSelectionMoveCursorPastBlocks @ medium-editor.js?body=1:1821Selection.importSelection @ medium-editor.js?body=1:1732MediumEditor.importSelection @ medium-editor.js?body=1:6474MediumEditor.restoreSelection @ medium-editor.js?body=1:6478MediumEditor.extensions.form.extend.completeFormSave @ medium-editor.js?body=1:3359MediumEditor.extensions.form.extend.doFormSave @ medium-editor.js?body=1:3355MediumEditor.extensions.form.extend.handleTextboxKeyup @ medium-editor.js?body=1:3423

nmielnik:
While investigation #935, I found a variety of issues:

* If a selection began in the first `<p>` after some lingering white space, exporting and then importing the selection was busted, as in the selection would not be restored correctly. The result was #935, where trying to create a link within that `<p>` caused an exception to be thrown.
* Previously, `util.getClosestBlockContainer()` and `util. getTopBlockContainer()` would return `false` if the passed in parameter was a node without any block containers other than the editor element.  Thus, the code could return `false` when it really should have been returning the actual editor element (since the editor element is the 'block container' that contains everything).  Our code expected there to be an element, an instead was getting `false`, which was the cause of the previous exception.

After fixing the issue and adding some better tests, more issues popped up:
* When importing a selection, the code should always try to have the range start at the beginning of an element rather than the end of an element.  It wasn't doing this before.
* Importing/exporting selection was broken when the selection began with an `` tag.
* Importing/exporting a collapsed selection at the end of the editor was not working and wasn't be tested.

So after these fixes, in cases where a selection starts in between elements, importing/exporting a selection will always try to make the selection start at the beginning of an element instead of always having it start at the end of an element.  Example:

```html
<p>This is <i>italic text </i><b>and bold text</b></p>
```
If you select the word 'and' inside of the `<b>` tag, and then exported/import selection, the previous code would have the selection start at the end of the 'italic text ' inside of the `<i>` tag.  From a visual standpoint, this is equivalent to having the selection start at the beginning of the `<b>` tag, but from a programatic standpoint there are definite advantages to having the selection be enclosed within the `<b>` tag, not to mention that this is almost always what the user intended.

The code now will have the selection be restored at start at the beginning of the `<b>` tag.  The edge cases where we would want to have the selection start at the end of the `<i>` tag are now treated as edge cases and handle correctly, instead of having all cases behave that way.

### DIFF
--- a/src/js/selection.js
+++ b/src/js/selection.js
@@ -114,8 +114,13 @@
                 if (node.nodeType === 3 && !foundEnd) {
                     nextCharIndex = charIndex + node.length;
                     if (!foundStart && selectionState.start >= charIndex && selectionState.start <= nextCharIndex) {
-                        range.setStart(node, selectionState.start - charIndex);
-                        foundStart = true;
+                        if (node.parentNode === root && selectionState.start === nextCharIndex && node.nextSibling) {
+                            range.setStart(node.nextSibling, 0);
+                            foundStart = true;
+                        } else {
+                            range.setStart(node, selectionState.start - charIndex);
+                            foundStart = true;
+                        }
                     }
                     if (foundStart && selectionState.end >= charIndex && selectionState.end <= nextCharIndex) {
                         if (!selectionState.trailingImageCount) {
@@ -273,8 +278,21 @@
             if (node.nodeType !== 3) {
                 node = cursorContainer.childNodes[cursorOffset];
             }
-            if (node && !MediumEditor.util.isElementAtBeginningOfBlock(node)) {
-                return -1;
+            if (node) {
+                // The element isn't at the beginning of a block, so it has content before it
+                if (!MediumEditor.util.isElementAtBeginningOfBlock(node)) {
+                    return -1;
+                }
+
+                var previousSibling = MediumEditor.util.findPreviousSibling(node);
+                // If there is no previous sibling, this is the first text element in the editor
+                if (!previousSibling) {
+                    return -1;
+                }
+                // If the previous sibling has text, then there are no empty blocks before this
+                else if (previousSibling.nodeValue) {
+                    return -1;
+                }
             }
 
             // Walk over block elements, counting number of empty blocks between last piece of text
--- a/src/js/selection.js
+++ b/src/js/selection.js
@@ -113,8 +113,12 @@
                 // If we hit a text node, we need to add the amount of characters to the overall count
                 if (node.nodeType === 3 && !foundEnd) {
                     nextCharIndex = charIndex + node.length;
+                    // Check if we're at or beyond the start of the selection we're importing
                     if (!foundStart && selectionState.start >= charIndex && selectionState.start <= nextCharIndex) {
-                        if (node.parentNode === root && selectionState.start === nextCharIndex && node.nextSibling) {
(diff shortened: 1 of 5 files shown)


nmielnik:
Looks like there are some failures in various browsers, ading [WIP] prefix until I'm able to resolve those

nmielnik:
Bah, looks like there are failures in IE9 and Safari, putting this PR back into [WIP] and looking at it now.

src/js/selection.js
```diff
@@ -269,8 +319,21 @@
...
+                    return -1;
+                }
+
+                var previousSibling = MediumEditor.util.findPreviousSibling(node);
+                // If there is no previous sibling, this is the first text element in the editor
+                if (!previousSibling) {
+                    return -1;
+                }
+                // If the previous sibling has text, then there are no empty blocks before this
+                else if (previousSibling.nodeValue) {
```
j0k3r: I wanted to comment about merging these two conditions into one, since there perform the same return, but finally, it's more clear with the comments to have them separated :)

j0k3r:
Looks good to me :+1: 
A big thanks for all the explanation, both in the PR and in the code comment, you rock! :dizzy:

nmielnik:
Thanks man :blush: 

The investigation took a while and I had to keep a lot of context in my head, if I hadn't been putting in lots of comments, I'd probably still be lost somewhere curled up in a ball wondering why I ever started looking into this.

PR #938, opened 2016-01-20
Fixed image insert event on file drag over

kaave:
Add `` tag after `filereader.readAsDataURL` event on image file dropped.
bacause img src is NOT DataURL format on editableInput fired.

### DIFF
--- a/spec/drag-and-drop.spec.js
+++ b/spec/drag-and-drop.spec.js
@@ -46,7 +46,7 @@ describe('Drag and Drop TestCase', function () {
             expect(editor.elements[0].className).not.toContain('medium-editor-dragover');
             // File API just doesn't work in IE9, so only verify this functionality if it's not IE9
             if (!isIE9()) {
-                expect(MediumEditor.util.insertHTMLCommand).toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


nmielnik:
@kaave I'm not sure if I understand what issue you're trying to fix here, could you give a bit more context?  

It looks as through there are some code style failures with your branch as well, could you take a look?

spec/drag-and-drop.spec.js
```diff
@@ -46,7 +46,7 @@ describe('Drag and Drop TestCase', function () {
             expect(editor.elements[0].className).not.toContain('medium-editor-dragover');
             // File API just doesn't work in IE9, so only verify this functionality if it's not IE9
             if (!isIE9()) {
-                expect(MediumEditor.util.insertHTMLCommand).toHaveBeenCalled();
+                // expect(MediumEditor.util.insertHTMLCommand).toHaveBeenCalled();
```
nmielnik: Why did you want to disable this part of the test case?

nmielnik:
Closing this and tracking any future fix as part of #985

PR #941, opened 2016-01-20
Fix for infinite loop AND gruntContribJasmine that errors out

rachelsison:
* The main bug I found was that the editor goes into an infinite loop when there is any `node.nodeType > 3` . In my particular use case I had some comments in the html (`node.nodeType ===  8`) and it just bounced back and forth between the continue statement and the conditional. My solution was to pop the node off the stack and continue with the traversal.

* The version of `grunt-contrib-jasmine` that is listed in the package.json is causing errors and the testing units don't run properly. References here for details on the errors and its solution: 

>https://github.com/gruntjs/grunt-contrib-jasmine/issues/193

* I added a test, that passes with my fix. It breaks without it, but does not break in a fun way. The testing suite goes into an infinite loop (caused by the bug) and doesn't continue executing the rest of the tests. I fiddled with it a little while trying to figure out how to make a 'proper' test for infinite loops but my Jasmine skills are not quite up to par here, apologies.


rachelsison:
I can also spit these two issues into separate PRs if that's better for you.

nmielnik:
@rachelsison I think this is great as-is, looks like I might have made a merge conflict for you with the PR I just merged in though, could you take a look?

Thanks a ton for finding this and fixing it, yet another problem I introduced that someone else was kind enough to fix :sweat_smile:

rachelsison:
@nmielnik it's good to go, thanks :)

src/js/selection.js
```diff
@@ -311,6 +312,7 @@
             while (!stop && node) {
                 // Only iterate over elements and text nodes
                 if (node.nodeType > 3) {
+                    node = nodeStack.pop;
```
j0k3r: Shouldn't this be `pop()` ?

rachelsison: Yes. Not sure how that happened.. must have accidentally backspaced when I was searching through my file before saving. Will fix now.

rachelsison:
apologies, not sure how that happened. syntax error fixed, thanks.

PR #950, opened 2016-01-31
handled enter event for empty h2/h3 tag 🔨

### LINKED ISSUE #924: event.preventDefault or disableDoubleReturn don't work when element is other than a 'p' tag
First and foremost, thank you for the excellent plugin!

I am getting issues when using 'disableDoubleReturn' or just a simple preventDefault on hitting Enter, when the node (line) is empty.

When the element is a simple 'p' tag, everything works as expected. No content = no new line created on "Return'. I need to write something inside the created line to be able to press Enter and create another one.

However, when I write something in the line, call up the toolbar and transform it into an 'h2' from a 'p', then delete everything in that 'h2' and press 'Return', the line returns as usually. Same for the 'h3'.

Is this an issue or wanted behaviour? If it's the latter, how would I go about overriding it? If it's the first, maybe someone can point me in the right direction to fix it?

Thank you in advance!

rishijain:
This fixes #924

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -434,6 +434,45 @@ describe('Content TestCase', function () {
 
             expect(evt.preventDefault).toHaveBeenCalled();
         });
+
+        it('should add a new line when selected node is an h2/h3 tag with text in it and when disableDoubleReturn is true', function () {
+            this.el.innerHTML = '<p>lorem</p><h2>ipsum<br></h2>';
+            var editor = this.newMediumEditor('.editor', { disableDoubleReturn: true }),
+                p = editor.elements[0].getElementsByTagName('h2')[0],
+                evt;
+
+            placeCursorInsideElement(p, 0);
+
+            evt = prepareEvent(p, 'keydown', {
+                keyCode: MediumEditor.util.keyCode.ENTER
+            });
+
+            spyOn(evt, 'preventDefault').and.callThrough();
+
+            firePreparedEvent(evt, p, 'keydown');
+
+            expect(evt.preventDefault).not.toHaveBeenCalled();
+        });
+
+        it('should prevent new line being added when selected node is an empty h2/h3 tag and when disableDoubleReturn is true', function () {
+            this.el.innerHTML = '<p>lorem</p><h2><br></h2>';
+            var editor = this.newMediumEditor('.editor', { disableDoubleReturn: true }),
+                p = editor.elements[0].getElementsByTagName('h2')[0],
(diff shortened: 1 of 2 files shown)


src/js/core.js
```diff
@@ -73,10 +73,12 @@
             } else if (MediumEditor.util.isKey(event, MediumEditor.util.keyCode.ENTER)) {
                 // hitting return in the begining of a header will create empty header elements before the current one
                 // instead, make "<p><br></p>" element, which are what happens if you hit return in an empty paragraph
-                p = this.options.ownerDocument.createElement('p');
-                p.innerHTML = '<br>';
-                node.previousElementSibling.parentNode.insertBefore(p, node);
-                event.preventDefault();
+                if (!this.options.disableDoubleReturn) {
```
nmielnik: Should this `!this.options.disableDoubleReturn` check be merged into the existiing outer-if?

ie:
```js
} else if (!this.options.disableDoubleReturn &&
              MediumEditor.util.isKey(event, MediumEditor.util.keyCode.ENTER)) {
    p = this.options.ownerDocument.createElement('p');
    // rest of code
}
              
```

spec/content.spec.js
```diff
@@ -434,6 +434,45 @@ describe('Content TestCase', function () {
...
+            });
+
+            spyOn(evt, 'preventDefault').and.callThrough();
+
+            firePreparedEvent(evt, p, 'keydown');
+
+            expect(evt.preventDefault).not.toHaveBeenCalled();
+        });
+
+        it('should prevent new line being added when selected node is an empty h2/h3 tag and when disableDoubleReturn is true', function () {
```
nmielnik: Could you give these test cases different names/descriptions in some way?  Being named exactly the same thing can be a bit confusing...

j0k3r: @rishijain was it fixed when you force pushed your commit or I'm missing sth?

rishijain: @j0k3r Yes. I fixed those changes asked by @nmielnik  and then rebased and force pushed the commit as I do not want a separate commit just for fixing the typo.

j0k3r: Great, no problem :)

rishijain: I could have merged this already but I am waiting for @nmielnik  to have a quick look before it goes in. :smile:

j0k3r: Same here :D

nmielnik:
:+1: looks good my man!

PR #954, opened 2016-02-03
Deprecate util.getFirstTextNode

nmielnik:
Looking at our code coverage numbers, I found a method that is not referenced by our code anymore (`MediumEditor.util.getFirstTextNode()`).  We should get rid of this function since it's not used, however since it changes the API we'll have to settle with just deprecating it for now and remember to remove it in 6.0.0.

Since it's going to be around, and could potentially get used by other code moving forward, we should have coverage of this function.  I've added a few test cases to have 100% coverage of the helper method.

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -940,7 +940,7 @@
             }
 
             for (var i = 0; i < element.childNodes.length; i++) {
-                var textNode = Util.getFirstTextNode(element.childNodes[i]);
+                var textNode = Util._getFirstTextNode(element.childNodes[i]);
                 if (textNode !== null) {
                     return textNode;
                 }
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -928,7 +928,13 @@
             return element;
         },
 
+        // TODO: remove getFirstTextNode AND _getFirstTextNode when jumping in 6.0.0 (no code references)
         getFirstTextNode: function (element) {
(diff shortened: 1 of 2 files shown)


src/js/util.js
```diff
@@ -928,7 +928,13 @@
             return element;
         },
 
+        // TODO: remove getFirstTextNode AND _getFirstTextNode when jumping in 6.0.0 (no code references)
         getFirstTextNode: function (element) {
+            Util.warn('getFirstTextNode is deprecated and will be removed in version 6.0.0');
+            return Util._getFirstTextNode(element);
+        },
+
+        _getFirstTextNode: function (element) {
```
j0k3r: Do you really need to create a new function? Why don't you just put the warning message in the existing function?

nmielnik: It's recursive, so each call to the method produces many many warning messages...which I found out when I ran the tests :smile:

j0k3r: Oh! Didn't notice :stuck_out_tongue_closed_eyes:

nmielnik: And I didn't even fix that problem...hah, let me fix that now :stuck_out_tongue_closed_eyes:

nmielnik: Ok, now the recursive code will call the version of the method that doesn't spit out the warning.

j0k3r: :+1:

Issue #771, opened 2015-08-10
Microsoft Edge Support


nmielnik:
We need to:

- [x] Fix tests which are broken in Microsoft Edge browser on Windows 10 (~48 failures as of now)
- [x] Sanity check to make sure things look fine and working as expected in Edge
- [x] Add IE 11 + Windows 10 support into our SauceLabs test runs (Edge isn't supported yet)
- [x] Updated our list of supported browsers (if there's anywhere where we call this out)

#955 includes all the work needed for us to 'support' Edge, we'll cut a new minor release once that work is good to go.

j0k3r:
Is there any way to do 3. stuff in SauceLabs in a way that it doesn't mark the test as fail in case of failing test on Edge ? I don't know enough SauceLabs but with Travis you can define some case where you allow test to fail ([see here](http://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail)).

nmielnik:
I don't think there is a way with our current setup, but I'm not completely sure.  Looking through SauceLabs, it looks like they don't support Microsoft Edge yet, but they have lots of Windows 10 combinations so we might want to just start running the tests in Chrome, Firefox, and IE11 on Windows 10 to start with.

jdalton:
SauceLabs supports Edge 12 at the moment.

I noticed Edge 12-13 still don't support the Input event from `document.execCommand` so the sniff will need to be adjusted. Instead of a sniff though I have an inference that works well.

```js
document.execCommand = (function(_execCommand) {
  var isBuggy = true;

  function complete(element, result) {
    if (result) {
      var event = document.createEvent('Event');
      event.initEvent('input', true, false);
      element.dispatchEvent(event);
    }
    return result;
  }

  function getEditable(node) {
    while (node &&
        !(node.nodeType == 1 && node.hasAttribute('contenteditable'))) {
      node = node.parentNode;
    }
    return node;
  }

  function restore() {
    isBuggy = false;
    document.execCommand = _execCommand;
  }

  function exec(cmd, ui, arg) {
    return _execCommand.call(document, cmd, ui, arg);
  }

  function shimExec(selection, cmd, ui, arg) {
    return complete(editable, exec(cmd, ui, arg));
  }

  function shim(selection, cmd, ui, arg) {
    var editable = getEditable(selection.anchorNode);
    editable.addEventListener('input', restore, true);

    var result = exec(cmd, ui, arg);
[...]

PR #955, opened 2016-02-07
Edge Support: Fixes to medium-editor and test cases for Edge

nmielnik:
Fixes for #771
* Fixed 51 failed tests!
* All tests are now run on Edge in Windows 10 (via SauceLabs)

Changes & Stuff I've learned:
* Enabled Edge browser testing in SauceLabs!
* For `<a>` tags, Edge seems to add a protocol to the beginning of urls within the `href` attribute, even when one isn't specified.  Had to update the tests to account for this.
* Edge seems to handle 'indent' for `<blockquote>` tags the same way Firefox does, so expand special handling to include Edge (which means browser sniffing :disappointed:)
* Edge's implementation of `insertHTML` for `document.execCommand()` appears to have some bugs.  The first big part of this is to use our existing fallback for `insertHTML` even though Edge says it supports it.
* For our test cases in Edge, manually triggering a 'focus' event on the editor doesn't work as a consistent way to make the toolbar inspect the selection and update itself.  In order to get the tests to run consistently, we'll have to manually trigger a `click` event which will update the toolbar in all browsers.
* Edge's support for the `input` event on contenteditable isn't all the way there.  It triggers correctly when the user is editing and on cut/paste (which IE didn't), but it doesn't trigger when calling undo/redo.  There might be more cases, but for now I just gave up and fell back to the shim'd `input` event logic we used for IE.

### DIFF
--- a/spec/helpers/util.js
+++ b/spec/helpers/util.js
@@ -60,6 +60,10 @@ function isIE() {
     return ((navigator.appName === 'Microsoft Internet Explorer') || ((navigator.appName === 'Netscape') && (new RegExp('Trident/.*rv:([0-9]{1,}[.0-9]{0,})').exec(navigator.userAgent) !== null)));
 }
 
+function isEdge() {
+    return new RegExp('Edge/\\d+').exec(navigator.userAgent) !== null;
+}
+
 function isFirefox() {
     return navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
 }
--- a/spec/helpers/util.js
+++ b/spec/helpers/util.js
@@ -60,6 +60,14 @@ function isIE() {
     return ((navigator.appName === 'Microsoft Internet Explorer') || ((navigator.appName === 'Netscape') && (new RegExp('Trident/.*rv:([0-9]{1,}[.0-9]{0,})').exec(navigator.userAgent) !== null)));
 }
 
+function isEdge() {
+    var match = /Edge\/(\d+)/.exec(navigator.userAgent);
+    if (match !== null) {
+        return new Number(match[1]);
+    }
+    return 0;
+}
+
 function isFirefox() {
     return navigator.userAgent.toLowerCase().indexOf('firefox') !== -1;
 }
--- a/spec/helpers/util.js
+++ b/spec/helpers/util.js
@@ -220,5 +220,5 @@ function selectElementContents(el, options) {
 function selectElementContentsAndFire(el, options) {
     options = options || {};
     selectElementContents(el, options);
-    fireEvent(el, options.eventToFire || 'focus');
+    fireEvent(el, options.eventToFire || 'click');
 }
--- a/spec/helpers/util.js
+++ b/spec/helpers/util.js
@@ -60,10 +60,11 @@ function isIE() {
     return ((navigator.appName === 'Microsoft Internet Explorer') || ((navigator.appName === 'Netscape') && (new RegExp('Trident/.*rv:([0-9]{1,}[.0-9]{0,})').exec(navigator.userAgent) !== null)));
 }
 
+// If the browser is Edge, returns the version number as a float, otherwise returns 0
 function getEdgeVersion() {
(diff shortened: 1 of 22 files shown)


j0k3r:
:tada: :clap:

spec/helpers/util.js
```diff
@@ -60,6 +60,14 @@ function isIE() {
     return ((navigator.appName === 'Microsoft Internet Explorer') || ((navigator.appName === 'Netscape') && (new RegExp('Trident/.*rv:([0-9]{1,}[.0-9]{0,})').exec(navigator.userAgent) !== null)));
 }
 
+function isEdge() {
```
j0k3r: Shouldn't this be `getEdgeVersion()` instead ?

nmielnik: yeah, it can be.  I was trying to follow the naming convention, and had this return `0` if it wasn't edge, so that calls to `isEdge()` would work as you would expect, but the bonus is that you can also find out the exact version too.

If you think we should, I'm fine with renaming it.

j0k3r: For now, you only use it to retrieve the version number and never to actually test if the current browser is Edge. Am I wrong?

nmielnik: No, it's not used anywhere else.  I'll change the name.

spec/helpers/util.js
```diff
@@ -60,6 +60,14 @@ function isIE() {
     return ((navigator.appName === 'Microsoft Internet Explorer') || ((navigator.appName === 'Netscape') && (new RegExp('Trident/.*rv:([0-9]{1,}[.0-9]{0,})').exec(navigator.userAgent) !== null)));
 }
 
+function getEdgeVersion() {
+    var match = /Edge\/(\d+)/.exec(navigator.userAgent);
+    if (match !== null) {
+        return new Number(match[1]);
```
jdalton: Why return a number object instead of primitive?

src/js/util.js
```diff
@@ -437,7 +439,14 @@
...
             var selection, range, el, fragment, node, lastNode, toReplace;
 
-            if (doc.queryCommandSupported('insertHTML')) {
+            /* Edge's implementation of insertHTML is just buggy right now:
+             * - Doesn't allow leading white space at the beginning of an element
+             * - Found a case when a <font size="2"> tag was inserted when calling alignCenter inside a blockquote
+             *
+             * There are likely many other bugs, these are just the ones we found so far.
+             * For now, let's just use the same fallback we did for IE
+             */
```
jdalton: I'll report the `insertHTML` issue too.

nmielnik: Awesome, I'll work on pointing out the exact test cases that were failing so you can have some clear repro steps.

spec/helpers/util.js
nmielnik: This was an oldschool js trick I used to use to convert a decimal number within a string into a primitive number (ie `'12.123'` into `12.123`). Definitely open to a better way to do this.

jdalton: In this case its converting it to a Number object (so typeof result is "object") wasn't sure if that was intentional. The result here though will only ever be a whole number, like "12" or "13", because the `\d+` just matches `[0-9]`. If you want to convert that to a number primitive you can just do
```js
return +match[1]
```

nmielnik: Ugh, I definitely used the wrong regex here (I wanted the full decimal), and the `+` type casting is definitely much cleaner.  Looks like `+` is even slightly better than `parseFloat()` beyond being more compact, since it'll return 0 if the string was empty instead of `NaN`.  Thanks for the tip.

src/js/util.js
nmielnik: Test cases and some details on the issues I ran into with `insertHTML`: https://github.com/yabwe/medium-editor/issues/771#issuecomment-182233090

spec/helpers/util.js
jdalton:
```js
/Edge\/(\d+(?:\.\d+)?)/
```

PR #957, opened 2016-02-09
Fix some bad image references in tests

nmielnik:
Some of our tests had broken references to images, this should be fixed now (and there shouldn't be any 404s when running tests in a browser).


rishijain:
:+1:

PR #965, opened 2016-02-10
Don't add fileDragging extension is no file type is allowed. fixes #966

### LINKED ISSUE #966: FileDragging Extension is always on even without allowed files types
http://codepen.io/poda/pen/XXoKeE

in this example, we added `imageDragging` to be `false` which causes the `allowedTypes` to be empty for the file dragging extension. No file can be dragged onto the editting area, however the dragNdrop events are still captured and related styles are being applied. 

This is especially problematic, when you want its parent to handle DragNDrop events.

goldensunliu:


goldensunliu:
fixing.

spec/drag-and-drop.spec.js
```diff
@@ -19,14 +19,6 @@ describe('Drag and Drop TestCase', function () {
             expect(editor.elements[0].className).toContain('medium-editor-dragover');
         });
 
-        it('should add medium-editor-dragover class even when data is invalid', function () {
```
goldensunliu: @nmielnik what was the thoughts behind this behavior at the first place? 
wouldn't be better to not do drag and drop at all when no files type is allowed? https://github.com/yabwe/medium-editor/issues/966

nmielnik: There are 2 separate concepts with the drag-and-drop.

1) Without the extension, drag and drop will cause the browser to open the file by navigating to the file or opening the file in a separate tab.  When the extension is instantiated, it listens for the drop event over the editor and prevents the default action, which would be having the browser open the file.
2) The `allowedType` specifies that for a certain file type, the code should have some special handling for inserting the file into the editor.  Currently, we only have code that handles images, so `'image'` is the only meaningful value to pass into this array.

So, `allowedTypes` should not affect whether the extension exists or not, since it is intended for a different purpose (the parameter is a bit confusing and probably named incorrectly, that's my fault and I hope to fix it for 6.0.0).

nmielnik:
In 6.0.0, we plan to remove the `imageDragging` option completely, and instead define a `fileDragging` option which will allow you to turn the extension on/off, as well as specify which type of files you want to allow custom logic to insert (if the extension is on).

Until then, the test cases are still valid, and since the `allowedTypes` property of the extension is intended for a different purpose, I don't think it should affect whether the extension is instantiated or not.

goldensunliu:
sounds good! I will remove this for now

Issue #939, opened 2016-01-20
Unchecking the "Open in new window" checkbox does nothing in Firefox


lydell:
Steps to reproduce in Firefox:

1. Hover over a link with the `target="_blank"` attribute (either add it in the
   HTML source code, or add it by using MediumEditor) and click the URL popup
   that appears to edit the link. The link I used is very simple: It contains
   just one single, unformatted word.

2. In the URL edit form that opens, uncheck the "Open in new window" (make sure
   that it is checked).

3. Click the checkmark in the URL edit form to confirm the changes.

In Firefox (Developer Edition 45.0a2 on Ubuntu 15.04 in my case), the
`target="_blank"` attribute is still there. (I inspected the link with the
devtools to see it. You can also re-open the URL edit form and see that the
"Open link in new window" checkbox has been "re-checked"). In Chromium,
`target="_blank"` is removed as expected when unchecking the checkbox.

When running the debugger, I found that this line was executed: https://github.com/yabwe/medium-editor/blob/f7c522af3069546e602bdff1fe3d7768b89dd570/src/js/core.js#L1043

The code seems to rely on that running `document.execCommand('createLink',
false, url)` returns a fresh link without a `target` attribute. That's the case
in Chromium, but not in Firefox.

I made a reduced test case that only uses that line of code to demonstrate what
happens.

1. Open the following page:

   ```html
   <!doctype html>
   <title>contenteditable</title>
   <div contenteditable="true"><a href="/foo" target="_blank">test</a></div>
   ```

2. Select the entire link with the mouse (the word "test").

3. Run the following code in the console while the link is selected:

   ```js
   document.execCommand('createLink', false, '/bar')
   ```

4. Inspect the link with the devtools. Notice how the `href` attribute has
   changed from `/foo` to `/bar`, but the `target="_blank"` attribute remained.
   In Firefox, that is. When doing the same in Chromium, the `target` attribute
   vanishes (or, perhaps, isn't preserved, depending on how you look at it).

I suggest explicitly removing the `target` attribute to fix this problem.

nmielnik:
Thanks for pointing this out @lydell , if you have a chance to submit a fix for this we'd be glad to accept it, especially since you've been able to narrow down the code and describe an excellent test scenario.

Otherwise, hopefully I'll have some time to look into it or someone else might take a crack at fixing it :smile:

PR #967, opened 2016-02-11
fix for #939 🔨

rishijain:
Added a function `setTargetSelf` along the lines of `setTargetBlank`.

It is called whenever target is not suppose to be blank. The need to do this came up because of #939 . It worked well on all browsers except Firefox. So had to specifically remove target blank from the anchor tags. And one of the way I think to remove is to set it to `_self`.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -392,6 +392,35 @@ describe('Anchor Button TestCase', function () {
             expect(link.classList.contains('btn-default')).toBe(true);
         });
 
+        it('should remove the target _blank from the anchor tag when the open in a new window checkbox,' +
+                ' is unchecked and the form is saved', function () {
+            var editor = this.newMediumEditor('.editor', {
+                anchor: {
+                    targetCheckbox: true
+                }
+            }),
+                anchorExtension = editor.getExtensionByName('anchor'),
+                targetCheckbox,
+                link;
+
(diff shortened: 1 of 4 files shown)


rishijain:
Some tests are failing :cry: . So its a WIP right now.

spec/anchor.spec.js
```diff
@@ -68,7 +69,11 @@ describe('Anchor Button TestCase', function () {
             });
             expect(editor.createLink).toHaveBeenCalled();
             // A trailing <br> may be added when insertHTML is used to add the link internally.
-            expect(this.el.innerHTML.indexOf('<a href="http://test.com">lorem ipsum</a>')).toBe(0);
+            if (!isFirefox()) {
+                expect(this.el.innerHTML.indexOf('<a href="http://test.com" target="_self">lorem ipsum</a>')).toBe(0);
+            } else {
+                expect(this.el.innerHTML.indexOf('<a target="_self" href="http://test.com">lorem ipsum</a>')).toBe(0);
+            }
```
j0k3r: Seriously? :disappointed:

rishijain: Yes, firefox and chrome places target="_self" (or blank) at different positions. and thus need to add a check like this.

nmielnik:
:+1:

PR #971, opened 2016-02-11
add instruction to disable file dragging all together

goldensunliu:


nmielnik:
@goldensunliu thanks for doing this, it's a great idea to clarify in the documentation!

Could you potentially look into the wording of your documentation?  The File Dragging extension is the one that prevents drag & drop actions from actually executing anything.  If you disable the file dragging extension, you actually _ENABLE_ the drag & drop events to happen and be consumable...does that make sense?

goldensunliu:
@nmielnik I will use your exact words in my second paragraph, how does that sound?

nmielnik:
Actually, I thin what you have is fine, I think I misread it the first time.  Thanks!

Issue #972, opened 2016-02-12
Paste html does not work in Firefox (43.0.1)


daxhns:
In Firefox (43.0.1, windows ) when pasting html formatted text into the contenteditable div, all formatting is stripped out and only plain text is pasted, even if using paste extension, and disabling forcePlainText option.

			var comment_editor = new MediumEditor(js_element,  {
				buttonLabels: 'fontawesome',
				targetBlank: true,
				toolbar: {
					buttons: ['bold', 'italic', 'underline', 'quote', 'anchor', 'orderedlist', 'unorderedlist']
				},
				anchor: {
					linkValidation: true,
					placeholderText: 'Paste or type a link..',
				},
				paste: {
			        forcePlainText: false,
			        cleanPastedHTML: true,
			        cleanReplacements: [],
			        cleanAttrs: ['class', 'style', 'dir'],
			        cleanTags: ['meta', 'img']
			    },
				anchorPreview: false,
				autoLink: true,
				placeholder: false	
			});

Works fine on Chrome (formatted text is pasted).

j0k3r:
Yep this is a known bug (even if I can't find a related issue).

Firefox remove all html no matter how do you configure the paste extension. On the other hand it works great on Chrome.

I've started something 8 months ago to handle it like every other editor does:
- catch the paste event
- move cursor in a non `contenteditable` node (like an hidden `div` in the page) BEFORE the even is finalized
- the even is dispatched and the html content is in the non `contenteditable` node
- move this content to the MediumEditor node

But can't find the time to finalize it.

daxhns:
I have just found out that it actually works if you COPY html content FROM Firefox and than paste it in Firefox.

It does not work if you copy html content from Chrome, and then paste it in Firefox.

Madness!

j0k3r:
Yeah I ran into that too and it doesn't _always_ work :confused:

PR #973, opened 2016-02-12
Better Paste Handling

j0k3r:
Back on June 2015,  I started to work on a different way to handle HTML paste when using Firefox.

The main problem I found when you paste HTML in Firefox it removes every node and paste plain text no matter what paste configuration you have on MediumEditor.

I found that other editor handle that problem by cheating on the way Firefox handle contenteditable. When you paste HTML in a contenteditable, Firefox paste plain text only.

So the way to handle HTML paste in Firefox is:
- catch the paste event
- move cursor in a non `contenteditable` node (like an hidden `div` in the page) BEFORE the even is finalized
- the even is dispatched and the html content is in the non `contenteditable` node
- move this content to the MediumEditor node

Since the bug is still here (see https://github.com/yabwe/medium-editor/issues/972) I push my work on that part.

As far as I remember the first paste worked but all other one failed ...

### DIFF
--- a/src/js/extensions/paste.js
+++ b/src/js/extensions/paste.js
@@ -1,5 +1,12 @@
+/*global Util, Selection, Extension */
+
+'use strict';
+
+var PasteHandler,
+    keyboardPasteTimeStamp = 0,
+    pasteBinDefaultContent = '%ME_PASTEBIN%';
+
 (function () {
-    'use strict';
     /*jslint regexp: true*/
     /*
         jslint does not allow character negation, because the negation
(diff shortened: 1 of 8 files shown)


Dayjo:
Will this affect the usage of pasting from the context or edit menus (a surprising number of people do it this way).

j0k3r:
This first draft only handle the <kbd>CTRL</kbd>+<kbd>B</kbd> but yeah, we'll need to handle context menu

src/js/extensions/paste.js
```diff
@@ -86,48 +89,128 @@
...
+
+                // Grab plain text from Clipboard API or convert existing HTML to plain text
+                if (plainTextMode) {
+                    // Use plain text contents from Clipboard API unless the HTML contains paragraphs then
+                    // we should convert the HTML to plain text since works better when pasting HTML/Word contents as plain text
+                    if (that.hasContentType(clipboardContent, 'text/plain') && content.indexOf('</p>') === -1) {
+                        content = clipboardContent['text/plain'];
+                    } else {
+                        // What is innerText supposed to do? (It isn't defined)
+                        //content = MediumEditor.util.innerText(content);
```
j0k3r: Hum good point, I'll have a look on monday

j0k3r: I found nothing about that on my work dev machine. I guess it was to retrieve the html from `content`? I can't remember

j0k3r:
Huge thanks @nmielnik for finishing this stuff 👍 🚀

PR #974, opened 2016-02-13
Fixed missing `editableInput` event on autolinking

### LINKED ISSUE #959: Event `editableInput` doesn't fire on autoLink
when `autoLink` is true, on auto linking Event `editableInput` is never fired, are we expecting some other event to be fired?

edencore:
Fixes #959 

Is it correct?

### DIFF
--- a/spec/auto-link.spec.js
+++ b/spec/auto-link.spec.js
@@ -213,6 +213,20 @@ describe('Autolink', function () {
                 expect(links[0].textContent).toBe('http://www.example.blur');
             });
 
+            it('should fire editableInput after autolinking', function () {
+                var spy = jasmine.createSpy('handler');
+                this.el.innerHTML = 'http://www.example.com';
+
+                selectElementContentsAndFire(this.el);
+                this.editor.subscribe('editableInput', spy);
+                expect(spy).not.toHaveBeenCalled();
+
+                fireEvent(this.el, 'blur');
+                jasmine.clock().tick(1);
+
+                expect(spy).toHaveBeenCalled();
+            });
+
             it('should auto-link link within basic text', function () {
                 this.el.innerHTML = 'Text with http://www.example.com inside!';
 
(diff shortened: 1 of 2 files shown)


nmielnik:
Internally, whenever we could potentially have made changes to the html, we call `this.events.updateInput()`.  This method has a cache of the html so that it can inspect to see if anything actually changed (this way it avoids firing the `editableInput` event if the content hasn't actually changed).

In order to call the method, you need a reference to the `contenteditable` element, so you can either:
1. in `createAutoLink()`, use the reference to the anchor element, and loop up through its parent elements until you find a root contenteditable.  You can use `MediumEditor.util.isMediumEditorElement()` as a way to check if you've hit the root contenteditable.
2. You can move the call to the end of `performLinking()` method (instead of at the end of `createAutoLink()`).  The `performLinking()` method already has a reference to the editable element passed as a parameter.

Calling `updateInput()` would end up looking something like this:

```js
var editable = /* however you get a reference to the root contenteditable */;
this.base.events.updateInput(editable, { target: editable, currentTarget: editable });
```

Does that make sense?

edencore:
I really hope I got you correct this time. Feel a little bit ashamed for your wasted time on help with this simple issue

nmielnik:
@edencore no worries at all, `Events.updateInput()` is an internal method that's not really documented, so I wouldn't expect people to know about its existence or what it does :smile:

I was digging through the code a bit and I realized that the `performLinkingWithinElement()` method will never actually return `true`.  The method actually doesn't track that whether the document was modified now (yikes).

So, you can do 2 things here.  You can either remove the `if (documentModified)` clause from the line that you added, or you can try to fix the code so that `performLinkingWithinElement()` will return true when a link has been created.  

Since nothing really depends on whether `documentModified` is true or not, I would be fine with just removing that check from the code that you added, but it's up to you.

The last thing I would ask is, are you willing to add a test case for this (one to make sure that `editableInput` fires when auto-linking happens)?  You could just add another test to [`auto-link.spec.js`](https://github.com/yabwe/medium-editor/blob/master/spec/auto-link.spec.js), and if you're looking for examples of `editableInput` tests you can find them in [`events.spec.js`](https://github.com/yabwe/medium-editor/blob/master/spec/events.spec.js)

nmielnik:
@edencore thanks, this is great! 

Apologies for misleading you with the original issue (I hadn't thought it through completely).

edencore:
@nmielnik great! No worries, you helped me a lot and this was one of the most completely explained issues in my dev life :)

PR #981, opened 2016-02-17
ImageDragging insert image at the end of dropping target

### LINKED ISSUE #980: ImageDragging should drop file at the end of dropping target
Currently, on drag and drop of an image, the content is inserted on the editor's selection. example:
This produces two undesired behavior:

### 1. when there is an active selection in the editor, dropping a file will replace the selection. 
example:

### 2. current logic highlights the dropping target and make users believe that the image will be inserted behind where it is highlighted, when it is not. 
example:

### Ideally, the behavior should be on drop of an image, it is then inserted at the end of the element it was dropped upon
example:

goldensunliu:
fixes #980

### DIFF
--- a/src/js/extensions/file-dragging.js
+++ b/src/js/extensions/file-dragging.js
@@ -42,7 +42,12 @@
             // Prevent file from opening in the current window
             event.preventDefault();
             event.stopPropagation();
-
+            // Select the dropping target, and set the selection to the end of the target
+            // https://github.com/yabwe/medium-editor/issues/980
+            this.base.selectElement(event.target);
+            var selection = this.base.exportSelection();
+            selection.start = selection.end;
+            this.base.importSelection(selection);
             // IE9 does not support the File API, so prevent file from opening in the window
             // but also don't try to actually get the file
             if (event.dataTransfer.files) {


nmielnik:
This makes sense and is a step in the right direction, thanks @goldensunliu ! 

I think it'd be nice to try to take it a step further and actually try inserting the image at the exact location of the drop (not just the end of the element).  It's a bit more challenging to pull off, but would be an even better experience.

goldensunliu:
@nmielnik yeah, I would need to read the source code a bit more. I am fuzzy around how to caculate the exact offset I need for the selection. Do you know an similar case that I can take a look and gain some understanding? I will read through the code eventual if I need to.

nmielnik:
Yeah, since the content can consist of giant text nodes, you would need to find some way to map the x-y coordinates of the mouse to the exact point within the text.  I'm not sure how to do this (or if it's even possible) but I don't think there's anything in the medium-editor code that would help with this problem.  Might take some googling and some experimentation.

Also, after looking at #984, seems like you can just just select the `event.target` and then call `selection.collapseToEnd()` and avoid those extra `exportSelection()` and `importSelection()` calls in the file-dragging extension.  Not a big deal, but might be and optimization worth changing in the future.

Issue #983, opened 2016-02-18
Add ClearSelection method.


CupOfTea696:
Since there are various methods to set a selection, it'd be nice to have a method that clears the current selection and sets the cursor at the selection start / end. Maybe use a parameter to set the cursor at the selection start or end.

j0k3r:
Feel free to update `selection.js` :)

nmielnik:
Resolve by #1018

PR #984, opened 2016-02-18
core.clearSelection

CupOfTea696:
PR for #983

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -873,6 +873,16 @@
             }
         },
 
+        clearSelection: function (toEnd) {
+            var selection = this.options.ownerDocument.getSelection();
+
+            if (toEnd) {
+                return selection.collapseToEnd();
+            }
+
+            selection.collapseToStart();
+        },
+
         getFocusedElement: function () {
             var focused;
             this.elements.some(function (element) {


j0k3r:
Thanks !
Could you add some tests?

nmielnik:
@CupOfTea696 We have been putting our selection helpers methods in `selection.js` (accessible through `MediumEditor.selection`).  Was there a reason you wanted this to be on the core code instead?

I ask because exposing methods on the core-api makes them require a major release if we ever modify them, and it's good to reduce the surface area of those kind of methods if possible.

src/js/core.js
```diff
@@ -873,6 +873,16 @@
             }
         },
 
+        clearSelection: function (toEnd) {
```
nmielnik: Any reason not to call this `collapseSelection()` (to use an already existing, well-defined term)?

CupOfTea696:
@nmielnik If I put it in selection, how would I access it? From what I can see there is no way of accessing the `Selection` instance from a `MediumEditor` instance.

src/js/core.js
CupOfTea696: Because its implementation is different? `Selection.collapse` expects `parentNode` and `offset` parameters, whilst this method is a shortcut / switch between `collapseToStart` and `collapseToEnd`

j0k3r:
@CupOfTea696 You can access it using `MediumEditor.selection. clearSelection`

src/js/core.js
orthes: It's true that this function is only a wrapper around `collapseToStart()` and `collapseToEnd()` functions, so @nmielnik thinking makes sense. But if I as a non-native speaker would search through documentation, I would be looking for something like clear selection, cancel selection or reset selection. So I would rather stick with `clearSelection` name.

nmielnik:
If you're talking about the native selection instance, you'd access it by having the helper method take the `document` or `window` object as an argument, so you can call `.getSelection()` (we do this for almost all of the helpers in `selection.js`)

src/js/core.js
nmielnik: :+1:

nmielnik:
Closing this in favor of #1018 , @CupOfTea696 thanks for opening the issue and getting the ball rolling on this.

PR #835, opened 2015-10-08
ImageCaption multiline issue in Firefox 

ryandh:
https://github.com/orthes/medium-editor-insert-plugin/issues/230

Fix #834

PR #986, opened 2016-02-20
Fix #834 - Don't call formatblock when cursor is within block element

### LINKED ISSUE #834: Issue with Image caption, Multiline support in Firefox
it's a issue when using the plugin,

https://github.com/orthes/medium-editor-insert-plugin/issues/230

nmielnik:
This is a continuation of @ryandh 's work in #835.

When the cursor is directly within a block element (ie `<figcaption>`, `<h1>`, etc.), when the code calls `document.execCommand('formatBlock', null, 'p')` it did some unsavory things to the html, and in the case of `<figcaption>` in Firefox, it actually would keep moving the cursor to the beginning of the line after each keystroke, which presented your text in reverse order.

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -308,16 +308,58 @@ describe('Content TestCase', function () {
         });
 
         it('should call formatBlock when inside a non-header and non-anchor', function () {
-            this.el.innerHTML = '<p>lorem ipsum</p>';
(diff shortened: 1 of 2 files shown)


j0k3r:
Looks good to me! :+1: 
One less if :white_check_mark: 
Kudos to @ryandh for the first PR.

Also, the bug is kinda funny, writing in reverse order haha

Issue #982, opened 2016-02-18
disableExtraSpaces: true it's possible to insert more spaces


suskind:
With `disableExtaSpaces:true` if you have a space between two words and click with the mouse right after the first word and then press space bar, it's possible to insert one more space, and repeating the same more and more

rishijain:
Fixed via #987 .

PR #987, opened 2016-02-21
handle extra space when space already exists 🚧

rishijain:
Fix for #982 . Changes includes a test case which reproduces the bug filed in #982 .

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -166,6 +166,25 @@ describe('Content TestCase', function () {
 
             expect(evt.preventDefault).toHaveBeenCalled();
         });
+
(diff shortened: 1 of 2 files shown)


orthes:
:+1:

spec/content.spec.js
```diff
@@ -166,6 +166,25 @@ describe('Content TestCase', function () {
 
             expect(evt.preventDefault).toHaveBeenCalled();
         });
+
```
nmielnik: Could you add a comment that links to the issue that the test replicates?

rishijain: Done @nmielnik .

Issue #985, opened 2016-02-18
When dragging image into editor, `editableInput` fires before image src is set


nmielnik:
@kaave attempted to fix this with #938 but the fix had gone stale.

The issue is that when dragging an image into the editor, the file-dragging extension first inserts an image tag, then waits for the file to download before it updates the `src` of the `<img>` tag.

Since it inserts the `<img>` first, the `editableInput` event will fire BEFORE the `<img>` actually has a source and contains the actual image contents.  Ideally, the `<img>` wouldn't be inserted until it actually has a source so the `editableInput` gets triggered at a point where the image is valid.

LeeKevin:
All fixed.

nmielnik:
Fixed by #990 (thanks @LeeKevin).

The fix should is available in v5.14.4 (https://github.com/yabwe/medium-editor/releases/tag/5.14.4)

PR #990, opened 2016-02-24
Fix for #985

LeeKevin:
PR for #985

### DIFF
--- a/spec/drag-and-drop.spec.js
+++ b/spec/drag-and-drop.spec.js
@@ -53,7 +53,12 @@ describe('Drag and Drop TestCase', function () {
         });
 
         it('should remove medium-editor-dragover class and add the image to the editor content', function () {
-            var editor = this.newMediumEditor(this.el);
+            var editor = this.newMediumEditor(this.el),
+                editableInputListener = jasmine.createSpy();
+
+            editor.subscribe('editableInput', editableInputListener);
+            expect(editableInputListener).not.toHaveBeenCalled();
+
             fireEvent(editor.elements[0], 'dragover');
             expect(editor.elements[0].className).toContain('medium-editor-dragover');
             fireEvent(editor.elements[0], 'drop');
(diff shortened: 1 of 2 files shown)


src/js/extensions/file-dragging.js
```diff
@@ -74,17 +74,12 @@
...
-                var img = this.document.getElementById(id);
-                if (img) {
-                    img.removeAttribute('id');
-                    img.removeAttribute('class');
-                    img.src = fileReader.result;
-                }
-            }.bind(this);
+            // attach the onload event handler, makes it easier to listen in with jasmine
+            fileReader.addEventListener('load', function (e) {
+                var addImageElement = document.createElement('img');
```
nmielnik: Could you use the `this.document` reference here for creating the element?

spec/drag-and-drop.spec.js
```diff
@@ -37,26 +37,55 @@ describe('Drag and Drop TestCase', function () {
...
+                // Pass into the event handler our dummy image source
+                eventListener.calls.mostRecent().args[1]({
+                    target: {
+                        result: 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7'
+                    }
+                });
+
+                // Expect that the image is inserted
                 expect(MediumEditor.util.insertHTMLCommand).toHaveBeenCalled();
             }
```
nmielnik: In the spirit of #985 could you add an assert in here to make sure `editableInput` fires whenever the image is inserted?

nmielnik:
This is great, thanks @LeeKevin! 

Just had a couple small recommendations if you don't mind.

nmielnik:
Looks like there's some test failure for IE9 and Edge, are you able to look into them at all?  I can try to help otherwise.

IE9
IE9

Edge
Edge

nmielnik:
So in IE9, `FileReader` doesn't exist, so seems like the `beforeEach()` for the drag-and-drop tests should just check for the existence of `FileReader` and not try to create a spy if it doesn't exist.

The Edge issue appears to just to be a bug where our implementation of `insertHTMLCommand` doesn't trigger `editableInput`.  I'll work on a fix for that ASAP.

nmielnik:
I believe #993 will resolve the test failure in Edge, so you'll only need to resolve the issue where `FileReader` doesn't exist in IE9 which is causing the drag-and-drop test case to fail.

j0k3r:
@LeeKevin I guess you have to rebase your branch with our master to fix Edge test

LeeKevin:
Hopefully this works.

j0k3r:
The rebase looks goods.

nmielnik:
Thanks for removing the `IE9` check, feature detection > browser sniffing.

All tests passed ([build log](https://travis-ci.org/yabwe/medium-editor/builds/111798019)) including the cross-browser testing in SauceLabs, thanks for the work @LeeKevin !

PR #993, opened 2016-02-25
Fix #992

### LINKED ISSUE #992: Edge: editableInput event not triggered when inserting HTML via util.insertHTMLCommand()
Looks like the custom implementation that's being used in place of `document.execCommand('insertHTML')` doesn't result in `editableInput` being triggered in Edge.

This was discovered by @LeeKevin 's tests in #990

nmielnik:
Currently, we have 2 special fallbacks for when browsers don't support some expected functionality:

1. If browsers don't fully support firing `input` on contenteditable (IE, Edge), medium-editor monitor calls to `document.execCommand` so that it can manually trigger the custom `editableInput` event.
2. If browsers don't fully support `document.execCommand('insertHTML')`, medium-editor has some fallback logic to manually insert html into the editor.

When the logic for 2) happens, the manual insertion of HTML causes either the built-in `input` event to fire, or it causes the custom `input` detection for 1) to trigger.

However, in Edge, it appears the custom detection logic doesn't trigger when the logic in 2) happens, so calls to `insertHTMLCommand()` won't ever trigger `editableInput` in Edge.

The fix for this is:

1. When the code monitors calls to `document.execCommand` by creating a wrapper method in place of it, it will also expose another helper method (`document.execCommand.callListeners`) which can be used to manually notify listeners that `document.execCommand` has been called, but won't actually try to execute the real code.
2. When `util.insertHTMLCommand()` executes, it will check to see if the new `callListeners` method exists, and if it does, it will call it so that the internal logic can make sure `editableInput` is triggered.

Fixes #992.  This fix is needed in order for #990 to be completed.  Thanks to @LeeKevin for helping to expose this!

### DIFF
--- a/spec/events.spec.js
+++ b/spec/events.spec.js
@@ -254,6 +254,48 @@ describe('MediumEditor.Events TestCase', function () {
 
             MediumEditor.Events.prototype.InputEventOnContenteditableSupported = originalInputSupport;
         });
+
+        it('should expose a method for calling all listeners manually', function () {
+            var originalInputSupport = MediumEditor.Events.prototype.InputEventOnContenteditableSupported;
+            MediumEditor.Events.prototype.InputEventOnContenteditableSupported = false;
+
+            spyOn(document, 'execCommand').and.callThrough();
+            var origExecCommand = document.execCommand,
+                mockInstance = {
+                    options: { ownerDocument: document }
+                },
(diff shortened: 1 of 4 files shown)


orthes:
Looks good to me. Good job figuring this Edge cases (pun intended :)).

PR #995, opened 2016-02-27
Remove hardcoded inlined display style for anchor editor

PSyton:
For better forms customization with CSS it's better to handle show|hide
state by class.
I've appended additional class for visible form and remove inline style
modifications.
Also I't moved show/hide form functional to form implementation.

### DIFF
--- a/src/sass/components/_toolbar-form.scss
+++ b/src/sass/components/_toolbar-form.scss
@@ -38,3 +38,7 @@
         text-decoration: none;
     }
 }
+
+.medium-editor-toolbar-form-active {
+    display: block;
+}
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -14,6 +14,23 @@ describe('Anchor Button TestCase', function () {
     });
 
     describe('Anchor Form', function () {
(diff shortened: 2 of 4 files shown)


orthes:
Thanks! This makes sense. But can you remove `dist/js/medium-editor.js` from the pull request? Also I see that you removed an old test from `spec/anchor.spec.js`, but you didn't add a new one. Can you do that? Thank you.

orthes:
Would you mind adding some tests as well?

PSyton:
I will add tests this days

src/sass/components/_toolbar-form.scss
```diff
@@ -38,3 +38,7 @@
         text-decoration: none;
     }
 }
+
+.medium-editor-toolbar-form-active {
+    display: block;
+}
```
j0k3r: Could you please add an empty line at the end of the file?

PSyton:
_toolbar-form.scss already has empty line at the end.

PSyton:
Added simple test

orthes:
Great! Looks good. Thank you.

PR #996, opened 2016-02-27
Update OS X to 10.11 in Sauce Labs

orthes:
Just a cosmetic change. I've noticed that we're not testing against the latest OS X 10.11 which causes that we're not testing against the latest Safari 9, but against Safari 8 ([the only version supported in OS X 10.10 by Sauce Labs](https://saucelabs.com/platforms/#osx))

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,7 +63,7 @@ module.exports = function (grunt) {
             platform: 'Windows 10'
         }, {
             browserName: 'googlechrome',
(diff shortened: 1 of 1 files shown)


j0k3r:
Can we keep Safari 8 and Safari 9 as a bonus? :)

orthes:
Your wish is my command :)

j0k3r:
Great :+1:

Issue #865, opened 2015-11-01
Text inside editable span wrongly wrapped with <p>


wollmers:
For proofreading and corrections of OCR output I want to allow only editing on the lowest level of `<span>word</span>`.

This is a (very simplified) example of the HTML structure:

    <p>
       <span class='line'>
          <span class='word'>The</span> 
          <span class='word'>quick</span>
       </span>
    </p>

    <script>var editor = new MediumEditor('.word');</script>

Now when I click into one of the words, the text inside the `span` is wrapped with `<p>` at the first key stroke and looks like this:

    <p>
       <span class='line'>
          <span class='word'>
              <p>The</p>
          </span> 
          <span class='word'>quick</span>
       </span>
    </p>

Tried hard, but could not find a configuration option, work-around or the location in the source code causing this.

I would expect to keep the HTML elements as they are if valid HTML. Also wrapping with an element without some class like e.g. 'medium-created' does not give me the chance to overrule the CSS with `display: inline;` So the lines are split into paragraphs for each word, when I want to navigate from one editable word to the next with the TAB key. That destroys the WYSIWYG.

wollmers:
Same behaviour with a plain 

    <h1>Header</h1>

which also gets the text wrapped with `<p>` after any keystroke, even just cursor right:

    <h1><p>Header</p></h1>

Also the layout does not keep position. Before:

![bildschirmfoto 2015-11-01 um 19 37 57](https://cloud.githubusercontent.com/assets/1275557/10870650/412ae1f2-80d0-11e5-8c97-70d8750ca95a.png)

After:

![bildschirmfoto 2015-11-01 um 19 40 14](https://cloud.githubusercontent.com/assets/1275557/10870664/72fccf92-80d0-11e5-9def-4ee8b4ecba62.png)

And worse for the spans:

![bildschirmfoto 2015-11-01 um 19 43 18](https://cloud.githubusercontent.com/assets/1275557/10870692/e234e50c-80d0-11e5-87f4-a58305df243c.png)

Ok, seems to be a browser specific issue. The above is for Chrome Version 46.0.2490.80 (64-bit).

Here Safari Version 9.0 (9537.86.1.56.2):

![bildschirmfoto 2015-11-01 um 19 51 34](https://cloud.githubusercontent.com/assets/1275557/10870729/07382cfa-80d2-11e5-9aa6-8f0e70275d91.png)

And FirfFox 41.0.2 seams visually ok:

[...]

Issue #829, opened 2015-10-05
Inner <p> on the element even if MediumEditor is applied to a text element


brunomperes:
Version: 4.12.11

I am calling MediumEditor to this element:
````HTML
<p>Lorem ipsum dolor sit amet consectetur.</p>
````

But after any key press, the element is transformed into:
````HTML
<p><p>Lorem ipsum dolor sit amet consectetur.<p></p>
````

Same happens to h1 - h6.
The behaviour is probably caused because of this [line](https://github.com/yabwe/medium-editor/blob/4.12.x/src/js/core.js#L144). Which causes to call formatBlock, if node has no children.

Since I haven't seen any recommendation to only apply the Editor to wrapping divs, my guess is that it is not an expected behaviour.

I'm thinking that a verification before formatting the block to add a <p> inside should fix it.
Something like:

````javascript
var skipFormattingTags = ['P', 'H1', 'H2', 'H3', 'H3', 'H4','H5', 'H6'];
if (skipFormattingTags.indexOf(node.tagName) === -1){
  if (node.getAttribute('data-medium-element') && node.children.length === 0) {
    this.options.ownerDocument.execCommand('formatBlock', false, 'p');
  }
}
````

I'd be happy to send a PR if the solution seems sufficient.

**Comment by nmielnik** [comment]
@brunomperes is there a reason why you wouldn't want to start with a `<div>` as the medium-editor element? Something like:

```html
<div class="editor"><p>Lorem ipsum dolor sit amet consectetur.</p></div>
```

where the `<div>` is what gets passed when initializing MediumEditor?

brunomperes:
Actually, yes. There are two unwanted behaviours (not sure if contenteditable's or editor's):

1) When I have a wrapper div, and click at the end of the line and press delete, it deletes other tags. For instance, clicking on the final mark and pressing delete twice, it changes from:
````HTML
<div class="editor"><p>Lorem ipsum dolor sit amet consectetur.</p><i class="fa fa-gear"></i><i class="fa fa-trash"></div>
````
To:
````HTML
<div class="editor"><p>Lorem ipsum dolor sit amet consectetur.</p></div>
````


2) It removes the default element tag when the all the content is removed, then another delete is pressed.
So, when I have:
````HTML
<div class="editor"><h2></h2></div>
````

nmielnik:
[...]

Issue #149, opened 2014-02-20
Inconsistent <p> wrapping


schnittchen:
Hi,

When the container div contains only, say, a single line of text, the text will jump as soon as the User moves around using cursor keys. The reason is that the line gets wrapped in a `<p>`. This does not happen when I make the initial content end in `<br/>`.

nmielnik:
@schnittchen apologies it's been so long, but we're trying to go back and cleanup some issues.

Are you able to help us determine if this issue still happens?

j0k3r:
And if it still happens, tell us on which version you are testing. Also, give some precision on which browsers your problem happens

schnittchen:
I won't find time soon. Feel free to ping me some time.

schnittchen:
Looks like this is still an issue: when the initial text value is "ddd", the editor will change it to "ddd<br>". After inserting a newline in the middle of the word and removing it again, the word is wrapped in a <p> tag. That's three times the same thing (to the user) being potentially rendered differently.

schnittchen:
Oh, and gh's markdown spoiled my comment, so here's the middle version with url encoding: "ddd%3Cbr%3E"

nmielnik:
Thanks for the clarification, looks like this is yet another issue of weird browser handling of contenteditable content.  I've had thoughts about trying to enforce that there can only be only block elements as direct children of the contenteditable `<div>`, meaning `ddd` could never be the content, it would always have to be `<p>ddd</p>`...

This is definitely something the medium-editor code could try to fix-up if anyone wants to give it a shot.

nmielnik:
This issue seems somewhat related to #34, if we try to fix this one it should fix that issue as well (or vice versa)

nmielnik:
As reported by @yoshokatana in #34 

> Ditto on that. I set `disableReturn` to true (which gets rid of it in most cases), but those extraneous `<p>` tags are still being added if I paste in text. I even did this, but it doesn't seem to be doing anything:

```js
paste: {
      forcePlainText: false,
      cleanPastedHTML: true,
      cleanTags: [
        'meta',
        'script',
        'style',
        'img',
        'object',
        'iframe'
      ],
      cleanReplacements: [
        [/<p>/ig, ''], // this isn't actually getting rid of them?
        [/<\/p>/ig, '']
      ]
    }
```

yoshokatana:
[...]

PR #997, opened 2016-02-27
Fixed single text nodes getting wrongly wrapped with p-tags on any keypress

### LINKED ISSUE #907: Changing Text Input Wraps text in <p> Tags
I have two editable content divs which are being stacked vertically and the text in both of them is centered. When I edit the fields, even just adding one character, the whole text is wrapped in <p> tags. This interferes with my page's formatting.

If this is intended behaviour rather than an issue, please can someone point me to the function or line number where the <p> tags are being applied? I've looked and cannot find it.

Thanks.

harzzn:
If an editor element had no children but contained text, any keypress would wrongly wrap the text in p-tags. Fixes #907, #829, #865, #149

See in-depth explanation here https://github.com/yabwe/medium-editor/issues/907#issuecomment-189675611

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -141,6 +141,26 @@
         }
     }
 
+    function handleKeydown(event) {
+
+        var node = MediumEditor.selection.getSelectionStart(this.options.ownerDocument);
+
+        if (!node) {
+            return;
+        }
+
+        // If the element is empty and user has inputted content,
+        // we'll mark the element with an attribute so we can act based on it on keyup event
+        if (!MediumEditor.util.isKey(event, MediumEditor.util.keyCode.TAB) &&
+            !MediumEditor.util.isKey(event, MediumEditor.util.keyCode.DELETE) &&
+            !MediumEditor.util.isKey(event, MediumEditor.util.keyCode.BACKSPACE) &&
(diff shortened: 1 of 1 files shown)


j0k3r:
Could you fix the coding standard error https://travis-ci.org/yabwe/medium-editor/builds/112262868#L374 ?

harzzn:
@j0k3r sure thing, fixed now.

nmielnik:
@harzzn thanks for looking into this, would you mind adding some tests to help verify it fixes the 4 issues you have mentioned?

Judging by the change, I'm assuming you're just looking for text nodes or `<span>`s that are the direct child of the parent editable element, and making sure those are wrapped in `<p>` tags?  Is that right?

harzzn:
@nmielnik Yeah, I can add the appropriate tests. I'll write a more detailed explanation of the functionality once I have the tests ready, as it should be easier to explain together with the tests.

nmielnik:
@harzzn could you give us an idea of what kind of tests you were planning to add?  I'd be willing to help get those tests added but I would definitely want your thoughts to make sure things are covered correctly.

j0k3r:
@harzzn any news about tests?

PR #1001, opened 2016-03-06
Destroy ie autolinks

### LINKED ISSUE #961: Auto-Link extension re-enables 'AutoUrlDetect' in IE when destroyed, even if there are still other actives instances present
Currently, when medium-editor is destroyed (via a call to `.destroy()`), it calls `destroy()` on all extensions that implement it.

When the auto-link extension is initialized, it makes sure that the `AutoUrlDetect` feature is turned off in IE [here](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/auto-link.js#L41), to prevent auto-link and IE's built in auto url detection from conflicting (especially since our auto-link implementation behaves better).  

When the auto-link extension is destroyed, it turns `AutoUrlDetect` back on [here](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/auto-link.js#L44-L49)

The problem is there can be multiple instances of MediumEditor instantiated, which could mean multiple version of the auto-link extension are also instantiated.  Thus, as soon as one instance is destroyed, it will turn `AutoUrlDetect` back on, which means this could now interfere with all the other auto-link extension instances which are still around.

The fix would be to have some sort of global mechanism that is shared accross all instances of the auto-link extension, which would make sure the `AutoUrlDetect` feature is only re-enabled if _all_ instances of auto-link have been destroyed.

dyzhu12:
Only re-enable AutoUrlDetect when all auto-link instances have been destroyed -- created an `isLastInstance()` function which checks if all other Medium Editor instances with auto-link enabled have been destroyed.

Fixes #961

### DIFF
--- a/src/js/extensions/auto-link.js
+++ b/src/js/extensions/auto-link.js
@@ -41,9 +41,20 @@
             this.document.execCommand('AutoUrlDetect', false, false);
         },
 
+        isLastInstance: function () {
+            var activeInstances = 0;
+            for (var i = 0; i < window._mediumEditors.length; i++) {
+                var editor = window._mediumEditors[i];
+                if (editor !== null && editor.getExtensionByName('autoLink') !== undefined) {
+                    activeInstances++;
+                }
+            }
+            return activeInstances === 1;
+        },
+
         destroy: function () {
             // Turn AutoUrlDetect back on
-            if (this.document.queryCommandSupported('AutoUrlDetect')) {
+            if (this.document.queryCommandSupported('AutoUrlDetect') && this.isLastInstance()) {
                 this.document.execCommand('AutoUrlDetect', false, true);
             }
         },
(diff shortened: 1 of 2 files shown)


src/js/extensions/auto-link.js
```diff
@@ -41,9 +41,20 @@
             this.document.execCommand('AutoUrlDetect', false, false);
         },
 
+        isLastInstance: function () {
+            var activeInstances = 0;
+            for (var i = 0; i < window._mediumEditors.length; i++) {
+                var editor = window._mediumEditors[i];
```
nmielnik: Each extension has a reference to `this.window` and `this.document`, which are equivalent to the `contentWindow` and `ownerDocument` options that can be optionally passed into the editor.  Could you update these references to `window` here to be `this.window` instead?

nmielnik:
@dyzhu12 The fix looks great and thanks for adding a test! Just need a small update and then we can merge this in.

dyzhu12:
@nmielnik Updated the `window` references to `this.window`

j0k3r:
Thanks ! :+1:

nchase:
:thumbsup:

PR #1009, opened 2016-03-10
Added support for multiple instances of contenteditable areas on the …

rvetere:
…same page :) oh and just a new ignore for .iml - project files of intteliJ IDEA.

I'm just starting a total rewrite from scratch of https://vp24.com - where i already implemented this awesome medium-editor clone more than a year ago. I've dirty hacked the source quite a bit back those days to implement much more things that where needed for vp24. things like color-picker, font-selection, alignments and more buttons leading back to basic functionality of our own editor.

and first of all - the support to have more than one instance of a working content-editable area on the same page, and everything handled from one instance of MediumEditor.

well, maybe i was just too dumb back those days and i didn't see an existing solution in there to achieve that smoothly - or is really everybody else dealing just with one big-ass instance per page/state? ^^

so, you can go ahead and tell me how that would be intended to achieve and we can forget this pull-request :D

i've added two new functions:

* addElements
`this will add an array of new elements to this.elements, and re-run initElements`

* cleanupElements
`this one will run trough all this.elements and check if they still exists in the DOM - and simply filter them out`

then i've added a new if in "initElements", so i can check if an element is already initialized and skip it.

after that i'm ending up with a solution like this in my ng2 directive:

`
      if (!window.mediumEditor) {
         window.mediumEditor = new MediumEditor(el.nativeElement, { ... });
      } else {
         window.mediumEditor.addElements(el.nativeElement);
      }
`

i'm slowly starting trough with my new editor implementation - so i guess i will contribute more things to editor in the future - like i have should done way earlier but better now than never! ;)


rvetere:
sorry i forgot to add new tests already.. will come up with them later, but tell me first if this is even worth to merge ;)

nmielnik:
@rvetere any reason you decided to close this PR?

rvetere:
@nmielnik yeah different ones back at this point..

* first i thought that MediumEditor is already supporting mutliple instances by creating multiple versions of a mediumEditor on the same page
* second, the code was far away from completeness at this point

in the meanwhile i realized that i still need the ability to have only 1 instance of MediumEditor and attaching/detaching several contenteditable areas dynamically during runtime.

Now i started over with a new try on the implementation and this is so far much better now :) It is also handling the event-attaching of custom events now if you have a subscribe on the editor which was not the case in first version. And it is not breaking the existing tests anymore like in the first try.. ^^

So.. currently i'm still hunting one last bug in my system where i'm not sure right now if it is in the new code of my mediumEditor or just in the implementation of it.

Let me check that first in the next few days.. and then i will come up with a new PR, introducing all those changes/improvements ;)

nmielnik:
Awesome, sounds great, thanks @reveter!  Let us know if there's anything else we can answer or help with.

PR #1011, opened 2016-03-13
Fix anchor showform

rvetere:
The anchor form extension received some great changes lately to remove the hardcoded show/hide functionality into css-classes. However, the change was not included in the currently commited dist/ folder.

After a new build, i noticed a wrong placement behavior of the anchor form - i found out that the new call for it was placed in the wrong order in the code.

It is mandatory that the show call is executed before the "setToolbarPosition" call - or it will fail to place correctly.

So this is all fixed with this pull-request ;)


j0k3r:
Thanks for that could you remove the dist files from your PR ?

rvetere:
But the dist files were exactly missing those new changes with the whole refactoring of anchor.js (and so the error with the miss-placement was not "there" when you just worked with the current dist, until you built the project one time on your own)

It is the idea to remove all dist things in pull-requests - and create the final dist on master after the merge?

So best would be if i just delete the branch again - close this pull-request - make this single one-line change in anchor.js - commit this in a new branch and make a new pull-request..? ;)

j0k3r:
As written in [our contributing file](https://github.com/yabwe/medium-editor/blob/master/CONTRIBUTING.md), you shouldn't make modification to dist files, only to source files.
We generated dist file when we prepare a new release only.

> So best would be if i just delete the branch again - close this pull-request - make this single one-line change in anchor.js - commit this in a new branch and make a new pull-request..? ;)

That would be the best move :smile:

rvetere:
Sorry i didn't read the contributing file completely - i just rushed over it ^^ And yes, this makes totally sense... but i'm asking myself why those anchor.js changes where missing in current dist/.. something went wrong during the last release maybe?

Anyway i will continue with this move like described - will be back soon with a new PR ;) :D

And i'm working on more contributions too... i've almost done with a nice mobile support on touch devices and some more things ;)

j0k3r:
The last release didn't included these changes: https://github.com/yabwe/medium-editor/compare/5.14.4...master

rvetere:
Ah ok, classic "chicken/egg" situation then :) Cool that i already found the issue so far before the next release

PR #1012, opened 2016-03-14
Fix anchor showform

rvetere:
As discussed earlier - here the new branch, cleaned up, without anything touching in dest/ ;-) Containing only the fix for the anchor "show-form" logic


j0k3r:
As a follow up, previous discution started at #1011

PR #1018, opened 2016-03-23
Adding clearSelection method to selection.js

dggriffin:
PR for #983.

I saw that there's currently an open pull request (#984) to implement this functionality, but for core rather than selection.js, so I decided to write this and the appropriate unit-tests.


nmielnik:
@dggriffin: This is great, thanks for following up with this, adding some tests, and some great comments!

j0k3r:
:+1:

gsouf:
Could not find it in version 5.23. What is removed?

PR #1020, opened 2016-03-24
Fix #994 - Prevent Firefox NS_ERROR_FAILURE: on keyup

### LINKED ISSUE #994: Firefox - contenteditable on h1 tag throws error
Example is here: https://plnkr.co/edit/7tzumlKQ3XRY6ABFeSna?p=preview . When I try to activate MediumEditor on tag h1, Firefox throws very general error NS_ERROR_FAILURE from this line of code:

```
// function handleKeyup
if (node.getAttribute('data-medium-element') && node.children.length === 0) {
    this.options.ownerDocument.execCommand('formatBlock', false, 'p'); // this line throws error
}
```

Of course, obvious solution is not using h1 tag, but I wonder is this known error. It even has some sense as h1 is not the best candidate for contenteditable, because nesting some tags inside h1 can causes incorrect markup.

nextend:
In Firefox if the editable is a node without children, then NS_ERROR_FAILURE: error thrown for every keyup until it gets a child node. 

Issue: https://github.com/yabwe/medium-editor/issues/994

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -149,7 +149,8 @@
             return;
         }
 
-        if (MediumEditor.util.isMediumEditorElement(node) && node.children.length === 0) {
+        if (MediumEditor.util.isMediumEditorElement(node) && node.children.length === 0  &&
++            !MediumEditor.util.isBlockContainer(node)) {
             this.options.ownerDocument.execCommand('formatBlock', false, 'p');
         }
 
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -149,8 +149,7 @@
             return;
         }
 
-        if (MediumEditor.util.isMediumEditorElement(node) && node.children.length === 0  &&
-+            !MediumEditor.util.isBlockContainer(node)) {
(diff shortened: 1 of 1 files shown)


rishijain:
@nextend Can you add some test case reproducing the bug you fixed.

src/js/core.js
```diff
@@ -149,7 +149,7 @@
             return;
         }
 
-        if (MediumEditor.util.isMediumEditorElement(node) && node.children.length === 0) {
+        if (MediumEditor.util.isMediumEditorElement(node) && node.children.length === 0 && !MediumEditor.util.isBlockContainer(node)) {
```
nmielnik: Are you passing an element other than a `<div>` as the root editor element?

The call to `MediumEditor.util.isMediumEditorElement(node)` will only return true if `node` is the element that the editor is based on.  So, if you're using a `<div>` as the root element then this extra condition in the `if` won't actually do anything because it will always return `false` (since `<div>` is not in the list of `blockContainerElementNames`).

nextend: I'm using a `<h1>` tag for the editor as I want the user to edit only its content.

nmielnik:
The commits here were moved into #1052, the work will be completed as part of that PR.

PR #1022, opened 2016-03-25
Use innerHTML in Events.updateInput() function only once to improve performance

orthes:
`Events.updateInput()` is called on every `keypress`. But when the editor contains large amount of data, `innerHTML` becomes very expensive - performance-wise. This simple PR cuts the lag time to half by calling `innerHTML` only once, not twice.


j0k3r:
:+1:

PR #1024, opened 2016-03-27
Dynamic add elements

rvetere:
As a follow-up of: https://github.com/yabwe/medium-editor/pull/1009
Check readme: https://github.com/rvetere/medium-editor/tree/dynamic_add_elements#dynamically-add-elements-to-your-instance

Is this stuff ok for you..? First i thought, we don't need something like that, i'm just going to create a new MediumEditor instance for each angular2 directive i have made for it in my app.

But this was ending in a huge dom-mess (dozens of toolbars in the DOM where only 1 would be needed for example) and also some strange behavior. Also i don't exactly knew first when those instances are killed, so.. i was afraid ending with a huge memory leak using this pattern.

And i created again my old addElements/cleanupElements functionality but much better now.

* had to change the code at central parts to only run once on every element, so we can call those methods now as much as we want without danger (and only the newly added element in the array will be initialized for example)

* i had to change the event contentCache variable from an array to an object, so it won't change the remaining indexes just by removing an element somewhere in it -> otherwise we can run very fast into huge problems where the cleanupElements fails because a given medium-editor-index is not found anymore in the cache variable

* i had to change the central "medium-editor-index" from a simple array index to a UID - this is because of an actually needed uniqueness.. if i have a cache key like "1" from an array having the indexes "0, 1, 2, 3".. and i'm deleting the index "1" -> "2, 3" will remain now.. the next newly added element will have index "3" which already exists in the code -> bam, error!

* i've introduced a new eventsCache variable which just stores the info which handlers were already attached to a certain elements uid, so we can multi-run this code now and skip already attached handlers on certain elements

I've also changed some for loops which were not using the high-speed pattern `for (var i = 0, len = array.length; i < len; i++)` -> which causes to access the "length" variable, calc the length of the array only once and not on every loop step like the classic variant `for (var i = 0; i < array.length; i++)` ;)

Oh, there are also new spec tests checking for the delicate parts of the implementation and i've got so far a 100% coverage in my report of all those new functions. I've also added a new section to the central README.md describing this possibility.


nmielnik:
@rvetere thanks for looking into this and getting a solution out there!  This is definitely a desirable feature and would be a great addition to the codebase.

I had some thoughts about ways to approach this and was hoping we could try brainstorming together about this (as well as anyone else that would like to participate).  I've created a [gitter channel](https://gitter.im/nmielnik/addDynamicElements) to make the discussion a little more fluid, care to join?

src/js/core.js
```diff
@@ -1088,6 +1138,66 @@
...
+
+            // Add new elements to our internal elements array
+            this.elements = this.elements.concat(filtered);
+
+            // Initialize all new elements (we check that in those functions don't worry)
+            initElements.call(this);
+            reAttachHandlers.call(this);
+        },
+
+        cleanupElements: function () {
```
nmielnik: What would you think about reducing this method to just be something like `removeElements` where you pass the elements you want to remove into the method directly?  Seems like something like finding if the element exists in the DOM or not, or the `checkExistenceTagName` option might be better to leave outside of the editor?

src/js/core.js
```diff
@@ -1088,6 +1138,66 @@
                 target.innerHTML = html;
                 this.events.updateInput(target, { target: target, currentTarget: target });
             }
+        },
+
+        addElements: function (elements) {
```
nmielnik: So we'll also need to do something like what is in `createElementsArray` here, since we need to make sure we do the setup for `<textarea>` elements.  Perhaps we could even try to break up parts of the `setup()` method so that the code paths for setting up the editor and adding elements can be as similar as possible?

src/js/events.js
```diff
@@ -317,11 +375,37 @@
         },
 
         attachToEachElement: function (name, handler) {
+            // build our internal cache to know which element got already what handler attached
+            if (!this.eventsCache) {
+                this.eventsCache = {};
+            }
+
             this.base.elements.forEach(function (element) {
+                var existingCache = this.eventsCache[element.getAttribute('medium-editor-index')];
```
nmielnik: Another way to approach making sure that all events handlers are attached to added elements is to modify this method to just cache all the handlers that had been passed to it in the past (maybe just an array of handler methods).

Then, when a new element is passed, we can just loop through the cached handlers and attach each handler to the newly added element.

src/js/core.js
```diff
@@ -351,7 +355,7 @@
             element.setAttribute('data-medium-editor-element', true);
             element.setAttribute('role', 'textbox');
             element.setAttribute('aria-multiline', true);
-            element.setAttribute('medium-editor-index', index);
+            element.setAttribute('medium-editor-index', this.guid());
```
nmielnik: I'm a bit concerned about replacing the `medium-editor-index` attribute directly with a guid for backward compatibility...

I saw you remove a `medium-editor-uid` attribute whenever the editor is destroyed but I don't see anywhere in the code where that attribute is created.  Were you planning to set the guid as `medium-editor-uid`?  Could you do that instead of changing the value of `medium-editor-index`?  If there isn't a need for `medium-editor-index` moving forward, we can look to remove it in the next major release (6.0.0).

nmielnik:
Any chance you might be able to clean up some of the commit history?  Seems like there are some duplicated commits, and some commits that updated/reverted the dist files...

src/js/core.js
rvetere: Totally agree - i'm refactoring the code accordingly i would say:

* remove cleanupElements
* remove the new function `findParentElementByTagName`
* add new function `removeElement`
* add new function `removeElements`
* add new function `addElement`

After that we're even - there are two fn's to add, two to remove with the same naming pattern. So it is also clear which one will take an element and which one an array as input ;)

Newly, the remove functionality will only check for the given "medium-editor-index" attribute to find his internal elements to match the ones we want to remove. So it is completely up to the implementation how one will do his "existence check" or, if you want to directly remove it on your "remove" hook :)

src/js/core.js
rvetere: Yes i've checked that again.. i was implementing my addElements method accordingly to the things i did find in "setup".. so i closely looked into `createElementsArray` and i thought, the things with the selector and quering the elements should not be "executed again" - it feels natural to me that i call an "addElements" method already with the elements as an array, not just giving him a new selector-string.

So, after looking closely into it again with the `<textarea>` in mind, the only code that actually should re-run is this part of `createElementsArray`:

`his.elements = [];`
`        elements.forEach(function (element, index) {`
`            if (element.nodeName.toLowerCase() === 'textarea') {`
`                this.elements.push(createContentEditable.call(this, element, index));`
`            } else {`
`                this.elements.push(element);`
`            }`
`        }, this);`

It is only the last loop which checks again all elements of his array, and if he finds anything as <textarea> - he converts it into a div.

So, what about if we extract this little loop into a new local function, maybe "_convertTextareas" and, we can use it at both parts? in `createElementsArray` and in my `addElements` ;) It is the only thing i see missing from the parts i let re-run i would say..

rvetere:
Ok i don't know how i can update an existing PR - maybe it's just because i'm still working on my existing branch my own fork.. ;)

Since i should cleanup this commit history anyway.. i guess i try to refactore first the last parts - and then just go ahead and create a total new branch, where i manually merge only the changes into it.

Then i can create a new PR which should be totally fine regarding the history ;)

I've already removed `cleanupElements`, added `removeElement(s)` and `addElement` and, made a solution for our textarea problem which was missing.. Now the last thing would be your nice idea about events handling and re-attaching all of them ;)

PR #1025, opened 2016-03-28
Fix 1014 - Convert spaces to %20 in links if linkValidation is true for the anchor extension

### LINKED ISSUE #1014: Space in .medium-editor-toolbar-input breaks anchor preview
If I add space inside a link, it is accepted and after that the link is no more editable through anchor preview.

dyzhu12:
Converts all instances of spaces to url friendly %20 if linkValidation for anchors is enabled. Example in screenshot mirrors same test case from the issue.
image

Fixes #1014

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -345,7 +345,25 @@ describe('Anchor Button TestCase', function () {
 
             link = editor.elements[0].querySelector('a');
             expect(link).not.toBeNull();
-            expect(link.href).toBe(validUrl.toLowerCase());
+            expect(link.href).toBe(validUrl);
+        });
+        it('should not change spaces to %20 if linkValidation is set to false', function () {
+            var editor = this.newMediumEditor('.editor', {
+                anchor: {
+                    linkValidation: false
+                }
+            }),
+                link,
+                anchorExtension = editor.getExtensionByName('anchor'),
+                validUrl = 'http://te%20s%20t.com/';
+
+            selectElementContentsAndFire(editor.elements[0]);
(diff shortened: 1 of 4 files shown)


src/js/extensions/anchor.js
```diff
@@ -242,7 +242,7 @@
                 return 'tel:' + value;
             } else {
                 // Check for URL scheme and default to http:// if none found
-                return (urlSchemeRegex.test(value) ? '' : 'http://') + value;
+                return (urlSchemeRegex.test(value) ? '' : 'http://') + value.replace(/\s/g, '%20');
```
nmielnik: Would it be worth trying to use `encodeURI` here to take care of all characters that need to be escaped, including spaces?

spec/anchor.spec.js
```diff
@@ -329,6 +329,42 @@ describe('Anchor Button TestCase', function () {
             expect(link).not.toBeNull();
             expect(link.href).toBe(validUrl.toLowerCase());
         });
+        it('should change spaces to %20 for a valid url if linkValidation options is set to true', function () {
+            var editor = this.newMediumEditor('.editor', {
+                anchor: {
+                    linkValidation: true
```
nmielnik: Good idea to have this feature be triggered by the `linkValidation` option.  Could you update the documentation to mention this additional functionality when `linkValidation` is used?  Check `README.md` and `OPTIONS.md` for references.

dyzhu12: @nmielnik updated the documentation and changed the space replacement with `encodeURI`

PR #1030, opened 2016-03-30
Updated test - Made the diff between input and validUrl only spaces

dyzhu12:
Made sure the only differences between input and validUrl are the spaces, since the tests are specifically for converting spaces -- hope this fixes the SauceLabs failures.


nmielnik:
The fix for this is continued in #1033

PR #1033, opened 2016-03-30
Fix failing anchor test

nmielnik:
Turns out, Chrome, IE, and Edge automatically encode urls when you set them as an href of an `<a>` tag.  However, Firefox, Safari, and our version of PhantomJS don't automatically encode these urls.

This means the test that was verifying that the href is NOT encoded when `linkValidation` is set to `false` would pass in PhantomJS, Firefox, and Safari, but was failing in the other browsers.

This change updates the test case to check for the value of the encoded/not-encoded url before it actually gets set into the html.


rishijain:
Thanks @dyzhu12 @nmielnik :+1:

PR #1038, opened 2016-04-01
Add/remove elements to existing instance

rvetere:
This PR will replace my last [Dynamic add elements PR](https://github.com/yabwe/medium-editor/pull/1024) - now with one single commit :+1: 

The changes are much simpler now as everything is refactored according to our discussion @nmielnik ;)

* All events of `attachToEachElement` will be stored in a simple `this.eventsCache` array
* When adding an element, only the necessary code is run and we can simply iterate the eventsCache again and attach everything only to this new element in `attachAllEventsToElement`
* `addElements` is written in a way such that `setup` will use it when initializing, so a shared code path is used both for initialization and adding elements dynamically.
* The elements-exists check together with `cleanupElements` is removed and replaced with simple `removeElement(s)` instead, also the option flag is gone for it
* readme is rewritten to match the new logic for a "cleanupElements"

Now the only thing left i remember is your concern about the introduction of a guid instead of the existing array index on `medium-editor-index`, regarding backwards compatibility.

Yeah that's right.. unfortunately i really need this change because otherwise the cleanup functionality becomes unclean (irony) over time especially with the routing state changes of my ng2 app.

Anyway this needs to be mentioned in a change log for the release which will include this change. And, as far as one is not using the index from medium directly as a number/index for his own array.. it would still "just work" even with the new guid.. and... at least for myself.. why would one do that - i'd rather hack my own extension/change in my own fork/branch to introduce the change i want and call an api :P

***
##### What's Left
- [x] Allow selector string to be passed to `addElements` and `removeElements`
- [x] Fix bug and broken test around passing `<textarea>` in via `addElements`
- [x] Add tests around `removeElements` for `<textarea>` elements
- [x] Fix issue with cleaning up `<textarea>` via calls to `removeElements` (and add a test!)
- [x] Add more tests around `removeElements`
- [x] Create a failing test, and fix issue around adding an element with `data-disable-return` and/or `data-disable-double-return` attribute.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -218,17 +218,23 @@
             selector = [selector];
         }
         // Convert NodeList (or other array like object) into an array
-        var elements = Array.prototype.slice.apply(selector);
+        this.elements = Array.prototype.slice.apply(selector);
 
         // Loop through elements and convert textarea's into divs
-        this.elements = [];
-        elements.forEach(function (element, index) {
+        convertTextareas.call(this);
+    }
+
+    function convertTextareas() {
+        var converted = [];
+        this.elements.forEach(function (element, index) {
             if (element.nodeName.toLowerCase() === 'textarea') {
-                this.elements.push(createContentEditable.call(this, element, index));
+                converted.push(createContentEditable.call(this, element, index));
             } else {
-                this.elements.push(element);
+                converted.push(element);
             }
         }, this);
+
+        this.elements = converted;
     }
 
     function setExtensionDefaults(extension, defaults) {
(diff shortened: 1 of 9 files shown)


src/js/core.js
```diff
@@ -645,6 +667,7 @@
                 element.removeAttribute('role');
                 element.removeAttribute('aria-multiline');
                 element.removeAttribute('medium-editor-index');
+                element.removeAttribute('medium-editor-uid');
```
nmielnik: It doesn't look like this attribute gets added anywhere in these changes...

src/js/core.js
```diff
@@ -1088,6 +1113,81 @@
...
+            this.elements = filtered;
+        },
+
+        removeElements: function (elements) {
+            elements.forEach(function (element) {
+                this.removeElement(element);
+            }.bind(this));
+        },
+
+        guid: function () {
```
nmielnik: How would you feel about moving this into `util.js`?  Just a way to reduce the number of methods exposed in the core API since it's not tied to the editor instance.

PR #1050, opened 2016-04-10
 Fix #1047 - Add 'positionedToolbar' custom event

### LINKED ISSUE #1047: Cannot edit link from anchor preview if toolbar is shown
I have input field with medium editor, where `updateOnEmptySelection: true`

In anchor preview i have `showWhenToolbarIsVisible: true` and it works fine, it shows preview when i hover over added link(and I have visible toolbar).

But, when i click on preview to edit it, it redirects me to the inserted url, instead of opening for edit.

Can you confirm if this is issue, or I'm missing some configuration setting, so it's not working fine?
I investigated it, but I couldn't find any setting..

nmielnik:
This PR fixes #1047 where clicking on an anchor preview when using the `showWhenToolbarIsVisible` option doesn't edit the link but instead navigates away from the page.

The issue was rooted with the toolbar manually calling into the anchor-preview's `hidePreview` event whenever the toolbar was being updated.  This flow didn't account for the `showWhenToolbarIsVisible` option and resulted in some improper handling of the click.

Now, instead of the toolbar calling directly into the anchor-preview extension, the toolbar & editor will emit a `positionedToolbar` custom event, which the anchor-preview can subscribe to.  This will allow the anchor-preview to handle the `showWhenToolbarIsVisible` option properly and fix the bug.

### DIFF
--- a/CUSTOM-EVENTS.md
+++ b/CUSTOM-EVENTS.md
@@ -23,6 +23,7 @@ If you need to override the editor's bult-in behavior, try overriding the built-
 - [Toolbar Custom Events](#toolbar-custom-events)
   - [`hideToolbar`](#hidetoolbar)
   - [`positionToolbar`](#positiontoolbar)
+  - [`positionedToolbar`](#positionedtoolbar)
   - [`showToolbar`](#showtoolbar)
 - [Proxied Custom Events](#proxied-custom-events)
       - [`editableClick`](#editableclick)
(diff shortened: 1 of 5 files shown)


CUSTOM-EVENTS.md
```diff
@@ -152,6 +153,9 @@ These events are triggered by the toolbar when the toolbar extension has not bee
 ### `positionToolbar`
 `positionToolbar` is triggered each time the current selection is checked and the toolbar's position is about to be updated. This event is triggered after all of the buttons have had their state updated, but before the toolbar is moved to the correct location.  This event will be triggered even if nothing will be changed about the toolbar's appearance.
 
+### `positionedToolbar`
+`positionedToolbar` is triggered each time the current selection is checked, the toolbar is displayed, and the toolbar's position was updated. This differs from the `positionToolbar` event in that the visibility and location of the toolbar has already been changed (as opposed to before those changes happene). This event will be triggered even if nothing was changed about the toolbar's appearance.
```
orthes: Small typo: "happene"

orthes:
Great fix. Using events in cases like this nicely cleans the code.

I've found small typo in docs, other than that , it's good to be merged.

CUSTOM-EVENTS.md
nmielnik: Good catch, typos in documentation always bug me.

Issue #180, opened 2014-03-22
 Enable / disable editors individually


daviferreira:
As stated by @akiarostami on #178, it should be possible to disable editors for specific elements inside a set of MediumEditors, by ID or something.

nmielnik:
Some additional thoughts and requirements for this were brought up in #694

houd1ni:
I need it in my app, will take a look at this soon!

Olgagr:
@houd1ni Any progress with that ? I need to disable toolbar for some elements inside contenteditable as well.

houd1ni:
@Olgagr yes! Thank you for reminder) I'll try to finish it tomorrow.

Olgagr:
Thanks @houd1ni !

PR #1073, opened 2016-04-27
Master

houd1ni:
Original issue: https://github.com/yabwe/medium-editor/issues/180

I failed to write the test: only created an appropriate shell for it.

I'm quite new in this kind of testing, so I just describe my approach.

The test has failed, though I'd added a child element with `.className='disableClass'` right options `{disable: '.disableClass'}` and actually in manual testing with debug tools, in at least chrome, events  haven't been fired: the `selectiveListener` shell returns before an original Event.


nmielnik:
@houd1ni sorry for taking so long to look at this.  My first question is, is there any reason you wouldn't just set `contenteditable="false"` on the elements you don't want to be editable?

houd1ni:
@nmielnik we discussed it in gittet a month ago. You'd suggested this approach, then I've tried and faild at least in cases like ```<span contenteditable="true">  </span>``` for instance, over an image, it shows ugly #. And without crutches it cannot be avoided.

src/js/util.js
```diff
@@ -1052,6 +1052,10 @@
             } else {
                 el.parentNode.removeChild(el);
             }
+        },
+
+        toArr: function (arraylike) {
+            return [].slice.call(arraylike);
```
nmielnik: Is there a benefit to calling `[].slice()` as opposed to `Array.prototype.slice`?

nmielnik:
Ah yes, thanks for reminding me!

So looks like I created a merge conflict for you with some of the changes that I merged in, apologies for that but we'll want to get that fixed first.

I can try to help out with writing tests and what-not if that's what you were looking for, or were you looking for additional feedback on the change and pointers on how to add some tests?

src/js/core.js
```diff
@@ -231,6 +231,21 @@
         }, this);
     }
 
+    function offDisabledElements(elements) {
+        // Should be checked that it exists outside the func.
+        elements.forEach(function (element) {
```
nmielnik: Once you merge the latest master into your branch, you'll likely want to move this logic into `initElement`.  It's a function that will be called for every editor element during initialization and when elements are added dynamically.

src/js/events.js
```diff
@@ -16,11 +16,27 @@
         // Helpers for event handling
 
         attachDOMEvent: function (targets, event, listener, useCapture) {
+            var selectiveListener = function (event) {
```
nmielnik: Was the goal of this feature to not emit _any_ DOM events on 'disabled' elements, or to just prevent any medium-editor behavior on 'disabled' elements?

houd1ni:
Im alive, will take a look on this soon :)

src/js/util.js
houd1ni: Sorry for megaslow response. Yes: it's just a bit smaller :)

houd1ni: But now I'm not sure that is's not slower. Gonna test it.

Issue #1060, opened 2016-04-14
Creating new line inside H-blocks


Greeny7:
There are some problems with headings - they are dissapearing when you make new line with caret inside this block.
I think it is about the same problem, which is this issue/PR about - https://github.com/yabwe/medium-editor/pull/382
Now creating new line in headings/block containers works good, but only if we have no nested tags inside. For example, if you make your H-tag underlined (by creating <u> tag inside), and hit return inside this block - function handleKeyup() will get <u> as parent block and will not prevent old editor behavior.
![editor-bug](https://cloud.githubusercontent.com/assets/12188900/14533637/8a8387b2-026e-11e6-99a3-b194cda00788.gif)

orthes:
Thanks for reporting this. I tried this in a pure `contenteditable` element without medium-editor to see if it's something that's broken in `contenteditable` itself, or we're doing something wrong there. And it looks like `contenteditable` without medium-editor works correctly, so we're screwing something up here.

Greeny7:
@orthes Yes, if I understand right, Medium Editor creating P elements on return-key hit to prevent the creation of div's - because div element creation on new line is a default behavior of condenteditable.
It follow from the handleKeyup() function in core.js. 
And there is a condition, that on return key hit it must check the type of parent tag and if it is a block element, like h1/h2/quote etc. - it must prevent formatBlock execution on new line.

I don't fully understand yet in what situations we need to prevent div-creating and how we must build the condition in handleKeyup() function, so I made a temporary repair only for H-tags (because headings destroying is most noticeable and ugly):
https://github.com/Greeny7/medium-editor/commit/302a0ecfbb1cd5a33c81d1e5577ae4c2759b8c06

I would be grateful if someone will clarify how to do a better condition to fully resolve this issue.

nmielnik:
@Greeny7 this is definitely something that needs to be understood and addressed.  I think the commit you pointed to makes sense, and it could be worth opening a PR agains the main repo and see if we can try to add some tests and improve upon it more iteratively to see if any other issues might come out of it.

There are many issues open related to the chunk of code that handles `keyup` in core.js (#907, #834, #829, #865, #149, and more).

There have also been a few PRs recently making some adjustments to this code ( #997, #994, #986).

It doesn't seem like any one person specifically has a good grasp of this code and what all it's intended to cover.  However, due to how critical that code path is, and all the problems that keep coming up, we might just have to keep experimenting with fixes like what you proposed and see what happens 😕

PR #1074, opened 2016-04-29
protect headers from newline

Greeny7:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | not yet
| Fixed tickets    | https://github.com/yabwe/medium-editor/issues/1060
| License          | MIT

### Description

This is temporary fix for https://github.com/yabwe/medium-editor/issues/1060 issue, I did it for a project in which I participate. It must be imporoved of course, because this bug involves not only `<h*>` elements, but every block-type elements.
It will be good if someone will give ideas for improvements and tests for this PR, I'm afraid to break something).

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -143,7 +143,9 @@
 
     function handleKeyup(event) {
         var node = MediumEditor.selection.getSelectionStart(this.options.ownerDocument),
-            tagName;
+            closestBlockTag,
+            tagName,
+            isHeader;
 
         if (!node) {
             return;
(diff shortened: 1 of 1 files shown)


nmielnik:
@Greeny7 thanks for opening this and apologies for the delay in getting some feedback.

1. Take a look at this part of [content.spec.js](https://github.com/yabwe/medium-editor/blob/master/spec/content.spec.js#L341-L394).  There you can find some example tests that test some of the code around where your change is. I would definitely recommend adding some tests to further verify the functionality you're adding. To run the tests, just run `grunt` from the command line.

2. Since you're calling `getClosestBlockContainer()` to get the first block element, you might be able to rip out the existing code that checks if the `node` var is a block element or part of an `<li>`.  You may just be able to get the element returned from calling `getClosestBlockContainer()` and see if you get a result other than the editor element itself.  If it's not the editor `<div>` then you know there was no block container and it's ok to allow the `<p>` to be inserted.  However, there might be some elements that we want to be able to have a nested `<p>` element inside of it (`<blockquote>` elements come to mind for some reason) so we might want to think about allowing other block containers to still contain a `<p>`, but seeing how everything work if you just inspect the block container would definitely be worth it if you're can try it out.

PR #1086, opened 2016-05-13
execActionInternal should pass opt to document.execCommand if action …

Stackle:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    |
| License          | MIT

### Description

I think it should pass opts parameter of MediumEditor.execAction through the 3rd parameter of Document.execCommand.

It will give a chance to caller to be able to execute command that are not support by MediumEditor's internal command, for example, `foreColor` command.

If a given command is not require any value to be passed, it will send undefined or maybe I think the browser just ignore it.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -338,7 +338,7 @@ describe('Anchor Button TestCase', function () {
                 link,
                 anchorExtension = editor.getExtensionByName('anchor'),
                 expectedOpts = {
-                    url: 'http://te%20s%20t.com/',
+                    value: 'http://te%20s%20t.com/',
                     target: '_self'
                 };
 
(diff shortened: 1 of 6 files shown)


j0k3r:
Good idea ! :+

nmielnik:
@Stackle good thinking with this!  One thing to think about, since we treat the `opts` argument as an object with one or more properties for something like `createLink`, `fontSize`, and `fontName`, should we try to keep that pattern and just pass `opts.value` to calls to `execCommand`?

I'm suggesting the name `value` because of what this argument is described as in [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) and [MSDN](https://msdn.microsoft.com/en-us/library/ms536419(v=vs.85).aspx).

So instead of
```js
editor.execAction('foreColor', 'red');
```
it would instead be
```js
editor.execAction('foreColor', { value: 'red' });
```

What do you think? If you do this, we could probably fix up `createLink`, `fontSize`, and `fontName` as a followup to standardize on using `opts.value` instead of `opts.url`, `opts.size`, and `opts.name` respectively.

Stackle:
@nmielnik 
Sorry for late in reply.

I agree with your idea, it sound great!
By the way, I think it might break someone extensions so, we have to make it backward compatible and deprecate an old style.

I'll working on that in a few days.

nmielnik:
Awesome, thanks for this @Stackle!

Issue #869, opened 2015-11-05
Extra data added when pasting from word.


brijeshb42:
A whole mess of html comments get added in a separate `p` tag when I try to paste contents from a word document.
![download](https://cloud.githubusercontent.com/assets/717550/10960303/9489ae2e-83a9-11e5-9406-f618a8c5db15.png)

nmielnik:
ugh

cperryk:
+1
![screenshot 2015-12-15 16 27 44](https://cloud.githubusercontent.com/assets/1660273/11824621/de9dd53a-a348-11e5-83b0-c4a120b1f107.png)

j0k3r:
@cperryk @brijeshb42 Could you give us a Word file that we can use to reproduce the bug?
Also, which browser are you using? What is your MediumEditor configuration?

Charlex:
Here's an example:
[comments example.docx](https://github.com/yabwe/medium-editor/files/86383/comments.example.docx)

Here's just a snippit of the comments I'm seeing pasted in to an empty paragraph:

```
[...]

PR #1087, opened 2016-05-17
Ensure #869 is fixed

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | #869 
| License          | MIT

### Description
#869 had been fixed by some previous changes, but there weren't any tests to verify it was fixed.  This change adds a test which verifies a (horrific looking) sample paste from html is handled correctly.

I also added another regex to filter out some of the `<!--[if !supportLists]--> ... <!--[endif]-->` elements added when copying and pasting from Word.


j0k3r:
Looks like you got a pretty nice Word document example !
Nice :+1:

PR #1088, opened 2016-05-17
Allow passing options to 'non-medium' commands

abijr:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | don't know it they are needed.
| Fixed tickets    | 
| License          | MIT

### Description

Allow passing options to comands executed through the 'execAction' method. Previously, ME didn't allow the execution of commands that need options (such as `insertHTML`, `insertText`, etc.)

E.g. allow passing text to the insertText action

--

I don't know where to add the tests for this, if any are required.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -562,7 +562,7 @@
             return result;
         }
 
-        return this.options.ownerDocument.execCommand(action, false, null);
+        return this.options.ownerDocument.execCommand(action, false, opts);
     }
 
     /* If we've just justified text within a container block


nmielnik:
@abijr could you take a look at #1086 ? Seems like the change you're going for is exactly the same as that one.  We were also suggesting using `opts.value` instead of just `opts` for what is passed into `execActionInternal`.

Also, you could build a test similarly or even the same way as the test that exists in that #1086

abijr:
Didn't notice that one... Yeah, it's basically the same. 

I can get the test done by the weekend.

nmielnik:
@abijr looks like #1086 was ready to go so I just merged that guy in...apologies for not taking your PR, it really is quite the coincidence that both you and @Stackle submitted a change for the exact same functionality within a few days of each other.  If you want to add any other new functionality or bug fixes don't hesitate to keep contributing!

abijr:
No problem. After checking, I actually liked his solution better than mine. It's just that my changes were sitting in there for about one or two weeks, and I just submitted the PR before looking again at the issues.

PR #1089, opened 2016-05-18
Fix null pointer during the cleanupElement event

lexderuijter:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description

I ran into an issue that after manually adding elements to the editor (through addElements), and later deleting them again there (through removeElements) was a TypeError being thrown.

It appears that contentCache is not set when only using dynamically added elements, thus while removing that element it will run into an error resulting the element not being removed properly.

This checks if contentCache is set.

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -366,7 +366,7 @@
             if (index) {
                 this.detachAllEventsFromElement(element);
                 
-                if(this.contentCache) {
+                if (this.contentCache) {
                     delete this.contentCache[index];
                 }
             }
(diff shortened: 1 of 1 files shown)


j0k3r:
Could you please fix coding style?

lexderuijter:
Updated

nmielnik:
@lexderuijter just an FYI, if you initialize the editor without passing any pre-existing elements, there will be a bunch of functionality that won't work.  None of the extensions will be initialized, and none of the code which fixes up the html after the user types will work.

Issue #1068, opened 2016-04-19
Quote is not working after orderedlist/unorderedlist list was removed


Olgagr:
### Description

If you have an orderedlist/unorderedlist in the middle of other text and remove it, then you try to add quote on this text, quote is not working any more.

### Steps to reproduce

1. Add paragraph
2. Add second paragraph
2. Make second paragraph ordered list
3. Remove ordered list from second paragraph
4. Try to add quote on the second paragraph

**Expected behavior:** blockquote should be created

**Actual behavior:** blockquote is not created

**Link to an example:** https://plnkr.co/edit/ZzFM4v85yfWT89GFWKA3?p=preview - Steps can be reproduce here.

### Versions

- medium-editor: 5.16.1
- browser: Chrome, Firefox, Safari 
- OS: OS X

nmielnik:
Thanks @Olgagr, I was able to reproduce this.  Interestingly enough, it seems like you're able to convert the 2nd paragraph into other block elements (pre, h1, h2, etc.) but not a blockquote.  Also, if you convert it into another block element, and then remove it (ie click `H1` twice) you are then able to convert it into a `<blockquote>` properly...

This is definitely a medium-editor bug as I was unable to reproduce this with a plain contentedtiable element.  The default contenteditable behavior appears to wrap the 2nd paragraph in a `<p>` after removing the ordered list.  Since it's wrapped in a `<p>`, the browser seems to convert that into a `<blockquote>` just fine.

However, in the medium-editor version, after removing the ordered list, the 2nd paragraph just becomes a text node and there is no `<p>`.  The browser is not able to handle converting the plain text node into a `<blockquote>`, but it's able to convert the plain text node into other block elements (`<pre>`, `<h1>`, etc).

Olgagr:
@nmielnik In core.js (https://github.com/yabwe/medium-editor/blob/master/src/js/core.js#L889), when adding list, we remove wrapping <p> element. So when we remove list, there is no wrapping element.  I think we would like to leave the existing behaviour that when creating the list, the wrapping <p> tag is removed. In utils.js we have unwrap method. Do we have something like wrap method as well ?

nmielnik:
Unfortunately we don't have a wrap method since wrapping a set of text within a new element is much harder than unwrapping.

[...]

PR #1097, opened 2016-05-24
fix bug #1068

vovatolstousov:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no 
| New tests added? | not yet
| Fixed tickets    | #1068 
| License          | MIT

### Description

[Bug fixed]

--


rishijain:
Hi @vovatolstousov , can you remove the `dist` files from the commit and make your changes appropriate file under in `src/js`.

j0k3r:
Also, could you try to add some tests so that before your change the test is failing and after applying your change the test is green?

Olgagr:
@vovatolstousov I think this not resolve the problem. The problem is that when a list is removed, the content of the list becomes text node, not block element (in this case it should be paragraph).

PR #1103, opened 2016-05-30
Add getContent method (Fix #492)

### LINKED ISSUE #492: Expose way to get the content of the fields monitored by MediumEditor
As per @dodyg 's suggestion in #139: 

>What is the impact of calling multiple ```new MediumEditor``` on the same element?
>
>I have about 10 fields that I bind using  ```new MediumEditor('.editable')``` however to get a value from a specified field i have to resort to do
>
>```
>new MediumEditor('#content_editor').serialize().content_editor.value
>```

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | #492
| License          | MIT

### Description

Expose simple helper for retrieving content from a MediumEditor instance as requested in #492.  The `getContent()` method is a natural compliment to `setContent()` and behaves as expected based on the behavior of `setContent()`.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1145,6 +1145,15 @@
             }
         },
 
+        getContent: function (index) {
+            index = index || 0;
+
+            if (this.elements[index]) {
+                return this.elements[index].innerHTML.trim();
+            }
+            return null;
+        },
+
         checkContentChanged: function (editable) {
             editable = editable || MediumEditor.selection.getSelectionElement(this.options.contentWindow);
             this.events.updateInput(editable, { target: editable, currentTarget: editable });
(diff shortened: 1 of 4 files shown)


src/js/core.js
```diff
@@ -1145,6 +1145,15 @@
             }
         },
 
+        getContent: function (index) {
+            index = index || 0;
+
+            if (this.elements[index]) {
+                return this.elements[index].innerHTML.trim();
```
j0k3r: I was about to suggest using `.serialize().value` instead but regarding `serialize()` it seems that it serialize all referenced MediumEditor instance since it loops on all `this.element`. Am I right?

nmielnik: Yeah, `serialize()` returns the value of all of the elements, so you're right this is just a helper to avoid getting the html for all the elements.  I could change serialize to to call `getContent()` once for each element, but it didn't seem like much of a gain...

PR #1104, opened 2016-05-31
Fix #719 - Respect form.reset for textarea elements within forms

### LINKED ISSUE #719: Reset Medium editor.
It would be nice if the editor (when tied to a textarea) automatically reset when the textarea's for resets.

Currently when I reset my forms using ```form.reset()``` everything gets reset (including the textarea) but not the medium editor.

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | #719 
| License          | MIT

### Description

The goal of this change was ensure that when a `<textarea>` element within a form is reset, that the editor `<div>` created by the editor is also reset back to its initial value.

To accomplish this, this change set also introduces a new `resetContent()` method for resetting any element back to its initial value when the editor was initialized or when the element was first added to the editor.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1178,17 +1178,20 @@
         },
 
         resetContent: function (element) {
-            var elements = this.elements;
+            // For all elements that exist in the this.elements array, we can assume:
+            // - Its initial content has been set in the initialContent object
+            // - It has a medium-editor-index attribute which is the key value in the initialContent object
+
             if (element) {
-                elements = [];
-                if (this.elements.indexOf(element) !== -1) {
-                    elements.push(element);
+                var index = this.elements.indexOf(element);
+                if (index !== -1) {
+                    this.setContent(initialContent[element.getAttribute('medium-editor-index')], index);
                 }
+                return;
             }
-            elements.forEach(function (el, idx) {
-                if (el) {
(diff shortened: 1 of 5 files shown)


src/js/core.js
```diff
@@ -333,15 +333,15 @@
         return !this.options.extensions['imageDragging'];
     }
 
-    function createContentEditable(textarea, doc) {
```
j0k3r: Shouldn't this be a problem if people were previously using this method by giving `doc`? I guess javascript will just ignore the second parameter?

nmielnik: Since `createContentEditable` is just an internal function, there shouldn't be a way for anyone to actually call it outside of our internal methods. So we don't need to worry about backward compatibility (if that's your concern) since only our internal code has access to it.

As part of the work to introduce `addElements()` I modified this method to take the `document` in as a param (instead of using `this.options.ownerDocument`) since that was the only time it used `this` as the editor.  However, in order to attach the event handler and call the helper method, it again needed a reference to the editor, so I went back to using `this` as the editor and no longer needed to take the `document` in as a param...that was my thought process at least.

PR #1108, opened 2016-06-01
Support any extension preventing blur of editor (Fix #830)

### LINKED ISSUE #830: ME's blur event shouldn't fire when users interact with custom extension's toolbar
I'm writing an ME extension that displays a custom toolbar when caret is on specific elements within *contenteditable* element. I would like to make this extension's toolbar to be part of ME so that interacting with my custom toolbar (clicking) doesn't fire ME's custom *blur* event.

Why do I want that? Similar to ME's own toolbar when you interact outside of the editor, toolbars should hide. That's why I'm also subscribing to ME's *blur* event to hide my custom toolbar.

**But the problem is** that when my toolbar shows up and I click on it, ME regards this as interactivity outside of the editor and fires *blur* event which consequently hides my toolbar that I just clicked on...

It would be best if ME had the ability during extension init phase to also register custom DOM sub-tree that extensions may be using and regard it as ME's DOM. Similar to how it does with anchor preview, and toolbar to determine external event. This is very much baked in to provided extensions, but custom extensions are left out in the cold.

**Excerpt from medium-editor.js** that does this checking
```javascript
// Check if the target is external (not part of the editor, toolbar, or anchorpreview)
var externalEvent = !MediumEditor.util.isDescendant(hadFocus, target, true) &&
					!MediumEditor.util.isDescendant(toolbarEl, target, true) &&
					!MediumEditor.util.isDescendant(previewEl, target, true);
```
This code should be checking custom extensions' registered DOM elements as well... This would be an **ideal solution**.

**But how do I solve this using current ME's behaviour?** I should basically intercept *blur* event somehow and cancel its triggering when I detect that interaction is with my custom toolbar. Or at least that's how I see it.

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | yes
| New tests added? | yes
| Fixed tickets    | #830
| License          | MIT

Currently, if any extension render an element on the page, if the user clicks on that element it counts as a click on an element outside of the editor, which causes the editor to `blur` and potentially hide the toolbar altogether.

This change set introduces a new `getInteractionElements()` method that any extension can implement.  This allows extensions to let MediumEditor know that it has created HTML elements that, when clicked, should not count as external interactions, and thus should not cause the editor to `blur` when being interacted with.

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -1,6 +1,26 @@
 (function () {
     'use strict';
 
+    function isElementDescendantOfExtension(extensions, element) {
+        return extensions.some(function (extension) {
+            if (typeof extension.getInteractionElements !== 'function') {
+                return false;
+            }
+
+            var extensionElements = extension.getInteractionElements();
+            if (!extensionElements) {
+                return false;
+            }
+
+            if (!Array.isArray(extensionElements)) {
+                extensionElements = [extensionElements];
+            }
+            return extensionElements.some(function (el) {
+                return MediumEditor.util.isDescendant(el, element, true);
+            });
+        });
+    }
+
     var Events = function (instance) {
         this.base = instance;
         this.options = this.base.options;
@@ -377,21 +397,16 @@
         },
 
         updateFocus: function (target, eventObj) {
-            var toolbar = this.base.getExtensionByName('toolbar'),
-                toolbarEl = toolbar ? toolbar.getToolbarElement() : null,
-                anchorPreview = this.base.getExtensionByName('anchor-preview'),
-                previewEl = (anchorPreview && anchorPreview.getPreviewElement) ? anchorPreview.getPreviewElement() : null,
-                hadFocus = this.base.getFocusedElement(),
+            var hadFocus = this.base.getFocusedElement(),
                 toFocus;
 
-            // For clicks, we need to know if the mousedown that caused the click happened inside the existing focused element.
-            // If so, we don't want to focus another element
+            // For clicks, we need to know if the mousedown that caused the click happened inside the existing focused element
+            // or one of the extension elements.  If so, we don't want to focus another element
             if (hadFocus &&
                     eventObj.type === 'click' &&
                     this.lastMousedownTarget &&
                     (MediumEditor.util.isDescendant(hadFocus, this.lastMousedownTarget, true) ||
-                     MediumEditor.util.isDescendant(toolbarEl, this.lastMousedownTarget, true) ||
-                     MediumEditor.util.isDescendant(previewEl, this.lastMousedownTarget, true))) {
+                     isElementDescendantOfExtension(this.base.extensions, this.lastMousedownTarget))) {
                 toFocus = hadFocus;
             }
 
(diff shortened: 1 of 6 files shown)


gpetrov:
I really like this one. 

I wonder if it will be possible for an extension to have "design-time" html just for the preview and activation within the editor - and "run-time" html that will be the final html that is saved in the content.

This will allow "run-time" html to get "translated" to "design-time" for display in editor and back on save.

Or is this another feature request? :)

nmielnik:
@gpetrov I don't fully understand the feature you're describing, but it sounds like it might be a separate feature request.  Could you open a separate issue with more details of what you're describing and try to include an example?

phiggins42:
@nmielnik  / @gpetrov I have implemented that with a custom "block" system. each block is contenteditable=false, with an optional inner contenteditable=true region. it allows custom chrome/templating around, but does not persist the chrome when the editor is serialized. each block has a "type", and an extension to handle it's own rendering/serialization. 

eg:

```
<div id="editor">
    <div data-block="$speciallist"><ul><li>item one</li></ul></div>
</div>
```

when you instantiate the editor, the `$speciallist` extension will draw a full rendered version of whatever special chrome, and the inner `UL` will be editable within the existing editable instance. so rendered is like:

```
<div id="editor">
       <div contenteditable=false data-block="$speciallist"><h3>This is a very special list!</h3><div contenteditable=true data-block-content><ul>...</ul></div></div>
</div>
```

where the chromed h3 is not editable or interactive in any way (though you can put buttons and any kind of UI in you desire) and when being serialized, only the inner `data-block-content` is returned in the markup. 

the $speciallist extension delegates it's button events on the editor node, so moving the inner block around within the editor canvas doesn't require teardown/re-event-listening. the paste handler I've written does the same reduction during paste, so if you were to copy/paste within the editor you're really only copying the serialized portion, which is re-built after paste end. 

a more clear usecase would be images. say all you want in the rendered output is 

```
<div data-type="$image"><img src="......"></div>
```

but want to maintain extra metadata about the image, or provide cropping/rotate/resize UI ... the $image extension would build up all the chrome, handle all the interaction, but be transparent to the editor itself, and wouldn't appear in any persistable data (allowing you to change the design of the chrome without having to change any of the underlying data. 

all of this, of course, breaks native undo/redo pretty badly. but it works. 

anyway i think this is what he's talking about ...

phiggins42:
to note, the original PR/code looks like it is talking about elements generated by an extension that are _outside_ of the editable canvas (which, because of parentage, would in fact blur the editor when interacted with). my block solution puts the interaction elements _inside_ the editor, and simply doesn't export them during serialize.

src/js/events.js
```diff
@@ -1,6 +1,26 @@
...
+            if (typeof extension.getInteractionElements !== 'function') {
+                return false;
+            }
+
+            var extensionElements = extension.getInteractionElements();
+            if (!extensionElements) {
+                return false;
+            }
+
+            if (!Array.isArray(extensionElements)) {
```
phiggins42: ie8 support dropped?

nmielnik: As far as I know, we've never supported IE8, and this is around for IE9 (which we still do support).  This is also not the first place we've used `Array.isArray()`.

nmielnik:
Ok, I think I get what you fellas are saying, and that's definitely not being addressed by this PR. Would either of you mind opening a separate issue describing the feature request?

gpetrov:
@phiggins42 this is exactly what I meant! I will write up another feature request about it @nmielnik 
It will be awesome if such functionality comes in the Mediaum-Editor - we can then make rich UI extensions.

PR #1112, opened 2016-06-02
Fix #1105 - Fix bug with anchor-preview when clicked

### LINKED ISSUE #1105: Anchor-preview extension needs to be refactored too
Because of the refactoring of Anchor Extension's showForm to take "value" option instead of "url"  
https://github.com/yabwe/medium-editor/commit/50503d5e4de4f8bb9535e51bde5949a0e61f48a8#diff-20ab57acb9b766e8d29282b1d91c3d28

Also the Anchor-Preview should be changed:

https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/anchor-preview.js#L158

to be "value" and not "url"

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | #1105 
| License          | MIT

### Description

When the anchor-preview was clicked, it was displaying 'undefined' in the anchor form, this was caused by a previous change where we changed over from using `opts.url` to `opts.value` to standardize on how we pass arguments into `document.execCommand()`.  Updating the anchor-preview was missed during that change.

There surprisingly isn't a test ensuring that when you click the anchor-preview, the `href` of that link is pre-filled into the anchor form, I've updated an existing test to assert that this happens.

### DIFF
--- a/spec/anchor-preview.spec.js
+++ b/spec/anchor-preview.spec.js
@@ -139,6 +139,9 @@ describe('Anchor Preview TestCase', function () {
             expect(toolbar.isDisplayed()).toBe(true);
             expect(anchor.isDisplayed()).toBe(true);
 
+            // textbox should contain the url
+            expect(anchor.getInput().value).toEqual('http://test.com');
+
             // the checkboxes should be unchecked
             expect(anchor.getAnchorTargetCheckbox().checked).toBe(false);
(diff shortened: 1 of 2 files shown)


rishijain:
👍 works perfect now.

PR #1113, opened 2016-06-04
Fix #234 - Cursor rendering incorrectly in firefox

### LINKED ISSUE #234: Cursor position bug in firefox
Make an editor with empty text. Then focus the editor in firefox, the cursor position is higher than the placeholder. Then I input some text in it, the cursor position become normal.

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #234 
| License          | MIT

### Description

In Firefox, when the editor element is completely empty (including when it doesn't have `<p><br /></p>`), the absolutely positioned pseudo-element used for a placeholder affects where the blinking cursor appears once the editor is focused.

When the pseudo-element is positioned `relative` in this case, everything renders as expected.  However, when positioned `relative`, the placeholder renders incorrectly for other cases and renders incorrectly in other browsers.

This fix ensures that only in firefox, and only when the editor element is completely empty, the placeholder is positioned `relative`.  As far as I can tell, this fixes the issue.

### DIFF
--- a/spec/placeholder.spec.js
+++ b/spec/placeholder.spec.js
@@ -138,7 +138,11 @@ describe('MediumEditor.extensions.placeholder TestCase', function () {
         if (match) {
             // In firefox, getComputedStyle().getPropertyValue('content') can return attr() instead of what attr() evaluates to
             expect(match[1]).toBe('data-placeholder');
-        } else {
+        }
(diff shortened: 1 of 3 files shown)


j0k3r:
:+1:

PR #1115, opened 2016-06-06
Fix slow CSS selector

nmielnik:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | updated
| License          | MIT

### Description

Fixed a slow CSS selector that our linter was complaining about:

```terminal
Linting dist/css/medium-editor.css...ERROR
[L209:C1]
WARNING: Unqualified attribute selectors are known to be slow. Unqualified attribute selectors are known to be slow. (unqualified-attributes) Browsers: All
```

To fix this, we can add a `medium-editor-element` class to the editor element, and update the CSS selector to use that instead of an unqualified attribute selector.

### DIFF
--- a/spec/elements.spec.js
+++ b/spec/elements.spec.js
@@ -32,10 +32,11 @@ describe('Elements TestCase', function () {
             expect(this.el.getAttribute('contenteditable')).toBeFalsy();
         });
 
-        it('should set element data attr medium-editor-element to true', function () {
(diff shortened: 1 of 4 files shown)


j0k3r:
I'm ok with that :+1:

Issue #1123, opened 2016-06-14
Editor breaks in electron environment 


sajus:
[Description of the bug or feature]
https://github.com/yabwe/medium-editor/blob/master/dist/js/medium-editor.js#L389
`if (typeof module === 'object') {
        module.exports = factory;
 }`

In electron environment, which use Nodejs, module is defined and editor tries to export itself. Which causes breaks to already exist electron code base.

sajus:
https://github.com/sajus/medium-editor/commit/61046c9cef3d5754f15395d47dfffbb35b7ec821

nmielnik:
@sajus is there something unique on the `module` object in electron that we could add to the detection here?  I'm not familiar with this at all, but it seems like we just need to differentiate between standard node and electron somehow inside the export logic.

nmielnik:
Seems like a workaround would be to delete the reference to `window.module` and `window.exports` like what is mentioned here: http://electron.atom.io/docs/faq/#i-can-not-use-jqueryrequirejsmeteorangularjs-in-electron

sajus:
@nmielnik you can detect electron environment from the below code.
`typeof module == "object" && process && process.versions && process.versions.electron`

Hope this resolves your query.

nmielnik:
@sajus can you try changing that chunk of code in your fork to this:

```js
(function (root, factory) {
    'use strict';
    var isElectron = typeof module == "object" && process && process.versions && process.versions.electron;
    if (!isElectron && typeof module === 'object') {
        module.exports = factory;
    } else if (typeof define === 'function' && define.amd) {
        define(function () {
            return factory;
        });
    } else {
        root.MediumEditor = factory;
    }
}(this, function () {
```

If that works for you, either you or I can open a PR and get the fix in.

sajus:
https://github.com/yabwe/medium-editor/pull/1125

durgeshahire4:
Please check PR of Medium Editor Tables : https://github.com/yabwe/medium-editor-tables/pull/30

sajus:
@j0k3r , @nmielnik can you please bump new version? Also NPM needs to be updated which is been used by medium-editor-tc https://github.com/tomchentw/medium-editor-tc-mention/blob/master/src/index.js#L1-L3

nmielnik:
@sajus we might try to get #1128 in first, and then we'll cut a new release

nmielnik:
@sajus 5.21.0 should now be available.

[...]

PR #1125, opened 2016-06-20
#1123 Fixes issue with electron environment

sajus:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 1123
| License          | MIT

### Description

Bug-ID #1123, fixes issue with electron environment.

### DIFF
--- a/src/wrappers/start.js
+++ b/src/wrappers/start.js
@@ -1,6 +1,6 @@
 (function (root, factory) {
     'use strict';
-    var isElectron = typeof module == 'object' && process && process.versions && process.versions.electron;
+    var isElectron = typeof module === 'object' && process && process.versions && process.versions.electron;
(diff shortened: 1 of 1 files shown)


src/wrappers/start.js
```diff
@@ -1,6 +1,7 @@
 (function (root, factory) {
     'use strict';
-    if (typeof module === 'object') {
+    var isElectron = typeof module == 'object' && process && process.versions && process.versions.electron;
```
j0k3r: Should be `===` instead of `==`

sajus: @j0k3r, thanks for pointing this out. I have updated this now.

j0k3r:
Thanks ! :+1:

Issue #714, opened 2015-06-30
Links broken if no protocol provided


filipkis:
When inserting links using the anchor toolbar button and typing for instance only www.google.com (instead of full path http://www.google.com) the href of the anchor will only be set to the typed value, which in turn will cause the links to be broken (directing to http://yourapp.com/www.google.com).

The protocol (http://) is nicely added to links that are created automatically (autoLink = true), but not for the ones created from the toolbar.

j0k3r:
We should add `http://` by default if the inserted link doesn't start with `https?://`.

filipkis:
On further inspection, the adding of protocol works if the `anchor: { linkValidation: true }` is set, so the question is should the protocol also be added when the validation option is turned off?

I'm guessing there should be some consistency between how autoLink and anchor work.

j0k3r:
In fact, the `linkValidation` is actually doing what you asked for. It adds `http://` on anchor link if it doesn't start with a protocol

See https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/anchor.js#L208-L211

I think the question is more, should `linkValidation` be turned `true` by default ?

And you're right about consistency about these two extensions.

nmielnik:
The problem here is that `linkValidation` is an implicit requirement of auto-link.  Auto-link can only do something if it actually detects a valid link, so there can't really be a concept of turning off link validation for auto-link.

So, I don't think these options can be synchronized, however I do think it could be valid for `linkValidation` to be `true` by default...as long as we can convince ourselves that this doesn't require a major release to do :sweat_smile:

j0k3r:
Thinking about it, why this kind of feature is disabled by default and why this is an option at all.

Since anchor is about to insert a link, why people should insert relative link instead of absolute one?

filipkis:
I'm guessing having it as an option would still be good. Maybe in some cases you want relative links because the content is supposed to run on different domains (e.g. local and production), or you might be just creating hash links.

j0k3r:
So we'll switch `linkValidation` defaut value to `true`. But we just need to allow hash url to be valid too.

@filipkis you're more than welcome to submit another PR about that :smile:

nmielnik:
For now, I think it will be better to leave things as there are currently, and not have linkValidation turned on by default.  I don't think the fact that auto-link requires valid links is a strong reason to make the link extension require protocols on urls.

[...]

PR #1143, opened 2016-07-13
Make linkValidation allow hash links

cruzanmo:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | 
| License          | MIT

### Description

Allows hash links, e.g. `#fragment-identifier`, when `linkValidation` is set to `true`. This was brought up in this discussion: https://github.com/yabwe/medium-editor/issues/714

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -203,6 +203,7 @@ describe('Anchor Button TestCase', function () {
             });
             expect(editor.elements[0].querySelector('a')).toBeNull();
         });
+
         it('should add http:// if need be and linkValidation option is set to true', function () {
             var editor = this.newMediumEditor('.editor', {
                 anchor: {
@@ -220,6 +221,7 @@ describe('Anchor Button TestCase', function () {
             expect(link).not.toBeNull();
             expect(link.href).toBe('http://test.com/');
         });
+
         it('should add tel: if need be and linkValidation option is set to true', function () {
             var editor = this.newMediumEditor('.editor', {
                 anchor: {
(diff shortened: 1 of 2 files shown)


spec/anchor.spec.js
```diff
@@ -329,6 +329,24 @@ describe('Anchor Button TestCase', function () {
...
+                anchorExtension = editor.getExtensionByName('anchor');
+
+            selectElementContentsAndFire(editor.elements[0]);
+            anchorExtension.showForm(validHashLink);
+            fireEvent(anchorExtension.getForm().querySelector('a.medium-editor-toolbar-save'), 'click');
+
+            link = editor.elements[0].querySelector('a');
+            expect(link).not.toBeNull();
+            expect(link.getAttribute('href')).toBe(validHashLink);
+        });
```
j0k3r: Could you add an extra space before and after this assertion?

cruzanmo: Good point, I will add spaces.

cruzanmo: I added consistent spacing between the tests, but did you mean spaces around L348?

src/js/extensions/anchor.js
```diff
@@ -235,7 +235,8 @@
             // Matches any alphabetical characters followed by ://
             // Matches protocol relative "//"
             // Matches common external protocols "mailto:" "tel:" "maps:"
-            var urlSchemeRegex = /^([a-z]+:)?\/\/|^(mailto|tel|maps):/i,
+            // Matches relative hash link, "#" followed by any character
+            var urlSchemeRegex = /^([a-z]+:)?\/\/|^(mailto|tel|maps):|^\#.+/i,
```
nmielnik: Should this allow just plain `#` as a valid link?

cruzanmo: In my opinion, the hash should be omitted if there is no fragment identifier; however, you are correct that it is valid, so it should be allowed here.

nmielnik:
Looks awesome @cruzanmo , thanks a ton!

Issue #857, opened 2015-10-28
Toolbar does not position properly when elementsContainer is inside absolute or fixed dom


disintegrator:
The top position of the toolbar gets screwed up because the window.pageYOffset is used in the calculation when `elementsContainer` is inside a parent with absolute/fixed position (think: modals).

On my page I follow these steps:
1. Scroll the page down a few pixels
2. Open a modal with medium-editor inside
3. Type and select text
Issue: toolbar vertical alignment is off

Note: I am using the editor in a react app with the following config (the two refs are divs):

```
this.editor = new MediumEditor(this.refs.editor, {
  elementsContainer: this.refs.medium
})
```

Seen below is a full screen modal over a scrollable body.
![image](https://cloud.githubusercontent.com/assets/678548/10787602/51225f02-7dc6-11e5-8d47-ce33c9a9a243.png)

nmielnik:
First off, it's good to hear you're having success with react. `contenteditable` and React have not gotten along for quite a while, as `contenteditable` inherently allows significant DOM manipulation outside of react's knowledge. Have you had to work through any issues with this?

There are 2 things I can think of around this:

1. There may need to be a change to `setToolbarPosition()` in `toolbar.js` (source code [here](https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/toolbar.js#L510)) to account for cases when it is inside an absolute or fixed position element.

2. There may be a workaround for your case using the 'diffLeft` and/or `diffTop` options for the toolbar.  You can find the documentation [here](https://github.com/yabwe/medium-editor/blob/master/OPTIONS.md#toolbar-options).

We would gladly accept a pull request to fix this for option 1, as the issue you are describing has likely been encountered by others, and probably be encountered by others moving forward as well :smile:

disintegrator:
@nmielnik thanks for the quick response.
With regards to react integration: I do make sure that the editor does not affect the virtual dom held by react.
I think diffLeft and diffTop should be treated as constants at initialisation time to fine tune the toolbar position.
After digging through the code, I have pinpointed the problem to be in positionToolbar() which simply relies on `this.window.pageYOffset` to determine the toolbar position which causes it to fall apart when placed under non-static parents. I am trying to figure out how to fix it by using looking at the toolbar's `offsetParent`s - I'll keep digging and create a PR if I have any success.

[...]

PR #1152, opened 2016-07-20
Fix #224 and #857 - toolbar in absolute container

### LINKED ISSUE #224: Positioning on absolute/fixed elements
Toolbar is not correctly positioned when parent element has ```position:fixed``` or ```position:absolute```.

Reference: #214

orthes:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | #224, #857
| License          | MIT

### Description

Fix positioning of a toolbar and an anchor preview when `elementsContainer` has `position: absolute` or `position: fixed`.

If `elementsContainer` is `absolute` or `fixed`, it will calculate positions relative to this element instead of window.

I also added a demo `demo/absolute-container.html` which demonstrates this behavior.

### DIFF
--- a/src/js/extensions/anchor-preview.js
+++ b/src/js/extensions/anchor-preview.js
@@ -103,13 +103,15 @@
 
         positionPreview: function (activeAnchor) {
             activeAnchor = activeAnchor || this.activeAnchor;
-            var buttonHeight = this.anchorPreview.offsetHeight,
+            var containerWidth = this.window.innerWidth,
+                buttonHeight = this.anchorPreview.offsetHeight,
                 boundary = activeAnchor.getBoundingClientRect(),
-                middleBoundary = (boundary.left + boundary.right) / 2,
                 diffLeft = this.diffLeft,
                 diffTop = this.diffTop,
-                halfOffsetWidth,
-                defaultLeft;
+                elementsContainer = this.getEditorOption('elementsContainer'),
+                elementsContainerAbsolute = ['absolute', 'fixed'].indexOf(window.getComputedStyle(elementsContainer).getPropertyValue('position')) > -1,
+                relativeBoundary = {},
+                halfOffsetWidth, defaultLeft, middleBoundary, elementsContainerBoundary, top;
 
             halfOffsetWidth = this.anchorPreview.offsetWidth / 2;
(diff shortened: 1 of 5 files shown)


src/js/extensions/toolbar.js
```diff
@@ -616,35 +616,72 @@
...
+
+            middleBoundary = boundary.left + boundary.width / 2;
+            positions.top = boundary.top - toolbarHeight;
+
+            // If container element is absolute / fixed, adjust top position according to container scroll position
+            if (elementsContainerAbsolute) {
+                positions.top += elementsContainer.scrollTop;
+            // If container element isn't absolute / fixed, adjust top position according to window scroll position
+            } else {
+                positions.top += this.window.pageYOffset;
```
nmielnik: Could the calculation of the adjustment of top be done within the `if(elementsContainerAboslute)` block above?

Something like:
```
var topAdjustment = this.window.pageYOffset;

 // If container element is absolute / fixed, recalculate boundaries to be relative to the container
 if (elementsContainerAbsolute) {
    // other existing logic...

    topAdjustment = elementsContainer.scrollTop;
}

middleBoundary = boundary.left + boundary.width / 2;
positions.top = (boundary.top - toolbarHeight) + topAdjustment;
```

That way we can have one less if-condition.

src/js/extensions/anchor-preview.js
```diff
@@ -119,12 +121,41 @@
...
+                relativeBoundary.height = boundary.height;
+                boundary = relativeBoundary;
+
+                containerWidth = elementsContainerBoundary.width;
+            }
+
+            middleBoundary = boundary.left + boundary.width / 2;
+            top = buttonHeight + boundary.top + boundary.height - diffTop - this.anchorPreview.offsetHeight;
+
+            // If container element is absolute / fixed, adjust top position according to container scroll position
```
nmielnik: Same feedback here as in toolbar.js, could we get rid of this extra `if()` block?

src/js/extensions/toolbar.js
orthes: Good point. I'll do that

Issue #1154, opened 2016-07-22
Paste with mouse is getting the Word style


tomasfrancisco:
### Description

When I paste with mouse click, the Word's style is passing to Medium Editor when I specifically configured him to 

```
paste: {
   forcePlainText: true,
   cleanPastedHTML: true
}
```
### Steps to reproduce
1. Paste with right click a formatted string from word, like a text highlighted.

**Expected behavior:** 

Should paste a plain text.
### Versions
- medium-editor: 5.21
- browser: Chrome 51.0.2704.102 (64-bit)
- OS: Mac OSX

nmielnik:
@tomasfrancisco try using `cleanPastedHTML: false`, I think that will ensure that it only use plain-text...

tomasfrancisco:
@nmielnik I already tested with that attribute, but it is still with the same result.

I changed the 

> target.addEventListener(event, listener, false);

 to 

> target.addEventListener(event, listener, false);

and worked, but I don't really know why.

nmielnik:
@tomasfrancisco are the two lines of code you listed above identical?

tomasfrancisco:
@nmielnik Sorry for that, I already edited the post. Could you check again, please?

JeanCarriere:
Up

youssmak:
Up

PR #1163, opened 2016-07-29
add regex to clean font tag provided by word / open office on paste

tomasfrancisco:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | [#1154](https://github.com/yabwe/medium-editor/issues/1154)
| License          | MIT

### Description

Removing font style tag added by editors like Microsoft Word or Open Office when cleanPastedHTML is true.

### DIFF
--- a/src/js/extensions/paste.js
+++ b/src/js/extensions/paste.js
@@ -60,7 +60,10 @@
             [new RegExp(/<\/?o:[a-z]*>/gi), ''],
 
             // Microsoft Word adds some special elements around list items
-            [new RegExp(/<!\[if !supportLists\]>(((?!<!).)*)<!\[endif]\>/gi), '$1']
+            [new RegExp(/<!\[if !supportLists\]>(((?!<!).)*)<!\[endif]\>/gi), '$1'],
+
+            // remove font tag
+            [new RegExp(/<font[^><]*>|<.font[^><]*>/gi), '']
         ];
     }
     /*jslint regexp: false*/


src/js/extensions/paste.js
```diff
@@ -60,7 +60,10 @@
             [new RegExp(/<\/?o:[a-z]*>/gi), ''],
 
             // Microsoft Word adds some special elements around list items
-            [new RegExp(/<!\[if !supportLists\]>(((?!<!).)*)<!\[endif]\>/gi), '$1']
+            [new RegExp(/<!\[if !supportLists\]>(((?!<!).)*)<!\[endif]\>/gi), '$1'],
+
+            // remove font tag
+            [new RegExp(/<font[^><]*>|<.font[^><]*>/gi), '']
```
nmielnik: Are there any attributes on the font tags that help identify it as word / open office?  The problem here is that when copying and pasting from various sources there can be valid `<font>` tags to maintain the font type, size, or color, so I'm not sure we'd want to always remove all `<font>` tags when pasting...

tomasfrancisco: Ok, my main ideia was to remove every formatted strings when paste with cleanPastedHTML.
What I'm trying to do is paste with no color, size or font type associated, just to match the existing style of the web app.
Maybe I'm not understanding what cleanPastedHTML means to you.

nmielnik: So my main concern is that with this PR, it will filter all `<font>` tags from the paste for every user of medium-editor who upgrades.  I could see scenarios where this might not be desired so I'm a bit hesitant to merge this into the main repo as-is (I'm happy to hear feedback from others on this as well though).  

Most of the existing global replacements in the paste handler apply to elements that are very specific side-effects from things like Word or Google Docs.  `<font>` tags could come from all kinds of sources, like copying text from another webpage, so it would be better if there's a way to target just `<font>` tags that come from Word and not all `<font>` tags.  Medium-Editor itself injects `<font>` tags when you a trigger some of the native formatting commands like change font size, change font face, or change font color.  So, if you tried copy and pasting from the editor back into the same editor (or another instance) those font changes would be lost.

Were you aware that you can always pass in your own extra custom RegEx objects for cleaning the pasted html when you initialize MediumEditor?  You can do this by overriding some of the [paste options](https://github.com/yabwe/medium-editor#paste-options).

So for filtering out the `<font>` tags for your instance, you could do something like this:

```js
var editor = new MediumEditor('.editable', {
    paste: {
        cleanPastedHTML: true,
        cleanReplacements: [[new RegExp(/<font[^><]*>|<.font[^><]*>/gi), '']]
    }
});
```

That would ensure that your custom replacements would be executed on each paste as well as when calls to `editor.cleanPaste()` occur.

tomasfrancisco: Yes, I understand your concern. 

For now, I just passed the custom replacement just as you said.

But as I said too, the cleanPastedHTML attribute, in my mind, makes sense that every string's style could be cleaned, I did not expect that string colour was passed to the field.

I'll just drop here the font tag, copied from open office:

`
<font color="#ff3333">String
test</font>
`

There's no way to distinguish from Word or other editor, I guess.

nmielnik: @tomasfrancisco I actually though that cleaning the style off of every string was he purpose of `forcePlainText: true` option.  My interpretation of `cleanPastedHTML` had been to remove unwanted tags, or to try to convert unusual tags into more standard ones.

PR #1176, opened 2016-08-12
Add option to disable tab key.

geoffliu:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| New tests added? | not needed |
| Fixed tickets | none |
| License | MIT |
### Description

Hey all. This is something we need for our current project, don't know if there's any interest in accepting upstream. 

cheers,
Geoff
##

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Using `'fontawesome'` as the buttonLabels requires version 4.1.0 of the fontawes
 * __delay__: time in milliseconds to show the toolbar or anchor tag preview. Default: `0`
 * __disableReturn__:  enables/disables the use of the return-key. You can also set specific element behavior by using setting a data-disable-return attribute. Default: `false`
 * __disableDoubleReturn__:  allows/disallows two (or more) empty new lines. You can also set specific element behavior by using setting a data-disable-double-return attribute. Default: `false`
+* __disableTab__:  enables/disables the use of the tab-key. You can also set specific element behavior by using setting a data-disable-tab attribute. Default: `false`
 * __disableExtraSpaces__:  when set to true, it disallows spaces at the beginning and end of the element. Also it disallows entering 2 consecutive spaces between 2 words. Default: `false`
 * __disableEditing__: enables/disables adding the contenteditable behavior. Useful for using the toolbar with customized buttons/actions. You can also set specific element behavior by using setting a data-disable-editing attribute. Default: `false`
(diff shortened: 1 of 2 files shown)


j0k3r:
Interesting.
Could you fix the style issue and maybe add some tests about that?

geoffliu:
@j0k3r I don't have a whole lot of bandwidth at present, but will add tests when I can.

nmielnik:
@geoffliu correct me if I'm wrong, but if I was summarizing the goal of this PR it would be:

When the user presses <kbd>tab</kbd> allow the browser's default behavior to occur (moving focus to another element?).  Don't indent text within a `<pre>` and don't change indentation level of '<li>` elements.

Is that right?

geoffliu:
@nmielnik Yes that's correct.

nmielnik:
Ok great, makes sense.  I think the name `disableTab` is a bit-misleading, I had thought it would actually prevent tab from doing anything and it wasn't until I read the code that I realized what it was doing.

Could you rename this to something like 'defaultTabBehavior' (default value is `false`)?  Something to say that the editor won't do anything when the user pushes <kbd>tab</kbd> but the default browser behavior will still happen?

And like @j0k3r mentioned, adding some tests will also help clarify what should happen so that make sure this feature is maintained and not broken in the future.

Thanks for this!

geoffliu:
Ah I see. I was just mirroring the naming of `disableReturn`  and `disableDoubleReturn`. Do we not want to keep them consistent? 

And yes, absolutely agree with tests, will do that as soon as I get the bandwidth.

nmielnik:
`disableReturn`, `disableDoubleReturn`, `disableExtraSpaces`, and `disableEditing` are all options that attempt to disable some functionality from the browser.  In this case, the editor has code that will attempt to add on top of the browser's functionality (doing some indentation).  

So, I think of the option you're adding as a bit different, since it will be turning off some extra code that is added by medium-editor, instead of turning off something the browser does by default.  But, that's just my opinion.

nmielnik:
@geoffliu another option to represent this feature is by creating a `tabIndent` option.  This option would be true by default, but if you set it to false then you get the behavior that you've created in this PR. Just another way to think about it.

mkawczynski07:
@geoffliu if u want "disable tab key":

```    
   this.mediumEditor.subscribe('editableKeydownTab', e => {
      e.preventDefault();
      e.stopPropagation();
    });
```

PR #1187, opened 2016-08-19
remove first empty paragraph on backspace

orthes:
| Q                | A
| ---------------- | ---
| Bug fix?         | more or less
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | -
| License          | MIT

### Description

Most users don't know the difference between backspace and delete. So if they create a paragraph as the first element in the editor and they want to delete it, they place caret to the paragraph and press backspace. Which does nothing. In order to delete that paragraph you need to press delete / fn+backspace on mac, or move caret to a beginning of next paragraph and press backspace there.

I've seen people struggling with this, so this PR solves this problem. If a caret is in the first and empty paragraph and you press backspace, it will do delete instead - it will delete the empty paragraph and move the caret to next one.

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -652,6 +652,22 @@ describe('Content TestCase', function () {
                 expect(this.el.innerHTML).toBe('<p>lorem ipsum</p><ul><li></li><li>lorem ipsum</li></ul>');
             });
         });
+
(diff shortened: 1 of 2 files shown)


j0k3r:
Crazy people.

nmielnik:
😞 (sigh)

PR #1203, opened 2016-09-07
Check for this.anchorPreview when hiding

websirnik:
Bug fix?  yes
New feature? no
BC breaks? no
Deprecations? no
New tests added? not needed
License MIT

This fixes the following error:

```
TypeError: Cannot read property 'classList' of undefined
File .../vendor-edit-mode.js line x col xxxxx in hidePreview
File .../vendor-edit-mode.js line x col xxxxx in detachPreviewHandlers
File .../vendor-edit-mode.js line x col xxxxx in updatePreview
```

I have this come up in my production app on Windows 10 Chrome 52.

### DIFF
--- a/src/js/extensions/anchor-preview.js
+++ b/src/js/extensions/anchor-preview.js
@@ -71,7 +71,9 @@
         },
 
         hidePreview: function () {
-            this.anchorPreview.classList.remove('medium-editor-anchor-preview-active');
+            if(this.anchorPreview){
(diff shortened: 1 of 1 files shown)


j0k3r:
Thanks for that.
Could you please fix the CS issue?

reco:
have the same problem. anybody on this?

slorber:
same here, is this fixed in newer releases?

Issue #1166, opened 2016-08-03
Keyboard focus is lost after pasting text (Firefox)


dcsaszar:
### Description

The editable area loses the (visible) input cursor after pasting text in Firefox.
Chrome is fine.

### Steps to reproduce

1. Use FF (47.0.1)
2. visit https://yabwe.github.io/medium-editor/
3. click the editable text
4. paste something (using ctrl+v)

**Expected behavior:** [What you expected to happen]

* the input cursor blinks after the pasted text
* keyboard text input is still possible

**Actual behavior:** [What actually happened]

* no input cursor
* typing is not possible
* pasting again still pastes correctly

### Versions

- medium-editor: 5.21.0
- browser: Firefox
- OS: Ubuntu

biggieman:
+1

nmielnik:
This could definitely be related to #1156 , so if anyone investigates fixing either issue please try to investigate both at the same time.

dcsaszar:
5087f4b84206aa7c76fe91204bff26f094dc1cb9 is the last working commit
after that the dist is broken for a while (7 commits in the history of https://github.com/yabwe/medium-editor/commits/master/src/js/extensions/paste.js)
3cc4165da287a4987a1e4807f4d24581dcffc698 has a working dist and shows the issue

On of the causes appears to be setting the `.focus()` to the pasteBin. If I comment out this line, it works.

dcsaszar:
I don't know why we need the `focus()` on the paste bin. So, currently I see two possible solutions:

* don't focus the paste bin at all (https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/paste.js#L315-L320)
* put the paste bin div into the editable element instead of the body (https://github.com/yabwe/medium-editor/blob/master/src/js/extensions/paste.js#L313 => sth like `this.getEditorElements()[0].appendChild(pasteBinElm);`)

j0k3r:
Focus is needed so the content you just pasted is written into the paste bin. Otherwise the whole pastebin shouldn't work.
I think we should better put the focus back to the editable element when pasting is done.

dcsaszar:
How would I know the pastebin isn't working? I tried deleting the focus stuff and sucessfully pasted in the default index.html in Firefox and Chrome.

j0k3r:
Well, if what you pasted was actually pasted, it means the whole paste stuff is working.
Which looks strange without the focus :neutral_face:

dcsaszar:
[...]

PR #1213, opened 2016-09-14
Fix paste blur

dcsaszar:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | What's BC?
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | #1166
| License          | MIT

### Description

Fixes Firefox lost input focus after paste

--

### DIFF
--- a/spec/paste.spec.js
+++ b/spec/paste.spec.js
@@ -562,6 +562,42 @@ describe('Pasting content', function () {
             contentEditables = document.body.querySelectorAll('[contentEditable=true]');
             expect(contentEditables.length).toBe(1);
         });
+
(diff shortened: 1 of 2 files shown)


j0k3r:
@nmielnik could you have a look?

dcsaszar:
I discovered why the `focus` is needed. Without it pasting text copied from outside of the editable element fails in Chrome. So we need a better solution. Closing for now.

dcsaszar:
Reopening, since I cannot reproduce what I've said about Chrome in my previous comment.

Issue #1216, opened 2016-09-19
CustomClassOption is not added when anchoring complex selected styled text


annayafi:
### Description

If you have a simple text that you wanna anchor, it works.
If you have a complex text, it's another story. Let's say something like :
`<b>this is a</b><i><b> complicated</b> <strike><b>markup</b></strike></i>`
If you have specified a button class to then function createLink, it won't be taken into account because it's going to handle the wrong node. The problem is the addClassToAnchors function, which won't traverse up to the parents node : 

```
if (el.nodeName.toLowerCase() === 'a') {
                for (j = 0; j < classes.length; j += 1) {
                    el.classList.add(classes[j]);
                }
            } else {
                **el = el.getElementsByTagName('a');**
                for (i = 0; i < el.length; i += 1) {
                    for (j = 0; j < classes.length; j += 1) {
                        el[i].classList.add(classes[j]);
                    }
                }
            }
```
### Steps to reproduce
1. Add style to your text
2. Select some text crossing several styles, for example select : " is a complicated mar".
3. Call createLink on this selection, specifying a buttonClass option. 

**Expected behavior:** The button class you specified is added to the created a element.
**Actual behavior:** The button class is not added to the created a element.

**Link to an example:** https://jsfiddle.net/m63wpj5a/
![edium editor a bug - jsfiddle20160919132453](https://cloud.githubusercontent.com/assets/8353486/18630392/9f632e10-7e6c-11e6-9bcd-6c9dccfa5703.gif)

Or even more simply:
![edium editor a bug - jsfiddle20160919135501](https://cloud.githubusercontent.com/assets/8353486/18631218/f02bec0c-7e70-11e6-8020-4b052a0b1cb4.gif)
### Versions
- medium-editor: current
- browser:chrome (but any will do)

PR #1217, opened 2016-09-19
#1216 issue fix

annayafi:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| New tests added? | yes : anchor.spec.js, l.495 |
| Fixed tickets | #1216 |
| License | MIT |
### Description

The problem was : when you call addClassToAnchors(el, buttonClass) function : 
- if el is an `<a>` : no problem, it adds the css class(es)
- if el has `<a>` children : no problem, it adds the css class(es) to children
- if el is a child of an `<a>` : the function never looks for parent and won't add the css class(es). 
  This happens very easily when you select more than a node, for example you select exactly `<i>hello</i>`, css is never added.


dist/js/medium-editor.js
```diff
@@ -1020,6 +1020,11 @@ MediumEditor.extensions = {};
             }
         },
 
+        /*
```
j0k3r: Could you please remove changes on `dist/*` files?

j0k3r: Yeah not really because we still need them to push a new release.
That's why we put a text in the [PR template](https://github.com/yabwe/medium-editor/blob/master/.github/PULL_REQUEST_TEMPLATE.md#please-dont-submit-dist-files-with-your-pr) (which no one read) saying: 

**Please, don't submit `/dist` files with your PR!**

annayafi: As for the test file, yes I don't know why the automatic indentation of my editor must have shifted lines which had nothing to do with my changes. I changed the test itself to be more relevant, specific and simple. Thank you for your review.

annayafi: Ok for dist files ! sorry then for submitting them previously.

nmielnik:
This is getting merged as part of #1217 (Thanks again for the fix @annayafi )

PR #1223, opened 2016-09-29
fix #1221 commonJS/electron environment detection

### LINKED ISSUE #1221: Build breaks in a CommonJS environment
The dist breaks any build which has a `module` object defined. This object is always defined in a CommonJS environment. The error I get is:

```
Uncaught ReferenceError: process is not defined
```

This error results from this check:

```javascript
var isElectron = typeof module === 'object' && process && process.versions && process.versions.electron;
```

`module` is an object because we have CommonJS but process is undefined because we are neither in node.js, nor in electron.

This issue is related to #1123.

paradite:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | no, not sure if needed
| Fixed tickets    | #1221
| License          | MIT

### Description

Fix detection of commonJS environment without electron as described in the comments in #1221

Please advise on if refactoring or new test cases are needed.

### DIFF
--- a/src/wrappers/start.js
+++ b/src/wrappers/start.js
@@ -1,6 +1,6 @@
 (function (root, factory) {
     'use strict';
-    var isElectron = typeof module === 'object' && process && process.versions && process.versions.electron;
+    var isElectron = typeof module === 'object' && typeof process !== 'undefined' && process && process.versions && process.versions.electron;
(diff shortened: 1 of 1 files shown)


j0k3r:
Looks good to me :+1:

Issue #1209, opened 2016-09-09
Cursor position at end of line giving entire P tag instead of Last span Tag


durgeshahire4:
### Description

Consider user enter 4-5 words and each word have different font-style, when user click at last cursor position at that time it will give default font-style of `P` tag not the last word font-style (means last `span` of `P` Tag.

### Steps to reproduce

1. Enter 4-5 words and each have different font-style (including font-size)
2. Now click at last cursor position after where the last word is completed
3. You can see it will select default state of font-style in editor . it will not preserve the last word font-style

**Expected behavior:** [What you expected to happen]
 Cursor at end of text should preserve font-style of last word

**Actual behavior:** [What actually happened]
 Cursor not preserve last word font-style. It select default font-style of `P` tag.

### Versions

- medium-editor:
- browser:
- OS:

j0k3r:
Could you provide a jsfiddle with the _4-5 words_ and explain what we have to do (like put cursor after word "surprise". Thanks

ravi-dadhaniya:
[NoteBook_Cursor_Issue.swf.zip](https://github.com/yabwe/medium-editor/files/464059/NoteBook_Cursor_Issue.swf.zip)


Attaching Zip file which contain video of issue we are facing. please open that video file in chrome browser.

As in video you can see when i clicked on ‘Hello’ word 'font-size' & 'type' get changed properly same as with word 'India'. But when i clicked after word 'India' (means next position) then its font-size and 'font-type' get reset to default (type: open Sans and size : 14). It should remain same as word ‘India’ .

PR #1224, opened 2016-09-30
Fixes #1208 and #1209

### LINKED ISSUE #1208: editableClick and editableKeyup is giving different event.target
### Description

editableClick and editableKeyup gives me different event targets, and editableKeyup needs to provide the proper event targets rather giving the editor instance.
### Steps to reproduce
1. Open the editor and type few lines  with different font styles for each lines
2. Using mouse, click on any one of these lines listen to editableClick and check the target font style. You will see proper styles with jQuery `$(event.target).css('font-family')` 
3. Now using keyboard move between lines, listen to editableKeyup and check the target, you will see editor instance as target. It should provide me the same target like editableClick.

**Expected behavior:** [What you expected to happen]
editableKeyup should give target as it editableClick does.

**Actual behavior:** [What actually happened]
editableClick and editableKeyup is giving different event.target
### Versions
- medium-editor:
- browser:
- OS:

sajus:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | comma-separated list of tickets fixed by the PR, if any 1208, 1209
| License          | MIT

### Description

[Description of the bug or feature]

--

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -148,6 +148,12 @@
             }
         },
 
+        getComputedStyle: function () {
+            var range = document.getSelection().getRangeAt(0),
+                element = range.commonAncestorContainer.parentNode;
+            return window.getComputedStyle(element, null);
+        },
+
         // Cleaning up
 
         destroy: function () {
@@ -516,6 +522,7 @@
         },
 
         handleClick: function (event) {
+            event.computedStyle = this.getComputedStyle();
             this.triggerCustomEvent('editableClick', event, event.currentTarget);
         },
 
@@ -540,6 +547,7 @@
         },
 
         handleKeyup: function (event) {
+            event.computedStyle = this.getComputedStyle();
             this.triggerCustomEvent('editableKeyup', event, event.currentTarget);
         },
 


sajus:
@j0k3r , this PR will fix #1208 and #1209 , can you please check this out?

durgeshahire4:
@j0k3r Can you please review this one?
Thanks!

j0k3r:
I've no idea what is that `event.computedStyle`

sajus:
@j0k3r, I'm capturing the style from the selection and sending this as payload in ```event``` object. ```event.computedStyle``` is custom variable attached to ```event``` object. We need to bubble this up for the mouse clicks and key ups, for which we have listeners already attached. 

Hope this answers your question.

j0k3r:
But where `event.computedStyle` is re-used after being assigned?

sajus:
Yeah that's good question! Let me explain that for you, In our implementation we subscribe for ```editableKeyup``` and ```editableClick``` from the editor object. 

```editor.subscribe('editableKeyup', function(event) {
          event.preventDefault();
          console.log(event.computedStyle.getPropertyValue('font-size'));
          console.log(event.computedStyle.getPropertyValue('font-family'));
          console.log(event.computedStyle.getPropertyValue('color'));
        });
``` 

With the ```event.computedStyle``` I'm able to get exact font styles which is impossible for the ```key events``` 

You can refer more on this from my comments on https://github.com/yabwe/medium-editor/issues/1208

j0k3r:
As far as I understand this is a custom fix for an edge case on your side only.
Then it can't be merged into MediumEditor.

Is there any other alternative?

sajus:
I'm afraid that there aren't any alternative here, if we are using ```contenteditable``` as true then it doesn't drill down to child node for targeting. And for the edge case click, the target doesn't matches. Only through selection object we can resolve these issues. 

Thats the reason why I have created generic function ```getComputedStyle``` which could be used else where in the MediumEditor.  In case if some one else also had faced the same issues.

j0k3r:
I can understand your concern.
Maybe an other validation from @nmielnik / @orthes is necessary.

sajus:
okies, @nmielnik / @orthes  could you please have a look into this?

chriskwright:
@j0k3r We're trying to get our project wrapped up this week. There are 4 bugs that are tied to this PR that @sajus submitted. Can you please let us know if there is code we need to change to fix the bugs on our end in the plugin we're using or if the PR is the best and approve? Your help is very appreciated. Thanks

PR #1225, opened 2016-09-30
Fixes an issue where while a user is holding down delete and clicks around both in and out of the content editable the node is null.

Garvice:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | yes/no |
| Deprecations? | no |
| New tests added? | not needed - Not even sure how to create a test for the behavior I'm experiencing. |
| Fixed tickets |  |
| License | MIT |
### Description

[Description of the bug or feature]
## 
#### Please, don't submit `/dist` files with your PR!

…round, both in and out of the contenteditable, the node can be null

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -52,8 +52,13 @@
     }
 
     function handleBlockDeleteKeydowns(event) {
-        var p, node = MediumEditor.selection.getSelectionStart(this.options.ownerDocument),
-            tagName = node.nodeName.toLowerCase(),
+        var p, node = MediumEditor.selection.getSelectionStart(this.options.ownerDocument);
+
+        if (!node) {
+            return;
+        }
+
+        var tagName = node.nodeName.toLowerCase(),
             isEmpty = /^(\s+|<br\/?>)?$/i,
             isHeader = /h\d/i;
 


j0k3r:
Hey thanks for that.
Any changes you add some tests for that?

Garvice:
For sure. Let me play around and see if I can simulate this behavior somehow.

j0k3r:
Did you succeed @Garvice ?

nmielnik:
@Garvice I've been experimenting with our demo locally and I really can't seem to reproduce this issue.  

I found a way to ensure I can click somewhere and cause the selection to be empty, which would ensure that the call to `MediumEditor.selection.getSelectionStart(this.options.ownerDocument)` would return null (which you fix would be addressing).  I put the cursor inside the editor, hold down the delete key, and then continue to click in and out of the editor, including clicking in a spot that would ensure that the selection doesn't have a valid `anchorNode`, but the error just won't happen.  I can't figure out a way that the `handleBlockDeleteKeydowns` event could possibly fire with the selection being empty unless there is some external code that would be mucking around with the timing of this event, or clearing the selection out manually before the event handler can run.

If you can provide some way to reproduce this consistently (steps in the demo, a JSBin, a jsfiddle, a codepen, etc.) I can try to help create a test so that we can merge this in...but if there isn't even a way to reproduce the issue manually, I don't think we can merge in a code change without a test.

PR #1233, opened 2016-10-16
Removed obsolete code

xErik:
Double negation does nothing.


j0k3r:
It does something: force returned value to be a boolean.

PR #1235, opened 2016-10-19
Eval html toolbutton

vendelin8:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | 
| License          | MIT

### Description

A new tool button for evaluating inserted html, basically for embedding videos or objects.

--

### DIFF
--- a/dist/js/medium-editor.js
+++ b/dist/js/medium-editor.js
@@ -3350,14 +3350,6 @@ MediumEditor.extensions = {};
             contentDefault: '<b>image</b>',
             contentFA: '<i class="fa fa-picture-o"></i>'
         },
-        'eval': {
-            name: 'eval',
-            action: 'eval',
-            aria: 'evaluate html',
-            tagNames: ['iframe', 'object'],
-            contentDefault: '<b>eval</b>',
-            contentFA: '<i class="fa fa-play-circle"></i>'
-        },
         'orderedlist': {
             name: 'orderedlist',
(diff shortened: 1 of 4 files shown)


nmielnik:
@vendelin8 would you mind just calling this 'html' instead of 'eval'?  Inserting raw html is a pretty common concept for text-editors, so I think it would be pretty intuitive what 'html' would mean.

My guess is that developer would have one set of expectations for what 'eval' means (scary javascript) and regular non-coder users would be perplexed as to what 'eval' could mean...

Also, there are 2 other places to mention your new button:
1. In the [README](https://github.com/yabwe/medium-editor/blob/master/README.md#all-buttons)
2. In the [extensions README](https://github.com/yabwe/medium-editor/tree/master/src/js/extensions#all-of-the-built-in-mediumeditor-buttons-are-just-button-extensions-with-different-configuration)

This is a nice and simple idea, I like it.  Sorry for your PR sitting for such a crazy long time, my attempt at recruiting more collaborators failed before I had to get pulled into a bunch of other projects.

vendelin8:
@nmielnik I don't mind at all, go for it!
Thanks for letting it happen, and for all the good stuff so far!  :)

nmielnik:
Merging this work in #1291 to be completed there.  Thanks again @vendelin8

PR #1239, opened 2016-10-21
Fix-autolink Fixed problem of moving cursor when auto-linking

EmilioMN92:


j0k3r:
Thanks for that. But.

- as describing in [the PR template](https://github.com/yabwe/medium-editor/blob/master/.github/PULL_REQUEST_TEMPLATE.md) (the one you completely removed when opening your PR) you should not **submit `/dist` files with your PR!**
- the PR template help us to understand what you are trying to fix, without it you force us to understand the code you wrote. Which means we are loosing some of our free time for that, which could be spend on other more important things.
- by avoiding writing description in your PR, you are kind of saying _I pushed that stuff could you merge it ASAP OKTHXBYE_
- for this kind of unexplicated changes, it still important for us to write unit test. You're probably not going to maintain the code you pushed in that repo, we do. Unit test help us in that way.

Could you please fix these points?

j0k3r:
This issue is fairly old and there hasn't been much activity on it. Closing, but please re-open if it still occurs.

Issue #1250, opened 2016-11-25
linkValidation bug in IE?


LuoPQ:
### Steps to reproduce

1. Select some texts
2. Click anchor button, then enter a '/' as a link.
3. preview then click the link, the link will be "http:///", and can't open a new tab even set the target="_blank".

**Expected behavior:** [What you expected to happen]
when enter "/" as link, it should be can't add "http://" before it, or open a new tab for this wrong link in IE.

**Actual behavior:** [What actually happened]
open the wrong link in current tab.

### Versions

- medium-editor: 5.22.1
- browser:  IE11
- OS:window 7

PR #1258, opened 2016-12-10
Only add schemes to URLs with hostnames

kevindew:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | yes
| Fixed tickets    | 1250
| License          | MIT

### Description

Fixes: https://github.com/yabwe/medium-editor/issues/1250

This changes the way a scheme is prefixed to a URL when linkValidation
is on. Previously any URL that did not match the scheme regex would be
prefixed with 'http://' which meant that an absolute path such as '/'
would be changed to 'http:///'

This changes it so that the 'http://' is only prefixed if the first part
of the path looks like a hostname, which allows absolute paths to be
unchanged and most relative paths to be unchanged.

I say most as a path such as "test.txt" would be matched as a hostname and
converted to 'http://test.txt' but this seems an acceptable trade off
and definitely better than trying to store a list of valid TLDs so
filenames and hostnames can be determined.

### DIFF
--- a/spec/anchor.spec.js
+++ b/spec/anchor.spec.js
@@ -356,6 +356,63 @@ describe('Anchor Button TestCase', function () {
             expect(link.getAttribute('href')).toBe(validHashLink);
         });
 
+        it('should not add a scheme to an absolute path', function () {
+            var editor = this.newMediumEditor('.editor', {
+                    anchor: {
+                        linkValidation: true
+                    }
+                }),
+                absolutePath = '/test',
+                link,
+                anchorExtension = editor.getExtensionByName('anchor');
+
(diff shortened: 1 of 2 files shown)


nmielnik:
@kevindew looks like this conflicts with your previous change that got merged in.  Could you fix the merge conflict and then we can get this guy in too?

kevindew:
@nmielnik sure, I've rebased it against master now.

nmielnik:
Awesome, all the tests pass in saucelabs, thanks for this @kevindew !

PR #1359, opened 2017-08-11
Change mouse cursor for placeholder to I-beam

gvx:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | 
| License          | MIT

### Description

Change mouse cursor for placeholder from default to I-beam

### DIFF
--- a/src/sass/components/_placeholder.scss
+++ b/src/sass/components/_placeholder.scss
@@ -3,6 +3,7 @@
 
     &:after {
         content: attr(data-placeholder) !important;
+        cursor: text;
         font-style: italic;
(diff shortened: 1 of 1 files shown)


j0k3r:
Why?

gvx:
IMO, the cursor changing to the default cursor when you're hovering the placeholder is unintuitive. Especially since when doing the same for regular `<input>` or `<textarea>` elements, the mouse cursor stays an I-beam.

PR #1361, opened 2017-08-15
medium-editor-insert plugin css fixes on beagle theme

mnussbaumer:
Added .scss and dist/.css changes for beagle theme to accomodate ´med…ium-editor-insert-plugins´ buttons and toolbar - On the original one it would change to a square button on hover, and the last button wouldn't show correctly either

| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

When using medium-insert-plugin directly with beagle theme on medium-editor, buttons on the toolbar of medium-insert-plugin and the delete button (both appear when an image has been added to the dom for upload) would change border-radius on hover, showing squares and obviously looking wrong. 

--


j0k3r:
Could you provide a jsfiddle reproducing the bug?
Thanks.

mnussbaumer:
@j0k3r do you have any working fiddle for medium-editor alone? 
I've tried setting JS to be jQuery, added medium and stylesheets as external resources but it doesn't seem to be able to load them? And then jquery upload doesn't really have a CDN dist, nor does medium-insert.

I have screenshots, but no really sure how I can go on showing this on a fiddle.
Thanks

j0k3r:
From a previous issue: https://jsfiddle.net/4xdom6ar/

mnussbaumer:
Thank you @j0k3r,
https://jsfiddle.net/4fzwcftx/1/

j0k3r:
Could you please explain step by step what should I do to reproduce the bug using your jsfiddle?

mnussbaumer:
@j0k3r yes, click on the "Lorem ipsum..." editable content. Go to the end of the line, hit carriage return to go to a new line. It will show you the ´+´ sign from the insert plugin. Click the plus, it will show you another 2 circles, click the first one to upload a picture. Choose a picture from your local drive, select. It will display the picture on the editor area. Now click the picture itself to show the plugin's toolbar & the ´x´ icon on it. Hover over the ´x´ icon and the right-most item on the plugin toolbar, you'll see the border-radius change between them.

mnussbaumer:
And the plugin isn't attributing classes for the first and last item in the nav bar so I might as well do a pr on the plugin's repo to fix that, still, the ´x´ close icon isn't on an ordered list so it will not be fixed by that.

src/sass/themes/beagle.scss
```diff
@@ -112,3 +112,9 @@ $medium-editor-placeholder-color: #f8f5f3;
     text-decoration: none;
   }
 }
+
+.medium-editor-toolbar-actions {
+  li, button {
+    border-radius: 50px;
```
j0k3r: You shouldn't use the hardcoded value `50px` but `$medium-editor-border-radius` instead

PR #1362, opened 2017-08-17
fix crash when 'extensions' in 'isElementDescendantOfExtension' is undefined

hankmander:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| License          | MIT

### Description

`extensions` can under some circumstances be `undefined`. This causes an error when calling `extensions.some()` and for example stops `execAction` from having effect. Not sure if this is a good way to solve the issue, but it prevents the error.

### DIFF
--- a/src/js/events.js
+++ b/src/js/events.js
@@ -2,7 +2,10 @@
     'use strict';
 
     function isElementDescendantOfExtension(extensions, element) {
-        if ( !extensions ) return false;
+        if (!extensions) {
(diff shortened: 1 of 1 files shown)


paltman:
Would love to see this merged.

j0k3r:
Ok build is dead.
Will fix it later.
Thanks @hankmander (I've no time to investigate a better solution for now)

PR #1364, opened 2017-08-25
Firefox createLink bug workaround

apptaro:
This fix workarounds FireFox createLink bug.
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=451142

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -1149,6 +1149,7 @@
                             this.importSelection(exportedSelection);
                         } else {
                             this.options.ownerDocument.execCommand('createLink', false, targetUrl);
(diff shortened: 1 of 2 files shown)


apptaro:
Without this fix and if you use Firefox, you can't create a link to url like http://www.städtlifest.ch/

j0k3r:
Could you add a test for that?

apptaro:
I am not very familiar with jasmine testing framework that this project uses. Can I write firefox specific testing code? It seems to me that jasmine uses phantomjs which is webkit based.

j0k3r:
Just write your test code to validate your change is ok, I guess it shouldn't introduce side behavior on other browser.
The default build use phantomjs but when merged to master we use Saucelabs to run against many different browsers

apptaro:
I added the test code.

PR #1396, opened 2017-12-19
Fix broken browser tests

nchase:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| License          | MIT

### Description

See individual commit messages for details.

At a high level, I think these tests broke because the browsers decided to slightly change the markup they generate for `contenteditable` elements under various conditions.

This changeset finds a couple of places where this has been a problem and normalizes the output to what the tests expect. I don't think either of the normalizations that I'm performing is dangerous (i.e. potentially compromises the integrity of the tests and what they're meant to ensure), but please let me know if this is actually wrong.

### DIFF
--- a/spec/content.spec.js
+++ b/spec/content.spec.js
@@ -649,8 +649,7 @@ describe('Content TestCase', function () {
                 fireEvent(target, 'keydown', {
                     keyCode: MediumEditor.util.keyCode.BACKSPACE
                 });
-                this.el.innerHTML = stripLinebreak(this.el);
-                expect(this.el.innerHTML).toBe('<p>lorem ipsum</p><ul><li></li><li>lorem ipsum</li></ul>');
+                expect(this.el.innerHTML).toMatch(/^<p>lorem ipsum<\/p><ul><li>(<br>)?<\/li><li>lorem ipsum<\/li><\/ul>$/);
             });
         });
 
@@ -812,7 +811,3 @@ describe('Content TestCase', function () {
         });
     });
 });
-
-function stripLinebreak(element) {
-    return element.innerHTML.replace('<br>', '');
-}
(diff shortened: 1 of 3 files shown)


nchase:
PS: I don't yet know if the Edge test failure is fixed. I couldn't get a read on what exactly was failing in Edge.

nchase:
Looks like edge just doesn't connect to the test runner URL at all.

nchase:
after some digging around on the web, I fixed the edge issue by adding 358fcd9.

nchase:
@j0k3r does this look good to you?

spec/content.spec.js
```diff
@@ -649,6 +649,7 @@ describe('Content TestCase', function () {
                 fireEvent(target, 'keydown', {
                     keyCode: MediumEditor.util.keyCode.BACKSPACE
                 });
+                this.el.innerHTML = stripLinebreak(this.el);
                 expect(this.el.innerHTML).toBe('<p>lorem ipsum</p><ul><li></li><li>lorem ipsum</li></ul>');
```
nmielnik: could you use a `matches` test here instead of `toBe`?  that will let you use a regex and not have to modify the actual element for the comparison.  I've use the regex thing in a bunch of tests in the repo.

nmielnik: Sorry, i meant `toMatch`, put too much trust in my memory.  Here's an [example](https://github.com/yabwe/medium-editor/blob/5.23.2/spec/core-api.spec.js#L158-L159)

j0k3r:
This can't be true or maybe it's already christmas? Both Noah and Nate are still alive on the same day. Amazing :cowboy_hat_face:

PR #1436, opened 2018-04-12
Cleanup after ordered/unordered list removal in Chrome

alexkurowski:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | no
| License          | MIT

### Description

Currently in Chrome when removing ordered/unordered list you get a `<span>` element followed by a `<br>`  for every line instead of just `<p>` elements. Furthermore, when inserting a line break in the middle of a span like this, everything that goes after gets wrapped in a div, which is unwanted.

This fix simply goes through these spans and replaces them with proper `<p>` elements.

### DIFF
--- a/src/js/util.js
+++ b/src/js/util.js
@@ -675,17 +675,60 @@
 
         cleanListDOM: function (ownerDocument, element) {
             if (element.nodeName.toLowerCase() !== 'li') {
-                return;
-            }
+                if (this.isIE || this.isEdge) {
+                    return;
+                }
 
-            var list = element.parentElement;
+                var selection = ownerDocument.getSelection(),
+                    newRange = ownerDocument.createRange(),
+                    oldRange = selection.getRangeAt(0),
+                    startContainer = oldRange.startContainer,
+                    startOffset = oldRange.startOffset,
+                    endContainer = oldRange.endContainer,
+                    endOffset = oldRange.endOffset,
+                    node, newNode, nextNode;
 
-            if (list.parentElement.nodeName.toLowerCase() === 'p') { // yes we need to clean up
-                Util.unwrap(list.parentElement, ownerDocument);
+                if (element.nodeName.toLowerCase() === 'span') {
+                    // Chrome & Safari unwraps removed li elements into a span
+                    node = element;
+                } else {
+                    // FF leaves them as text nodes
+                    node = startContainer;
+                }
 
-                // move cursor at the end of the text inside the list
-                // for some unknown reason, the cursor is moved to end of the "visual" line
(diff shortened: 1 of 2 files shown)


alexkurowski:
@j0k3r  Turns out this issue exists on firefox as well, but instead of spans, unwrapped `<li>` elements are added back as text nodes without a parent element.

Added test case, if ran without this fix, will fail with a following message: 
```
Expected 'lorem<br><ul><li>ipsum</li><li>dolor</li></ul>' to be '<p>lorem</p><ul><li>ipsum</li><li>dolor</li></ul>'.
```

Notice stranded `lorem<br>`.

src/js/util.js
```diff
@@ -673,19 +673,58 @@
             return false;
         },
 
-        cleanListDOM: function (ownerDocument, element) {
+        cleanListDOM: function (contentWindow, ownerDocument, element) {
```
j0k3r: This is a breaking change as you are changing the method signature. Could you better put that new parameter at the end and handle the case where that parameter won't be given?

src/js/util.js
```diff
@@ -673,19 +673,58 @@
...
+                    startOffset = oldRange.startOffset,
+                    endContainer = oldRange.endContainer,
+                    endOffset = oldRange.endOffset,
+                    node, newNode, nextNode;
+
+                if (element.nodeName.toLowerCase() === 'span') {
+                    // Chrome unwraps removed li elements into a span
+                    node = element;
+                } else {
+                    // FF leaves them as text nodes
```
j0k3r: What about Edge and Safari?

j0k3r:
Looks ok, could you squash your commits?
Then I'll merge it and we'll see what Saucelabs says for all supported browser 🙂

j0k3r:
There are failing tests:
- Platform: WIN8.1,firefox,
Passed: false
Url https://saucelabs.com/jobs/e30145ecc0f143c69edccdea52872059

- Platform: Windows 10,firefox,
Passed: false
Url https://saucelabs.com/jobs/aa3563cd793d450bbee2e0206351f576

- Platform: OS X 10.11,firefox,
Passed: false
Url https://saucelabs.com/jobs/bd9488d7887a44c4aaa8aa8d549f4065

- Platform: Windows 10,MicrosoftEdge,
Passed: false
Url https://saucelabs.com/jobs/b701daba68564fd2a36dc9ea3080cf36

You can then view them by going on the metadata tab and screenshots at the bottom of that tab.

MichalKrakow:
If li element has an <a> node it seperates it from rest of markup.
If all is good it wraps text in double paragraph.

Issue #1496, opened 2019-03-08
Using arrow keys to deselect text after applying style fails in Firefox


peterchappell:
### Description

In Firefox, if you select some text, apply a style via the toolbar and then use the left or right arrow key to de-select the text, nothing happens.

### Steps to reproduce

1. Use Firefox
2. In an editor, select some text
3. Apply style/formatting via the toolbar
4. After the style has been applied, use an arrow key to de-select the text

(Compare against other browsers)

**Expected behavior:** 

The text should be deselected and the cursor should be positioned accordingly (e.g. after applying a style and then clicking the "right" arrow the cursor should be to the right of what was previously selected.

**Actual behavior:** 

Nothing happens.

**Link to an example:** 

This can be reproduced on the [demo page](https://yabwe.github.io/medium-editor/demo.html) (using Firefox)

### Versions

- medium-editor: 5.23.3
- browser: Firefox 65.0.1 (64-bit)
- OS: MacOS (Mojave)

PR #1497, opened 2019-03-08
Ensure the focus stays in the editor after execAction for Firefox. #1496

peterchappell:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | no
| Fixed tickets    | #1496
| License          | MIT

### Description

For some reason in Firefox, the body is the focussed element after exec action. If there's a nicer way to do this (or to _stop_ the body being the focus) let me know and I'll be happy to investigate further.

### DIFF
--- a/src/js/core.js
+++ b/src/js/core.js
@@ -942,6 +942,12 @@
                 MediumEditor.util.cleanListDOM(this.options.ownerDocument, this.getSelectedParentElement());
             }
 
+            // https://github.com/yabwe/medium-editor/issues/1496
+            // ensure the focus remains in the editor for Firefox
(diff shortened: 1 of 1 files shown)


reneklacan:
@peterchappell 

Hello, thank you for this change, is there any downside of calling `.focus()` all the time, not just for FF?

Because it seems to be fixing some buggy behaviour I'm seeing on Chrome, specifically when you click on toolbar (with `static: true, updateOnEmptySelection: true`) button without any text selected.

Without `.focus()` editing area loses focus.

PR #1550, opened 2020-07-14
Dependency updates, minor fixes and resolve CI issues (build failing)

arteconceito:
| Q                | A
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| New tests added? | not needed, some fixed
| Fixed tickets    | none
| License          | MIT

### Description

Update dependency packages to latest versions possible to reduce vulnerabilities and make it possible to run the project on newer Node versions (up to v12 LTS). Make some minor adjustments to avoid JSHint warnings. Fix two tests that were failing (also for CI).

--


sauravkhdoolia:
@arteconceito What do you think, why CI build was failing?
Also, How did you find out about that Edge use Case!!

And thank you so much! That was so darn fast.

arteconceito:
@sauravkhdoolia the Edge issue was mentioned by @stefanpejcic on Jul 23, 2019. He adapted the code for Edge but forgot to still support other browsers that apply `<strike>` instead of `<s>`. Because of this, two tests were failing and that was causing the CI build to fail.

sauravkhdoolia:
@arteconceito You see SauceLabs fails the tests, I merged it for some reason but...

We've to get that SauceLabs thing to work somehow.

arteconceito:
@sauravkhdoolia I am already on it. It is due to the fact that the SouceLabs pipeline is setup to use Node v8, I will change it to the latest LTS version (v12). That should fix the issue. Another PR will be opened for that.

sauravkhdoolia:
Hmmm, that is weird. I've been looking through SL docs but in vain. All the docs related to Jasmin, Grunt, SauceLabs and Travis Addons are like 5 years old. Seems like world forgot of them. I'll wait for that.

arteconceito:
@sauravkhdoolia it would be nice to think about moving away from Grunt and use something more up-to-date and better maintained. For this project I would suggest Rollup.js. Should we create an issue to cover that?

sauravkhdoolia:
Sure, I think we should even move away from:

- SauceLabs to BrowserStack.
- Grunt to RollUp.js(Rollup support testing???)
- Jasmine  -> Mocha (may be)

PR #1559, opened 2020-08-05
Update _placeholder.scss

shadow-light:
| Q                | A
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| BC breaks?       | no
| Deprecations?    | no
| Fixed tickets    | none
| License          | MIT

### Description

Make it appear like element is editable (by showing text cursor) when hovering over placeholder (as would happen over a normal editable element). Currently a normal allow is shown (when hovering over placeholder) which makes it seem like the element is not editable, when it actually is.

--

### DIFF
--- a/src/sass/components/_placeholder.scss
+++ b/src/sass/components/_placeholder.scss
@@ -10,6 +10,7 @@
         white-space: pre;
         padding: inherit;
         margin: inherit;
+        cursor: text;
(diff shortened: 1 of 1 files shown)


arvidbjorkstrom:
Simple, but very nice tweak! :)

